PR #12, opened 2026-06-18
feat: add `nub node default <version>` command

### LINKED ISSUE #5: feat: allow setting specific node as default
Just configured a new machine so took advantage of it to checkout nub, but noticed that while it has a great support for handling node versions it does not support specifying a global node verison as the default one.

For example with a clean PATH, doing `nub node install lts` does populate the cache, but the system does not know where to look for the executable. This results in only nub knowing where to look for it, but other programs (or event the shims provided with `nub pm shim`) to remain locked out.

Would it be possible to have a command like `nub node default <version>` to set both NODE and PATH envs to act like a global fallback when `.node-version`/`.nvmrc`/`engines.node` is not present?

sandros94:
### user-global default Node

Resolves #5 

Adds `nub node default <version>`: the Node version used when a project pins nothing, and exposes it on `PATH` so non-nub programs find `node` / `npm` / `npx`.

Disclaimer: I've also used Opus to help me navigate and draft some of the changes, as I've written in rust only a handful of times.

#### What changed

Added a new command, new resolution step and rewrote the internal handling of shims to take into account `npm` and `npx` overridability. Generally speaking:

- `nub node default <version>` sets the default (auto-installs if uncached); `--unset` clears it.
- Resolution gains a lowest rung — `~/.nub/default`, after `engines.node`, before the bare-`PATH` node.
- The default's `bin/` is put on `PATH` via a `~/.nub/current` symlink.
- Shell-profile PATH block unified under `# nub path`: lists `~/.nub/shims` then `~/.nub/current` in fixed priority; the old `# nub shims` block migrates automatically.
- I've preferred to configure `nub node uninstall` to refuse the version that's set as the default, but we could rewrite this into auto-cleanup instead if required.
- docs updated accordingly

#### Why (decisions)

- **Unified PATH block.** Deterministic ordering, so a `pm shim`-routed `npm` always outranks the default distribution's `npm` regardless of which feature was enabled first.
- **PATH wiring is non-fatal.** The record is written before the symlink/profile step, so a wiring failure still leaves a working default plus a warning; re-running retries only that step.
- **Uninstall guard.** Prevents a dangling `~/.nub/default` and a broken `~/.nub/current`.
- **Fallback fixes a dead-end.** With no pin and Node only in nub's cache, resolution errored (`NoNodeOnPath`) — which also broke the PM shims on a clean system. The default is the user-wide fallback.
- **Default beats the system `node` in unpinned projects.** Matches nvm / volta / fnm: "default" (as far as I know) is what runs when nothing else specifies.
- **Stored outside the cache, as a concrete version.** A cache wipe must not drop an opted-in default; recording the resolved version (not `lts` / `22`) keeps the record and store in lockstep.
- **Exposes the whole `bin/`, as plain Node.** `npm` / `npx` are relative symlinks into `../lib` and only work with the bin dir intact; plain Node keeps a bare `node` unsurprising, while `nub <file>` still auto-switches per project.

#### Additional internal breaking change

I've touched the shipped `nub pm shim`: its profile marker changes `# nub shims` → `# nub path`, rewritten to the unified block. Existing blocks migrate on the next `pm shim` / `node default` / `unshim`.

While asking Opus for some review of the changes it also noted that both `SHIMS_POSIX_PATH_LINE` and `SHIMS_FISH_PATH_LINE` were stale, unused, public consts, so personally agreed on cleaning them up. Easy readdition if I've missed the underlying logic for those 🙏

#### Kept out of scope

Windows `PATH` automation — `default` records the version (resolution + shims use it) but prints a manual `PATH` instruction, exactly like shims one.

### DIFF
--- a/crates/nub-cli/src/cli.rs
+++ b/crates/nub-cli/src/cli.rs
@@ -4330,7 +4330,8 @@ fn run_node(args: &[String]) -> Result<i32> {
              \x20 install [<version>...]   provision version(s) into nub's cache (bare: the project pin)\n\
              \x20 ls                       list versions in nub's cache\n\
              \x20 uninstall <version>      remove a version from nub's cache\n\
-             \x20 pin <version>            write the project's Node pin"
+             \x20 pin <version>            write the project's Node pin\n\
+             \x20 default <version>        set the user-global default Node (--unset to clear)"
         );
         return Ok(0);
     }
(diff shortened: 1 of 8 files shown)


colinhacks:
Thanks for the proposal. For now Nub will keep no-pin projects resolving Node from PATH rather than adding a global default Node setting, so this PR will not be merged.

PR #24, opened 2026-06-19
nub-native: recover from mutex poison in tsconfig cache

nerkoux:
Both lock accesses in `load_for_dir` used `.unwrap()`, which would panic if the lock had been poisoned by an earlier panic. This changes them to `.unwrap_or_else(|e| e.into_inner())`, matching the pattern already used elsewhere in the codebase, so a previous transient panic doesn't cause all future `loadTsconfig` and `resolveTs` N-API calls to fail.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3043,7 +3043,7 @@ dependencies = [
 
 [[package]]
 name = "nub-cli"
-version = "0.1.2"
+version = "0.1.4"
 dependencies = [
  "anyhow",
  "aube",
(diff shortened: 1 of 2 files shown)


colinhacks:
Merged. Thanks @nerkoux — recovering the inner value via `unwrap_or_else(|e| e.into_inner())` is the right fix here: the tsconfig cache is a plain insert-or-get `HashMap`, the `Loaded` values are built before the lock is taken and are immutable, so a poisoned mutex leaves the map fully consistent and the recovered state is safe to use. Verified a clean release build and clippy on nub-native locally. (The Clippy/Format CI red is a pre-existing formatting issue in `cache.rs` on main, unrelated to this change, and is being handled separately.)

colinhacks:
Shipped in v0.1.5 (now published).

nerkoux:
Thanks for the acknowledgement @colinhacks 😁

PR #31, opened 2026-06-20
node: unflag wasm-modules (compat-backfill) + shadow-realm

colinhacks:
Adds two feature-matrix rows so nub auto-injects the experimental flag on the Node versions that still require it.

## wasm-modules (mechanical compat-backfill)

\`import` of `.wasm` files is **default-on in current Node** — the `--experimental-wasm-modules` flag became a NoOp at **24.5.0** (24.x line) and **22.19.0** (22.x line) via [nodejs/node#57038](https://github.com/nodejs/node/pull/57038). Per the unflag-on-prior-versions rule (default-on in current Node ⇒ backfill it where it was flagged), this is a mechanical banded unflag, same shape as the live sqlite row.

Bands (injected where the flag both exists and is required):

```
[18.19, 22.19)  inject   # 22.x line, pre-backport
[22.19, 23.0)   native   # backport landed
[23.0,  24.5)   inject   # 23.x EOL before the backport — stays flagged
[24.5,  inf)    native   # default-on
```

The 23.x line never received the #57038 backport (EOL before 24.5), so it stays flagged through the end of the line — verified against `.repos/node` `doc/api/esm.md`.

**Resolver passthrough:** nub leaves `.wasm` import specifiers untouched. `TRANSPILE_EXTS` covers only `.ts/.tsx/.mts/.cts/.jsx`, and the additive resolver (`resolveTs`) returns `null` for a `.wasm` specifier, so it falls through to Node native wasm loader — no transform, no re-resolve.

## shadow-realm

`--experimental-shadow-realm` exists from Node 18.13/19.0 (below the floor) and has **never** been made default-on through Node 26 (TC39 Stage 3, rides V8 `--harmony-shadow-realm` staging). Inject across the whole supported floor `[18.19, inf)`, like vm-modules.

> [!IMPORTANT]
> shadow-realm is **NOT** the mechanical case wasm-modules is — it is a Stage-3, not-default-on-anywhere global, and turning it on by default is a product/posture call, not a backport. The thread `.fray/v0x-shadow-realm-unflag.md` recommends DEFER. It is included here per a coordinator-relayed decision, but a coordinator relay is not user authority — **this row needs Colin own sign-off before merge.** If the call is to defer, drop the shadow-realm commit/row and ship wasm-modules alone.

## Tests

- Band tests in `feature_matrix.rs` (matrix layer) and end-to-end `compute_inject_flags` band tests in `flags.rs` (the spawn path) for both features.
- `cargo test -p nub-core --lib` → 274 passed; clippy + fmt clean.

### DIFF
--- a/crates/nub-core/src/node/feature_matrix.rs
+++ b/crates/nub-core/src/node/feature_matrix.rs
@@ -149,6 +149,23 @@ pub static FEATURES: &[Feature] = &[
         )],
         evidence: "flag added Node 9.6.0 (#14253); never unflagged through Node 26",
     },
+    // ── ShadowRealm global ──────────────────────────────────────────────────
+    // `--experimental-shadow-realm` gates `globalThis.ShadowRealm` (TC39 Stage 3:
+    // an isolated global with fresh intrinsics). The flag was added in Node
+    // 18.13.0 / 19.0.0 — below nub's 18.19 floor — and has NEVER been made
+    // default-on through Node 26 (still experimental, rides the V8
+    // `--harmony-shadow-realm` staging). So, like vm-modules, inject across the
+    // ENTIRE supported floor: [18.19.0, ∞). It survives as an accepted bool at
+    // every higher version, so the open-ended band never over-injects into an
+    // unsupported range.
+    Feature {
+        name: "shadow-realm",
+        mitigations: &[(
+            band((18, 19, 0), None),
+            Mitigation::Unflag("--experimental-shadow-realm"),
+        )],
+        evidence: "flag added Node 18.13.0/19.0.0; never default-on through 26 (TC39 Stage 3)",
+    },
     // ── EventSource global ──────────────────────────────────────────────────
     // #51575 ("add EventSource Client"). Landed on the 22.x line at 22.3.0 and was
     // backported to the 20.x LTS line at 20.18.0. The 21.x line was already EOL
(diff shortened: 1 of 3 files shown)


colinhacks:
## CI failure — do not merge

**Test (macos-14, node 24)** and **Test (windows-latest, node 24)** fail. Both on the same bug introduced by the shadow-realm row:

```
ERR_WORKER_INVALID_EXEC_ARGV: Initiated Worker with invalid execArgv flags: --harmony-shadow-realm
```

**Root cause:** `--experimental-shadow-realm` implies `--harmony-shadow-realm` internally. Node adds the implied V8 harmony flag to `process.execArgv`. The worker polyfill (`runtime/worker-polyfill.mjs:120`) passes `process.execArgv` directly to `NodeWorker` — Node 24 rejects `--harmony-*` flags in worker `execArgv` with `ERR_WORKER_INVALID_EXEC_ARGV`.

Failing tests: `worker_transpiles_ts_entry`, `worker_message_roundtrip`, `worker_throw_surfaces_to_parent_onerror` — all worker tests on Node 24.16.0.

**Fix required in `runtime/worker-polyfill.mjs`** (line 120): filter `--harmony-*` flags from `execArgv` before passing to NodeWorker:

```js
execArgv: process.execArgv.filter(f => !f.startsWith('--harmony')),
```

This is the correct fix regardless — any harmony-implying flag injected by nub would trigger this. The filter is safe: harmony flags are V8 internals that Node documents as invalid for workers.

This fix should land before or with this PR. The wasm-modules row is clean; only shadow-realm triggers this.

PR #36, opened 2026-06-20
pm: implement npm-only verbs natively (whoami/owner/token/search/pkg/set-script)

colinhacks:
Implements the formerly-refused npm-only PM verbs natively, so nub no longer returns \`ERR_NUB_NPM_ONLY_COMMAND\` (exit 82) for them. No shell-out — everything hits the registry directly (reusing aube's existing \`.npmrc\` auth + registry client) or edits the local \`package.json\`.

## Verbs

| verb | implementation |
|---|---|
| \`search <q>\` | registry \`/-/v1/search\` (public), pnpm-shape output + \`--json\` |
| \`whoami\` | \`/-/whoami\` with \`.npmrc\` bearer auth |
| \`owner ls/add/rm\` | collaborators GET + maintainers read-modify-write PUT |
| \`token list/create/revoke\` | \`/-/npm/v1/tokens\` CRUD (goes one step beyond pnpm, which doesn't implement \`token\` — npm parity) |
| \`pkg get/set/delete/fix\` | local \`package.json\` edits via a ported property-path module |
| \`set-script <name> <cmd>\` | \`scripts.<name>\` sugar over the same writer |

\`stage\` is dropped (it is not a real npm/pnpm command) — it now falls through to the unknown-command path.

## Architecture

The divergence is structural, not a runtime gate: the implementations are new \`aube::commands::<verb>::run\` modules in \`vendor/aube\`; nub's \`pm_engine\` calls them directly. Standalone aube's own \`lib.rs\` dispatch is left routing these verbs to its \`npm_fallback\` stub, so **standalone aube is byte-for-byte default-preserving**.

User-facing login hints go through \`aube_util::cmd\` so they render \`nub login\` under the nub embedder (verified: \`whoami\` unauthenticated prints \`run \`nub login\` first\`).

## vendor/aube

Pin bumped \`b88db61\` → \`29f233c\`. The aube-side change lives on \`nubjs/aube:native-npm-verbs\` — **merge that into \`nub-fork\` and re-point the pin to the \`nub-fork\` commit before merging this PR** (the pin currently points at the \`native-npm-verbs\` commit, which is pushed/CI-safe but should graduate onto \`nub-fork\`).

## Tests

- aube: property-path units, \`pkg\` manifest round-trips, wiremock coverage for whoami/search/owners.
- nub-cli: rewrote the test that asserted the old npm-only refusal to cover native \`pkg\`/\`set-script\` round-trips and assert the registry verbs no longer emit \`ERR_NUB_NPM_ONLY_COMMAND\`.
- Build / clippy (\`-D warnings\`) / fmt clean on both sides; smoke-tested the real binary.


colinhacks:
Shipped in v0.1.9: https://github.com/nubjs/nub/releases/tag/v0.1.9

PR #40, opened 2026-06-21
cli: dispatch -r/--filter before info verbs to the PM engine

colinhacks:
## Problem

`nub -r <verb>` and `nub --filter <x> <verb>` for the read-only info verb family (`list`/`ls`/`la`/`ll`, `why`, `outdated`, `licenses`, `audit`, `peers`, `query`, `check`, `sbom`, `bin`, `root`, `view`, `search`) fell through to the Node file-runner instead of dispatching to the PM engine:

- `nub -r list` → `Error: Cannot find module 'list'`
- `nub --filter a list` → `node: bad option: --filter`

`pnpm -r <cmd>` / `pnpm --filter <x> <cmd>` are the canonical workspace invocations, so this was a high-traffic compat break.

## Root cause

`normalize_leading_run_flags` reorders a leading run-flag (`-r`/`--recursive`/`--filter`/`-F`/…) to sit after the verb, but only when the verb was in the hand-curated `NORMALIZABLE_LEADING_FLAG_VERBS` list — which covered the install family only. The info family was absent, so the verb token was never reordered and reached Node's file-runner as a `--require` module / bad option.

## Fix

Replace the hand-curated list with `is_normalizable_leading_flag_verb`, which recognizes any PM verb nub can dispatch — the same authority the bareword dispatch arm consults (`SUBCOMMANDS` + the engine verb registry `pm_engine::lookup_verb`). Tying the reorder set to the dispatch set keeps them from drifting: every verb nub dispatches in first position is now also reorderable behind a leading workspace flag. Verbs that don't accept `-r`/`--filter` remain safe to reorder — clap/the engine rejects the unsupported flag downstream with a proper non-zero error, matching pnpm.

## Exit-code note

The conformance catalog flagged a secondary "exit 0 on crash" defect. Reproduced against the pre-fix code: `nub -r list` returned raw exit **1** and `nub --filter a list` returned exit **9** (Node's bad-option code) — both non-zero. The reported "exit 0" was a measurement artifact of a `| head` pipe (PIPESTATUS of the tail). The file-runner faithfully propagates Node's exit code via `exit_code_from_status`; no exit-code change was needed.

## Test

Adds `leading_global_flags_before_info_verb_reach_engine` to `cli_grammar_parity.rs`. The existing grammar table can't guard this case (appending `--help` short-circuits before dispatch), so the new test dispatches for real and asserts the file-runner crash markers (`Cannot find module`, `node: bad option`) are absent.

## Verification

- `cargo fmt --check` — clean
- `cargo clippy --all-targets --all-features -- -D warnings` — clean
- `cargo test -p nub-cli --test cli_grammar_parity` — 8 passed (1 pre-existing ignore)
- `cargo test -p nub-cli --test pm_verbs` and the cli unit tests — green
- e2e against a tmp workspace fixture: `nub -r list` / `nub --filter a why a` now reach the engine; file-run, `--import` passthrough, and `nub -r <file>` (no-verb) paths unchanged.

Refs the pnpm conformance work (D1).

### DIFF
--- a/crates/nub-cli/src/cli.rs
+++ b/crates/nub-cli/src/cli.rs
@@ -1008,41 +1008,38 @@ const PM_VERBS: &[&str] = &[
     "migrate",
 ];
 
-/// The verbs a leading flag-run may precede. pnpm accepts `pnpm -r <verb>` for
-/// every workspace-aware verb, not just `run`/`exec` — so a leading `-r`/`-F`/…
-/// before an install-family verb (`install`/`ci`/`add`/…) must reorder the same
-/// way, or it falls through to the Node-passthrough path and Node fails on
-/// `--require install` (`Cannot find module 'install'`).
-const NORMALIZABLE_LEADING_FLAG_VERBS: &[&str] = &[
-    "run",
-    "exec", // the original run/exec set
-    "install",
-    "i",
-    "ci", // native install-family verbs
-    "add",
-    "remove",
-    "rm",
-    "uninstall",
-    "un", // engine add/remove family
-    "update",
-    "up",
-    "dedupe",
-    "prune",
-    "rebuild",
-    "fetch",
-    "link",
-    "unlink",
-    "import",
-];
+/// Whether a leading flag-run may be reordered to sit *after* `verb`. pnpm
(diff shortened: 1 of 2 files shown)


colinhacks:

PR #41, opened 2026-06-21
pm: align list/outdated --json empty state and `run` no-arg exit with pnpm

colinhacks:
Fixes two pnpm-compat error-contract divergences from the conformance audit (D2, D3). The third (D5, fetch/prune no-lockfile exit codes) is routed back as an open question rather than landed — see below.

## D2 — `--json` emits no JSON in the empty state

With a valid package.json but no install/lockfile:

- `nub list --json` emitted empty stdout plus a prose stderr note, so `nub list --json | jq` got no input.
- `nub outdated --json` emitted empty stdout.

Real pnpm emits a valid JSON array (`[{name, version, path, ...}]`) for `list` and `{}` for `outdated`, both exit 0.

Fix: under `--json` (or `--format json`), the no-lockfile short-circuit now emits the correct empty shape on stdout — a one-element importer-header array for `list` (name/version/path, matching nub's populated `list --json` header), `{}` for `outdated`. The non-JSON paths keep the rebranded "No lockfile found" note on stderr, so the brand boundary is unchanged.

## D3 — `nub run` (no script) exited 1; pnpm exits 0

`nub run` with no script name now exits 0 and lists the package's scripts on stdout (or prints `There are no scripts specified.` when none), matching `pnpm run` with no args. Previously it exited 1 with the listing on stderr, so CI that probes `pnpm run` and branches on the exit code saw a spurious failure.

This is the explicit no-arg `run` verb mirroring pnpm. It does not relax nub's no-implicit-script-shortcuts stance — bareword `nub test`/`nub start` are still rejected; only the no-arg `run` path changed.

## D5 — routed to a maintainer decision, not landed

`fetch`/`prune` with no lockfile exit non-zero (fetch exit 1, prune exit 10 `ERR_NUB_NO_LOCKFILE`); pnpm exits 0 `Already up to date`. Grounding showed the strict guard is a deliberate, explicit design in the embedded engine (an intentional `requires a lockfile` error message, not a NotFound bubbling through), and pnpm's `prune` actually resolves+installs rather than no-op'ing. Flipping an engine-deliberate error into silent success is an error-contract judgment call, so it is surfaced for maintainer sign-off rather than landed here.

## Tests

- Updated the run-no-script integration test to the new exit-0/stdout contract.
- Added `run_without_script_in_scriptless_package_exits_zero` (D3, no-scripts case).
- Added `json_queries_emit_parseable_empty_shape_without_a_lockfile` (D2, list/outdated/`--format json`).

Verified locally: `cargo fmt --check`, `cargo clippy --all-targets --all-features -- -D warnings`, scoped `cargo test -p nub-cli` (info_engine + integration), and e2e against the dev binary diffed vs pnpm 10.15.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3043,7 +3043,7 @@ dependencies = [
 
 [[package]]
 name = "nub-cli"
-version = "0.1.7"
+version = "0.1.9"
 dependencies = [
  "anyhow",
  "aube",
@@ -3079,7 +3079,7 @@ dependencies = [
 
 [[package]]
 name = "nub-core"
-version = "0.1.7"
+version = "0.1.9"
 dependencies = [
  "anyhow",
  "blake3",
@@ -3107,7 +3107,7 @@ dependencies = [
 
 [[package]]
 name = "nub-native"
-version = "0.1.7"
+version = "0.1.9"
 dependencies = [
  "blake3",
  "json5",
--- a/crates/nub-cli/src/cli.rs
+++ b/crates/nub-cli/src/cli.rs
@@ -2164,10 +2164,22 @@ fn run_script(
     // No script name (`nub run`): list available scripts instead of a raw clap
     // "required argument" error — same shape as the missing-named-script path.
     let Some(script) = script else {
-        bail!(
-            "no script specified\n\nAvailable scripts:\n{}",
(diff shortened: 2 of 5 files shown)


colinhacks:

PR #42, opened 2026-06-21
pm: flatten npm nested-object overrides into selector keys

colinhacks:
## Summary

`overrides_map()` in `vendor/aube/crates/aube-manifest/src/lib.rs` read override values only via `as_str()`, silently dropping any entry whose value is a nested object. npm allows the nested-object `overrides` form, e.g.:

```json
{ "overrides": { "foo": { ".": "1.0.0", "bar": "2.0.0" } } }
```

A dropped entry meant the intended version pin was never applied, producing a silently-wrong resolved dependency graph.

## Fix

npm's nesting encodes an ancestor chain outermost-first, which is the same direction as pnpm's `parent>child` selector. The engine now flattens the nested form into pnpm `>`-joined selector keys instead of dropping it:

- a string child `"child": "spec"` becomes `<chain>>child`;
- the reserved key `"."` targets the chain's own innermost package (`<chain>`);
- a nested object recurses with the child appended;
- a version-pinned outer key (`"bar@2.0.0"`) passes through to the selector verbatim.

pnpm's own `overrides` are string-only (pnpm errors on an object value), so this path is reached only for npm-authored manifests; the pnpm/yarn paths are unchanged.

## Engine change

Lands on the `vendor/aube` fork (`nub-fork` commit f1f602b) and bumps the submodule pin. The change is default-preserving for standalone aube (pnpm path byte-for-byte unchanged) and is upstreamable to jdx/aube.

## Tests

In `aube-manifest`: replaced the obsolete `overrides_map_skips_object_values` test (which asserted the bug) with regression tests for nested child, `.`+sibling, deep 3-level chain, versioned parent, alias value, and an explicit non-silent-drop guard. Existing `aube-resolver`/`aube-lockfile` override tests already cover that the flattened selectors match correctly.

Verified locally: `cargo clippy --all-targets --all-features -- -D warnings` clean, `cargo fmt --check` clean, `cargo build -p nub-cli --profile fast` green against the new pin; aube-side `cargo test -p aube-manifest` (121) / `-p aube-resolver override` (26) / `-p aube-lockfile override` (28) all pass.

Refs the B-10 pm-compat correctness item.


colinhacks:

PR #48, opened 2026-06-21
pm: honor yarn supportedArchitectures, mTLS cert/key, npmAlwaysAuth

colinhacks:
Reads three previously-ignored `.yarnrc.yml` settings and routes them through the existing engine machinery, **plus** a bundled one-line trust-list fix to unblock CI on a single pin advance. Bumps the `vendor/aube` pin to `e7aaafc` (on `nub-fork`).

## Yarn config (B-3 / B-9 / B-12)

- **supportedArchitectures** — translated to the JSON-object `supportedArchitectures` setting (the same channel `packageExtensions` uses) and unioned into the resolver's platform filter at install time, so optional/platform deps are filtered for the declared `os`/`cpu`/`libc`. The nub-side FATAL gate that aborted on this key is removed — the arch-filter resolver is built and now fed for yarn, same as the pnpm path.
- **httpsCertFilePath / httpsKeyFilePath** (top-level and per-host `networkSettings`) — the referenced PEM files are loaded and emitted as the inline `cert`/`key` the registry client's mTLS identity consumer already takes.
- **npmAlwaysAuth** (top-level, per-registry, per-scope) — translated to `always-auth`; the registry client now attaches the home registry's token to requests on a different origin (e.g. tarballs on a separate CDN) when set, instead of sending them unauthenticated.

All three are additive reads gated on the user's explicit yarn config; the default (no setting) path is unchanged.

## Bundled CI unblock: nanoid trust-downgrade

`nanoid@3.3.14` was published to npm without trust evidence after an earlier version carried a trusted publisher, so `trustPolicy=no-downgrade` (the aube default) rejects it with `ERR_NUB_TRUST_DOWNGRADE` (exit 23) at `nub install` — which breaks the daily-driver on `main` and every open PR. `nanoid` is added to `DEFAULT_TRUST_POLICY_EXCLUDES` alongside the other provenance-churn packages. Bundled here so a single pin advance fixes both concerns at once; the other PRs rebase onto this after merge.

## Verification

- aube: 7 new yarn-config unit tests + the trust-list test (39 trust tests) pass; `cargo build` clean.
- nub: `cargo clippy --all-targets --all-features -- -D warnings` clean, `cargo fmt --check` clean, `cargo test -p nub-cli unsupported_config` green.
- e2e: a `.yarnrc.yml` with a `supportedArchitectures` block no longer aborts `nub install`. The daily-driver `install` step now passes against the new pin (it exits 23 on `main` due to nanoid); main + `npm_config_trust_policy=off` runs the full daily-driver green, confirming nanoid is the sole install-stage blocker.

Docs: the yarn compat tables (`yarn.mdx`, `install/index.mdx`) now mark the three yarn settings as supported.

### DIFF
--- a/crates/nub-cli/src/pm_engine/unsupported_config.rs
+++ b/crates/nub-cli/src/pm_engine/unsupported_config.rs
@@ -243,10 +243,11 @@ pub(crate) enum ScanResult {
 ///
 /// The FATAL set is deliberately SHORT — only fields whose silent omission
 /// produces a correctness-divergent install AND which nub cannot honor:
-/// npm `legacy-peer-deps` (different peer graph), npm `install-strategy=nested`
-/// (different resolution/layout), and yarn `supportedArchitectures` (changes
-/// which platform deps land — the arch-filter resolver isn't built). PnP stays
-/// a WARN+downgrade (handled separately in `warn_if_pnp_requested`), not
+/// npm `legacy-peer-deps` (different peer graph) and npm
+/// `install-strategy=nested` (different resolution/layout). yarn
+/// `supportedArchitectures` is NOT here — the engine honors it via the
+/// arch-filter resolver. PnP stays a WARN+downgrade (handled separately in
+/// `warn_if_pnp_requested`), not
 /// promoted. `checksumBehavior`/`enableHardenedMode` are NOT here: aube verifies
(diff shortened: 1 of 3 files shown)


colinhacks:

PR #49, opened 2026-06-21
run: honor npm_config_reporter=silent and add /regex/ script selection

colinhacks:
Two pnpm-compat gaps on the `nub run` front-door surface, each verified against real pnpm 10.15.1.

## 1. `npm_config_reporter=silent` suppresses the run preamble
pnpm suppresses its `> name@ <script>` banner when `npm_config_reporter=silent` is set; nub's `$ <cmd>` preamble keyed only on the `-s` / `--reporter` flags. Now honored when no explicit `--reporter` flag is present. Only `reporter=silent` suppresses — `npm_config_loglevel=silent` does not, matching pnpm.

## 2. `/regex/` script selection
`nub run "/^build:/"` now runs every script whose name matches the literal, in package.json order — pnpm's regex script selector (`getSpecifiedScripts` + `tryBuildRegExpFromCommand`). An exact name still runs just that one script; a literal carrying RegExp flags is rejected with pnpm's message; a regex matching nothing is a missing-script error (exit 1).

Matched scripts each get their pre/post lifecycle and the forwarded args. The set runs concurrently by default (`min(4, cpus)`; `--sequential` serializes in package.json order), and the overall exit code is the first failing script's code under the default bail.

Note: space-separated `run a b` is intentionally NOT a multi-script form — `nub run a b` and `pnpm run a b` both run only `a` with `b` as an arg. Only the regex form was a real gap.

## Verification (vs pnpm 10.15.1)
- reporter: `npm_config_reporter=silent` drops the preamble; `npm_config_loglevel=silent` keeps it; `--reporter=silent` flag drops it.
- regex: same matched scripts, same order, same passthru-args, same exit codes (failure → the script's code, no-match → 1, success → 0); byte-identical flags-rejection message.

## Tests
- 5 unit (`select_scripts`: exact-name wins, regex matches all in order, no-match, plain-name no-match, flags rejected).
- 3 integration (reporter-silent suppression incl. loglevel-does-not; regex runs the matching set + no-match exit 1; failure exit-code propagation).

Adds the `regex` crate as a direct nub-core dependency (already present transitively).


colinhacks:

PR #51, opened 2026-06-21
docs(agents): record yarn read-only as settled design position; fix stale NUB_ claim

colinhacks:
## Summary

- Adds a named, load-bearing design position to the core-design-positions section of `AGENTS.md` recording that **yarn is read-only** — the mutating-install family (`install`/`add`/`remove`/`update`/`dedupe`/`ci`) refuses to write `yarn.lock` (classic or berry), `nub pm use yarn` being the only write path (one-shot identity switch, not normal PM flow). Includes the WHY so the decision is not re-litigated.
- Corrects a stale note in the blog/marketing copy section that called `"No NUB_* environment variables"` a fine public claim. Three sanctioned user-facing PM knobs exist (`NUB_CACHE_DIR`, `NUB_CONCURRENCY`, `NUB_PRIMER_TTL`), so the claim is false — updated to the correct narrower public promise.

## Test plan

- Docs-only change; no build needed.
- Verified the two edited paragraphs in context; no line-wrap introduced (long-line convention maintained).


colinhacks:

PR #54, opened 2026-06-21
test: replace wall-clock timing with structural concurrency check in workspace_parallel_timing

colinhacks:
## Summary

- `workspace_parallel_timing` was flaky on contended Windows CI runners: it asserted that the parallel run's wall-clock elapsed time was ≥1 s faster than the serial run, a margin that fails when the runner is under load (confirmed failure on #49 unrelated to its diff).
- Replace the wall-clock delta with a **structural concurrency assertion**: each new `slow-stamp` script writes millisecond-precision start/end timestamps to files in a temp directory (path passed via `NUB_STAMP_DIR`). After the parallel run, the test checks that at least two execution windows overlap — something that is only possible if packages ran concurrently.
- The serial control run (`--workspace-concurrency=1`) is checked to produce non-overlapping windows, confirming the assertion would catch a genuine regression where `--parallel` silently ran packages one at a time.
- The existing `slow` scripts are unchanged; `slow-stamp` is a new script added alongside them in each fixture package.

## Test plan

- [x] `cargo test -p nub-cli --test integration workspace_parallel_timing` passes (run 3× locally, all green in ~6 s each)
- [x] `cargo clippy --all-targets --all-features -- -D warnings` clean
- [x] `cargo fmt --check` clean
- [x] Verified with manual logic check that non-overlapping (serial) windows produce `any_overlap = false` → the assertion would fire on a genuine regression


colinhacks:

PR #45, opened 2026-06-21
site: rework start.md into a permission-gated adoption flow + order llms.txt by nav

colinhacks:
Two site-only changes.

## `site/public/start.md` — procedural, permission-gated adoption flow

Replaces the prior prose guide with a numbered walkthrough for an AI coding agent adopting nub in a project:

1. Install nub (check `nub --version` first; install only on user approval).
2. Learn nub's surface via `nub agent docs` / the online docs.
3. Use nub for everyday commands (file/script/package runner, install) — a before/after table.
4. Offer a *read-only* migration investigation (redundant-tooling table).
5. Surface findings as a plan, implement only on an explicit yes.
6. Record the convention in `AGENTS.md`.

The governing rule throughout: explain before you act — nothing installs, edits, or removes without the user's go-ahead; reading/learning is the only unprompted action.

## `site/src/app/llms.txt/route.ts` — order the Docs index by nav, not `getPages()`

- The `/llms.txt` Docs section previously listed pages in fumadocs' arbitrary internal `source.getPages()` order (FAQ first, etc.), ignoring the `meta.json` nav order.
- It now walks the fumadocs page tree depth-first (the tree already reflects `meta.json`): each folder emits its index/overview page first, then its children in meta order; separators are skipped.
- Any page the tree doesn't surface is appended at the end, so a newly-added or misconfigured page never silently drops out of the index.
- Blog section unchanged.

### Verification

- `tsc --noEmit` passes clean.
- Empirically dumped the resulting order against the real page tree: all **23** doc pages present, in nav order — `/docs` → Runtime (+ TypeScript, JSX, Decorators, Module resolution, Env, Loaders, Workers, Web Storage) → Watch → Run → Nubx → Package manager (+ pnpm, npm, Bun, Yarn) → Node → pm → pm-shim → FAQ → GitHub Action.

PR #55, opened 2026-06-21
start.md: add step 7 — offer to add a nub Claude Code skill

colinhacks:
## Summary

- Adds a new **step 7** to the permission-gated `start.md` adoption flow, offering to write `.claude/skills/nub/SKILL.md` into the user's project
- The step is permission-gated (consistent with the rest of the flow): the agent describes what it will write, then waits for an explicit yes before creating the file
- If accepted, the agent runs `nub agent skill` for the content (with a fallback to `https://nubjs.com/skill.md`)
- Drops the old parenthetical at the bottom of step 6 that mentioned `nub agent skill` in passing (the new step covers this more directly)
- No migration skill — migration is handled by start.md + the /guides/bun-to-nub guide

The skill content itself (`site/public/skill.md`) is **not changed** — it is the existing evergreen skill file that `nub agent skill` already prints, and this PR just wires the adoption flow to offer installing it.

## For review

This is a **prototype for Colin's review** — the offer wording, the step placement (after step 6 / before Notes), and the skill content at `site/public/skill.md` are all open to adjustment before this is the live default. Nothing changes in code paths or behavior; only `site/public/start.md` (a markdown file served to AI agents) is modified.

## Test plan

- [ ] Read `site/public/start.md` top-to-bottom and confirm step 7 fits naturally after step 6
- [ ] Check that the offer wording matches start.md's register (permission-gated, describe-then-ask style)
- [ ] Confirm `site/public/skill.md` is the right content to install (it is what `nub agent skill` outputs)
- [ ] Verify the `https://nubjs.com/skill.md` fallback URL is correct (public/ files are served at the root)

Refs #45


tordans:
AFAIK the cleanest cross-agent-harness-way to do this today is…
- put the skill in `.agents/skills/nub/SKILL.md` which AFAIK cursor and most other harnesses read
- create an optional `.claude/skills/nub/SKILL.md` that uses clause `@.agents/skills/nub/SKILL.md` Markdown-Import/Link-Syntax that is described in https://youtu.be/FDxW2bfBOWE?si=SWRKYfHLFZqF2oEv&t=495 and https://code.claude.com/docs/en/memory#import-additional-files

Or, another approach would be to just have the installer suggest to run one of the skills directory. That would have the advantage to delegate this to some likely already used party which will also set up a lock file of skills so updating will be easier later.
- https://www.skills.sh/ – `nubx skills add nubjs/skills` would list all skills form that repo and allow installing them with the harness the user selects
- https://github.com/withastro/rosie is the same AFAIK

colinhacks:
Thanks — this is the right call. Adopted in 1e1943a: the skill now goes to `.agents/skills/nub/SKILL.md` as the cross-harness home, with an optional `.claude/skills/nub/SKILL.md` that `@`-imports it on Claude Code rather than duplicating. Kept it agent-agnostic — the `.agents/` path is the default and the Claude import is just the convenience layer. Also noted the skills-manager route (`nubx skills add nubjs/skills` / rosie) as a future option for once a `nubjs/skills` repo exists.

colinhacks:

PR #61, opened 2026-06-21
fix(pm): three bun.lock writer round-trip defects (registry URL, block order, bin placement)

colinhacks:
Bumps the vendored `aube` pin to `27ce644` to land three bun.lock **writer** round-trip fixes (all in `vendor/aube/crates/aube-lockfile/src/bun/`). Each was verified against bun 1.3.14's own writer in `.repos/bun/src/install/lockfile/bun.lock.zig`, and each ships a byte-identity regression fixture the pre-existing native-bun fixture didn't exercise.

## Fixes

- **B-4 (data loss) — registry URL dropped.** The bun.lock parser discarded npm tuple slot 1 (the registry/tarball URL) and the writer hardcoded `""`. A scoped/private-registry dep therefore lost its URL on round-trip and the next resolve re-routed to the default npm registry (404 / name-squat risk). Slot 1 is now captured on parse — gated to the registry shape (a non-integrity string *before* the meta object, never a git repo-tag *after* it) — onto `LockedPackage.tarball_url`, and re-emitted. The empty slot (bun's default-registry marker) stays empty.
- **B-7 (byte-identity) — top-level block order.** nub emitted `overrides → patchedDependencies → trustedDependencies`; bun's native order is `trustedDependencies → patchedDependencies → overrides → catalog → catalogs`. Reordered to match.
- **B-8 (byte-identity) — per-package `bin` placement.** nub emitted `bin` before the platform filters; bun emits `os → cpu → libc → bin` (`bin` last). Moved.

All three are **default-preserving for standalone aube** (byte-identity-correctness + data-preservation = strict improvements) and **upstreamable to jdx/aube**.

## Verification

- `cargo test -p aube-lockfile` in an isolated clone: **415 passed, 0 failed** (incl. the existing `test_write_byte_identical_to_native_bun` + the three new fixtures).
- `cargo clippy -p aube-lockfile --all-targets -- -D warnings`: clean.
- nub side: `cargo clippy -p nub-cli --all-targets --all-features -- -D warnings` clean; `cargo fmt --check` clean; nub-cli builds against the new pin.

## Follow-ups (discovered, out of scope here)

The RT-1 fixture surfaced two *additional* nub-vs-bun byte-identity drifts that are **not** part of this lane (block-order vs rendering): nub renders a named `catalogs` inner object inline (bun: multi-line) and emits `"packages": {\n  }` for an empty section (bun: `{}`). Tracked for a separate lane; data still round-trips through parse.

Refs the bun-lockfile-writer-fixes lane.


colinhacks:

PR #62, opened 2026-06-21
docs(nubx): document the minimumReleaseAge cooling window

### LINKED ISSUE #60: Clarify if `minimumReleaseAge` applies to `nubx`
https://nubjs.com/docs/install#security goes into details on `minimumReleaseAge` but https://nubjs.com/docs/nubx does not mention it.

The ideal setup would IMO be if the projects `minimumReleaseAge` is respected for `nubx`.
I assume the default 24h does apply?

There are tickets in eg. Bun to allow for a global `minimumReleaseAge` to be configured but given the default 24h I think this is not that important.

colinhacks:
Issue #60 (tordans) asked whether `minimumReleaseAge` applies to `nubx` and noted the nubx docs didn't mention it. Investigating surfaced a conformance gap: `nubx`'s registry-fetch (`nub dlx`) installs into a scratch TMPDIR project, so it saw only user/global npmrc + the 24h default — a **project-local `.npmrc`** override (registry, scoped auth, `minimumReleaseAge`) silently did not apply, whereas npm's `npx` reads the invocation cwd's `./.npmrc`.

Verified npx's behavior empirically (project `.npmrc` → bad registry makes `npx` fail; → good registry makes it run), then conformed `nubx` to it.

**Changes:**
- **aube fork (`vendor/aube`, nub-fork `b0d8e98`):** `feat(dlx)` forwards the user's project-scope `.npmrc` into the dlx scratch install before it runs, mirroring how runtime resolution already reads the user's project (not the scratch dir) for `.nvmrc`/`devEngines`. Only the project scope is forwarded; user/global npmrc are read independent of cwd. Best-effort — no project `.npmrc` writes nothing, behavior unchanged. Default-preserving for standalone aube.
- **Pin bump** `e7aaafc` → `b0d8e98` (also pulls in the upstream `aube-lockfile` bun.lock round-trip fix that landed on nub-fork since the old pin).
- **Docs:** nubx "Supply-chain cooling window" section now states project-local `.npmrc` applies to `nubx`, matching npx.

**Verified e2e against `target/fast/nub` at the bumped pin:**
- Project `.npmrc` `minimumReleaseAge` gate → `nubx cowsay` blocked with `ERR_NUB_NO_MATURE_MATCHING_VERSION`.
- Project `.npmrc` registry → `nubx` fetches from it (matches npx).
- No project `.npmrc` → 24h default still applies, cowsay runs (no regression).
- Project `minimumReleaseAge=0` → disables the gate.

Gates: `cargo fmt --check` + `cargo clippy --all-targets --all-features` clean for nub-cli; `cargo clippy -p aube --all-targets` clean for the changed crate; integrated tree builds at the new pin.

Closes #60


colinhacks:

PR #69, opened 2026-06-22
ci: mechanize code-quality + brand boundary as lints

colinhacks:
Mechanizes two things that were previously only review-enforced, so they can't regress by a review miss. Mirrors uv's `clippy.toml`/`[lints]` pattern, but scoped to a **clean, zero-violation subset** — no ocean-boiling, no large refactor.

## What landed

**Workspace restriction lints** (`[workspace.lints.clippy]`, inherited by every nub crate via `[lints] workspace = true`), promoted to warn → hard-gated by the existing `cargo clippy --all-targets --all-features -- -D warnings` CI job:

- `dbg_macro`, `todo`, `unimplemented`

All three measured at **0 violations** across the whole workspace (2026-06-21), so they land clean and act purely as a forward guard against unfinished/debug code shipping.

**Static brand-boundary gate** — `tests/brand-lint/check-env-reads.sh`, wired as a step in the `clippy` CI job (cheap grep, no build, same `run_rust` gate). Fails if nub crate *source* READs an `AUBE_*`-branded env var. Enforces the settled AGENTS.md invariant ("the nub runtime respects ZERO `AUBE_*` env vars") mechanically. Currently 0 violations; verified the gate both passes clean and catches an injected violation. SETTING an `AUBE_*` var for a child the engine spawns (`Command::env`) is not matched — only reads are.

## What was deliberately omitted (and why)

Measured violation counts made these a refactor, not a review-miss guard — noted as follow-ups, not in this PR:

- `print_stdout` / `print_stderr` — a CLI legitimately prints and there is **no output wrapper** (239 sites across 15 production source files). Per the thread's guidance, skipped absent a wrapper.
- `unwrap_used` / `expect_used` / `panic` / `exit` — 1300+ sites, overwhelmingly in tests. Would need `allow-in-tests` config + nontrivial production cleanup.
- Forcing all env access through a wrapper (uv's `disallowed-methods` approach, ~89 direct `std::env::var` sites) — the larger refactor that would let `disallowed-methods` ban direct env access; the grep-gate holds the boundary today without it.

## Verification (local, pre-push)

- `cargo clippy --all-targets --all-features -- -D warnings` → exit 0 (new lints clean)
- `cargo fmt --check` → clean
- `tests/brand-lint/check-env-reads.sh` → passes; injected `env::var("AUBE_CACHE_DIR")` → correctly fails

Refs the contribution-quality investigation (uv/ruff adopt-now pick #1).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -166,6 +166,11 @@ jobs:
           components: clippy
       - uses: Swatinem/rust-cache@v2
       - run: cargo clippy --all-targets --all-features -- -D warnings
+      # Static brand-boundary gate: no AUBE_*-branded env-var READS in nub crate
+      # source (the embedded engine's brand is never nub's config surface). Cheap
+      # grep, no build — runs alongside clippy on the same run_rust gate.
+      - name: Brand lint (no AUBE_* env reads)
+        run: tests/brand-lint/check-env-reads.sh
 
   test:
     name: Test (${{ matrix.os }}, node ${{ matrix.node }})
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3043,7 +3043,7 @@ dependencies = [
 
 [[package]]
 name = "nub-cli"
-version = "0.1.9"
+version = "0.1.10"
 dependencies = [
  "anyhow",
  "aube",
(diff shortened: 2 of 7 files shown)


colinhacks:

PR #72, opened 2026-06-22
fix(pm): treat ~/.npmrc as user-trusted when running from $HOME

colinhacks:
## Problem

Running `nubx` / `nub dlx` (or any install) from your home directory printed spurious untrusted-source warnings about your own `~/.npmrc`:

```
WARN ignoring https-proxy from untrusted source Project: committed `.npmrc` cannot set registry proxies  code=WARN_NUB_UNTRUSTED_PROXY
WARN ignoring strict-ssl=false: Project source is not trusted (committed `.npmrc` cannot disable TLS validation)  code=WARN_NUB_UNTRUSTED_STRICT_SSL_DISABLE
```

`~/.npmrc` is the user's global npm config and should be user-trusted — but when `cwd == $HOME`, the loader read it **twice**: once as the trusted `User` source and once as an untrusted `Project` (committed-`.npmrc`) source, because `project_dir.join(".npmrc")` resolved to the same physical file. The untrusted copy tripped the subprocess-settings trust gate, so the user's own `https-proxy` and `strict-ssl=false` were refused (the warnings printed twice — the double-load tell).

## Fix

Vendored aube bump to `nub-fork` `a00e896`. Before reading `project_dir/.npmrc` as an untrusted `Project` source, skip it when its **canonical path equals the canonical user-rc path** (`$HOME/.npmrc`, or the `userconfig`/`NPM_CONFIG_USERCONFIG` override). A genuine committed repo `.npmrc` is a different file, so it is still read as untrusted — **the trust gate that blocks a malicious repo `.npmrc` from setting a proxy / disabling TLS is fully preserved.** Applied at every project-read site (incl. the project-pointed `npmrc-auth-file`).

## Verification

- Two regression tests in `aube-registry` (home-case deduped + nested-committed-stays-untrusted), both confirmed to fail without the load.rs fix.
- Empirical: `cd $HOME && nub dlx ...` → zero untrusted warnings (fixed); same `.npmrc` from a nested dir as a real committed file → warnings still fire (gate intact).
- `cargo clippy -p aube-registry --all-targets -- -D warnings` clean; changed files `cargo fmt`-clean.

Upstreamable to jdx/aube — standalone aube has the identical same-file double-read (latent-bug-fix class, default-preserving).


colinhacks:

PR #73, opened 2026-06-22
node: bare `nub node` prints status then help

colinhacks:
Bare `nub node` now follows the idiomatic command-group convention: it prints the resolved-Node status block, then the verb listing — made clearer by the `nub node which` subcommand. `nub node --help`/`-h`/`help` still prints help only (unchanged).

The shared help text is factored into a single `NODE_HELP` const so the bare form and the `--help` form never drift. Discovery on the bare path is best-effort: if no Node resolves, the help still prints rather than erroring.

New bare-`nub node` output:

```
node 26.2.0
  path      /Users/.../bin/node
  resolved  package.json#engines.node (>=22.15.0)

nub node — manage Node versions

Usage: nub node <command>

Commands:
  which                    print the resolved Node binary path (why → stderr)
  install [<version>...]   provision version(s) into nub's cache (bare: the project pin)
  ls                       list versions in nub's cache
  uninstall <version>      remove a version from nub's cache
  pin <version>            write the project's Node pin
```

Verified locally: `cargo build -p nub-cli --profile fast`, `cargo clippy --all-targets --all-features -- -D warnings` (clean), `cargo fmt --check` (clean), scoped integration tests green; dev binary confirmed for `nub node`, `nub node -h`, `nub node which`. Adds one integration test (`bare_node_prints_status_then_help`).

### DIFF
--- a/crates/nub-cli/src/cli.rs
+++ b/crates/nub-cli/src/cli.rs
@@ -5030,28 +5030,35 @@ fn run_node(args: &[String]) -> Result<i32> {
         anyhow::anyhow!("could not locate nub's cache directory (no $HOME / $XDG_CACHE_HOME)")
     })?;
 
+    // The verb listing — shared by `--help`/`-h`/`help` and the bare form, so the
+    // two never drift. Bare `nub node` prepends the resolved-Node status block.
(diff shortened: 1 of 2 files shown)


colinhacks:

Issue #71, opened 2026-06-22
Running `nub add` in workspace package, re-installs node modules due to `WARN_NUB_GVS_MODE_CHANGED`


Stupremee:
I'm currently trying to migrate an internal monorepo from pnpm + node + turborepo, to use `nub`.

For package management, I would like to keep `pnpm`, so I ran `nub pm use pnpm`.
In the workspace root I ran `nub install`, which works fine.

Now the problem is, if I run `nub add some-dep` inside any package of my monorepo, `nub` re-installs the whole node_modules with the following message:

```
WARN global virtual store disabled → enabled; removing /.../node_modules and reinstalling from scratch code=WARN_NUB_GVS_MODE_CHANGED
```

After the `nub add` command re-installed everything, when running `nub install` in the monorepo root again, I get the exact same error.

colinhacks:
Thanks for the detailed report — investigating now. The flip-flopping `WARN_NUB_GVS_MODE_CHANGED` between a workspace member (`nub add`) and the root (`nub install`) looks like the global-virtual-store mode is being resolved from the current directory rather than the workspace root. We'll get a fix up.

colinhacks:
Fixed in #74 (merged). Root cause: the GVS mode-change check predicted the *requested* mode (global virtual store on, off-CI) but the linker materializes per-project (GVS off) whenever `hoist` is true (the default) — so every manifest-mutating install (`nub add`) saw a phantom disabled→enabled transition and wiped node_modules, ping-ponging with the next install. The check now computes the *effective* mode (folding `hoist` + node-linker), so root and member agree and there's no spurious reinstall. Ships in the next release.

Stupremee:
Cool, thanks for the quick fix!

PR #74, opened 2026-06-22
fix(pm): stop spurious GVS reinstall when running `nub add` in a workspace member

colinhacks:
## Problem

In a pnpm-incumbent monorepo (`nub pm use pnpm`), `nub install` at the root works, but `nub add <dep>` inside any workspace member re-installs all of `node_modules` from scratch with:

```
WARN global virtual store disabled → enabled; removing /.../node_modules and reinstalling from scratch code=WARN_NUB_GVS_MODE_CHANGED
```

and it ping-pongs with the next root install. (Reported by an external user migrating a pnpm + turborepo monorepo to nub.)

## Root cause

The pre-link global-virtual-store **mode-change check** (`reset_on_mode_change`) compared the existing `.aube/`/`.nub/` tree against the *raw requested* GVS mode (`planned_gvs` = override, else `!CI` → `true` off-CI). But the linker forces **per-project** materialization whenever the hidden hoist tree is enabled (`hoist=true`, the default) or the layout is `hoisted` — both `Linker::link_all`/`link_workspace` fall back to `without_global_virtual_store()` in that case.

So on a default project the check predicted GVS-**on** while the linker always wrote GVS-**off** (per-project). Every install that bypassed the fast path — which `nub add` always does, since it mutates a manifest — saw a phantom `disabled → enabled` transition, wiped `node_modules`, and reinstalled. The next install saw the inverse, ping-ponging.

(Single-package repeat `nub install` masked the bug via the up-to-date fast-path short-circuit, which returns before the mode-change check.)

## Fix

Compute the **effective** GVS mode the same way the linker decides it — folding in `hoist` and `node-linker` via a shared `resolve_node_linker` (extracted from the link phase so the two stay in lockstep) — and feed that to `reset_on_mode_change`. Genuine mode changes (e.g. `hoist=false` GVS toggles, or switching to CI) still reset correctly.

Implemented in `vendor/aube` (`nub-fork` `3778dc8`, a default-preserving latent-bug fix); this PR bumps the pin.

## Verification

- New aube unit tests on the effective-mode truth table (`gvs::tests`, 4 cases).
- New aube e2e bats regression (`test/add.bats`): root install → `add` in a member → assert no `WARN_AUBE_GVS_MODE_CHANGED`, root reinstall stays warm. Confirmed it **fails** on the pre-fix code and passes after.
- Manual e2e against a pnpm-incumbent monorepo fixture: member `add` and root reinstall both clean (no warning, no wipe); legitimate `hoist=false` GVS on→off transition still triggers the reset.
- `cargo fmt --check` + `cargo clippy --all-targets -- -D warnings` clean on aube. (One pre-existing, unrelated `--save-peer` bats failure exists on the pin base — not touched here.)

Refs #71


colinhacks:

PR #75, opened 2026-06-22
tests: command×flag conformance harness (full CLI surface, differential vs reference PM)

colinhacks:
## What

A durable, data-driven harness that exercises nub's **full command × flag surface** against a real project and asserts each command behaves — exits cleanly where it should, fails *correctly* where it should, never leaks the embedded-engine brand, and (where parity is claimed) agrees with the reference package manager.

Lands under `tests/conformance/cmdflag/` (a sibling to the existing lockfile-conformance harness, which verifies round-trip fidelity — a different axis). This one verifies that every wired verb + its major flags actually *runs* in a real repo.

## Why

Shallow happy-path probing let a real-machine `nub audit` failure ship. The fix is a committed, exhaustive surface sweep run on a cadence — the differential-fixture methodology at scale.

## Contents

- `inventory.tsv` — 60 cells, enumerated authoritatively from `crates/nub-cli/src/cli.rs` (the clap `Command` enum + `NodeCommand`) and `crates/nub-cli/src/pm_engine/mod.rs` (`ENGINE_VERBS`, restricted to the WIRED match arms in each `*_family.rs`). Stubbed verbs are intentionally omitted.
- `run.sh` — hermetic runner: sandbox `HOME`/`XDG_*`, `USER_NPMRC` seed (covers the custom `registry=` ~/.npmrc condition), an `nub install`-primed read-only copy for query verbs, fresh throwaway copies for `mut`/`net` cells so the fixture is never dirtied, a reference-PM parity hook (`REF=1`), and an `aube` brand-leak canary.
- `expectations.txt` — fixture-tuned known-non-zero / red list; an `XPASS-STALE` flag forces it to shrink when a cell's outcome changes.
- `README.md` — the loop, the inventory schema, the diverse repo fixture-set plan for the fan-out.

## Verification

- No-NET deterministic core sweep on a fresh **zod** clone (`@912f0f51`, pnpm-10 workspace monorepo): **RESULT: OK** (green).
- Full `NET=1` sweep with a custom-registry `~/.npmrc` surfaced two real nub findings (triaged separately, not fixed here — recommend-only):
  - `run <missing> --if-present` exits 1; pnpm/npm exit 0 (parity divergence).
  - `nub agent skill` output contains the literal string "aube" (brand leak in public agent-facing copy).
- No Rust changes — test tooling only.

Refs the conformance-matrix effort.

### DIFF
--- a/tests/conformance/cmdflag/README.md
+++ b/tests/conformance/cmdflag/README.md
@@ -1,0 +1,96 @@
+# Command×flag conformance harness
+
+Exercise nub's **full CLI command × flag surface** against a real project and
+assert each command behaves: exits cleanly where it should, fails *correctly*
+where it should, never leaks the `aube` brand, and — where parity is claimed —
+agrees with the reference package manager.
+
+This is a **different axis** from the lockfile harness one level up
+(`tests/conformance/run.sh`), which verifies lockfile round-trip fidelity. This
+one verifies that every wired verb + its major flags actually *run* in a real
+repo. It exists because shallow happy-path probing let `nub audit` ship a
+real-machine failure (it failed under a normal `~/.npmrc` carrying a custom
+`registry=`). The fix is a durable, exhaustive surface sweep run on a cadence.
+
+## Loop
+
+For a given nub binary + a real project fixture, the runner:
+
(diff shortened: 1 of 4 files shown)


colinhacks:

PR #82, opened 2026-06-23
fix(pm): bun.lock writer byte-identity — multi-line named catalogs + collapse empty packages

colinhacks:
Two pure-rendering drifts in the bun.lock writer (`format_bun_lockfile`, `vendor/aube/crates/aube-lockfile/src/bun/write.rs`) diverged from bun's native output. In both cases the data round-trips correctly through parse; only the emitted bytes differed from what `bun install` would produce — a gratuitous diff against bun's own lockfile.

- **R-1 — named `catalogs` rendered inline.** The per-catalog inner object was emitted as `{ "k": v }`. bun renders it multi-line — one dep per indented line with a trailing comma (`.repos/bun/src/install/lockfile/bun.lock.zig:349-380`).
- **R-2 — empty `packages` block.** Was emitted as `"packages": {\n  }`. bun collapses an empty block to `"packages": {}` via its `first` guard (`bun.lock.zig:388,675-682`). A dependency-free root triggered this.

Adds two byte-identity regression tests (`test_write_byte_identical_named_catalogs_multiline`, `test_write_byte_identical_empty_packages_collapses`) that parse a hand-authored bun.lock and assert the re-written bytes are identical to bun's form.

Re-lands aube fork commit `b4d0dc0` (nubjs/aube `nub-fork-bunlock-r1r2`) in the new plain-vendored aube model — an ordinary nub PR touching `vendor/aube/**`, no submodule / pin bump. The old fork branch is now historical.

## Verification
- `cargo test -p aube-lockfile` (from `vendor/aube`): 419 unit tests + proptest round-trips pass, including both new byte-identity tests.
- `cargo clippy -p aube-lockfile --all-targets --all-features -- -D warnings`: clean.
- `rustfmt --edition 2024 --check` on the two changed files: clean.
- `cargo check -p nub-cli` (nub workspace, aube as path dep): builds.

### DIFF
--- a/vendor/aube/crates/aube-lockfile/src/bun/tests.rs
+++ b/vendor/aube/crates/aube-lockfile/src/bun/tests.rs
@@ -565,14 +565,11 @@ fn test_write_byte_identical_to_native_bun() {
 /// top-level metadata blocks, so a wrong block order slipped through.
 ///
 /// Scope note: this asserts byte-identity for the four blocks bun and
-/// nub render identically. Named `catalogs` (object-of-objects) and an
-/// EMPTY `packages` block are deliberately excluded — nub currently
-/// renders a named catalog's inner object inline (bun renders it
-/// multi-line) and emits `"packages": {\n  }` for empty (bun emits
-/// `{}`). Those two rendering drifts are real but out of this lane's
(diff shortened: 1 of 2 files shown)


colinhacks:

PR #85, opened 2026-06-23
perf(native): stamp transpile-cache build-id at compile time

jdalton:
## What

The transpile cache key folded in `exe_hash()` — `current_exe()` + `fs::read` + `blake3`. Because `nub-native` is a cdylib loaded into the host Node process, `current_exe()` resolves to the **Node** binary (~100 MB+), so every process's first transpile read and hashed it *before* the cache-hit check. It also never changed when nub itself was rebuilt, so the dev-rebuild cache invalidation the old comment describes never actually fired.

## Change

- `build.rs` stamps `NUB_NATIVE_BUILD_ID` (git short SHA, `-dirty` suffix for an uncommitted working tree, empty string when git is unavailable) with zero runtime I/O; reproducible at a fixed clean commit, so a release's rebuilds reuse the cache.
- `cache.rs` folds the const into the key at the same byte position/separators and removes `exe_hash()`.
- `CACHE_SCHEMA` 4 → 5 so prior on-disk entries are cleanly ignored — schema is hashed into the key preimage, so filenames are disjoint and an old entry is a miss, never mis-read.

## Effect

Removes the per-process ~100 MB binary read+hash from the warm transpile path. Measured warm-cache `nub hello.ts`: ~55–68 ms faster (User-CPU and min wall-clock; the eliminated work is an I/O-bound read, so the mean is noisier under load). No transpile-parity regressions; `cargo build --release` clean.

### DIFF
--- a/crates/nub-native/build.rs
+++ b/crates/nub-native/build.rs
@@ -32,10 +32,7 @@ fn main() {
         println!("cargo:rerun-if-changed={}", git_dir.join("HEAD").display());
         println!("cargo:rerun-if-changed={}", git_dir.join("index").display());
         // Packed/loose ref updates (e.g. a branch move) also shift the short SHA.
-        println!(
-            "cargo:rerun-if-changed={}",
-            git_dir.join("refs").display()
-        );
+        println!("cargo:rerun-if-changed={}", git_dir.join("refs").display());
         println!(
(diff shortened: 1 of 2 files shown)


colinhacks:
Yikes. Good catch! Huge win.

colinhacks:

PR #87, opened 2026-06-23
test(native): extract nub-cache-key crate to lock cache-key invalidation

colinhacks:
Locks the transpile cache-key invalidation contract that #85 introduced (the compile-time build-id + `CACHE_SCHEMA` folded into the key), which previously had no test.

`nub-native` is a `cdylib` N-API addon with `test = false` — a `cargo test` harness can't link the `napi_*` symbols (resolved by the Node host at dlopen). So the pure key derivation (the blake3 preimage) is factored into a new napi-free `nub-cache-key` crate that `nub-native` calls into, with the tests living there:

- `cache_key_changes_when_build_id_changes` — a changed build-id yields a different key (the point of #85's build-id stamp).
- `cache_key_namespaced_by_schema` — the schema is hashed into the key, so the 4→5 bump makes the two eras' filenames disjoint.
- `cache_key_is_stable_for_identical_inputs` — determinism.

Behavior is byte-identical: `cache_key` delegates with the same `NUB_VERSION`/`CACHE_SCHEMA`/`BUILD_ID` consts and an identical preimage. `nub-native` keeps `test = false`, so the whole-workspace `cargo test` (including the Windows CI leg) still never builds the cdylib's harness — the new tests run on every platform via the standalone crate.

Refs #85.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3042,8 +3042,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "nub-cache-key"
+version = "0.1.12"
+dependencies = [
+ "blake3",
+]
+
+[[package]]
 name = "nub-cli"
-version = "0.1.11"
+version = "0.1.12"
 dependencies = [
  "anyhow",
  "aube",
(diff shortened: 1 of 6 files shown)


colinhacks:
Shipped in v0.1.13: https://github.com/nubjs/nub/releases/tag/v0.1.13

PR #97, opened 2026-06-23
feat(add): mirror npm save-prefix for bare-version add under npm incumbent

colinhacks:
## What

When the project's incumbent package manager is **npm**, `nub add <pkg>@<exact>` (or a bare `nub add <pkg>` that resolves to a concrete version) now writes the saved specifier with npm's default save-prefix applied (`"^1.2.3"`), matching real `npm install`. pnpm/bun/nub-identity projects keep literal-preserve (`"1.2.3"`), matching real pnpm/bun. An explicit range (`^1`, `~1.2`, `>=1`) is preserved verbatim on every PM.

This is the npm-incumbent behavior nub users expect (Option A, decided by the maintainer).

## How

- **`vendor/aube` EngineContext**: new per-invocation posture `npm_save_prefix_on_bare_exact` (default `false` = upstream-neutral; standalone aube byte-for-byte unaffected). Set `true` only under an npm compat surface.
- **`vendor/aube` add manifest writer**: a bare *exact* version (detected via `Version::parse` on the user's explicit range) is written with the resolved save-prefix when the posture is on. Explicit ranges always fall through to verbatim preservation.
- **`vendor/aube` settings**: new `saveExact` knob so an existing `.npmrc save-exact=true` (or env) pins the exact version on add — npm and pnpm both read this knob; previously only the CLI `--save-exact` flag took effect.
- **nub `pm_engine`**: sets the posture `true` only for `ConfigSurface::NonPnpmCompat { role: "npm" }`.

The resolved save-prefix is honored, so `.npmrc save-prefix=~` yields `"~1.2.3"`; `save-exact=true` yields the exact pin `"1.2.3"`.

## Mutation harness

Un-gates `m1-add-noconflict npm` + `m3-add-dedup npm` (removed from `tests/mutation/expected-failures.txt`). The full differential harness is now **9/9 PASS** (npm legs produce package.json specifiers byte-equivalent to real npm; pnpm/bun literal-preserve legs unchanged).

## Verification

- `cargo clippy --all-targets --all-features -- -D warnings` and `cargo fmt --check`: clean.
- aube unit tests: `aube-settings` 61 passed, `aube` add 71 passed.
- Mutation harness (real npm 11.13.0 / pnpm 10.15.1 / bun 1.3.14): 9/9 PASS.
- e2e parity vs real npm: `ms@2.1.3`→`^2.1.3`, `~2.1.0`→`~2.1.0` (preserved), bare `ms`→`^2.1.3`, `save-exact=true`→`2.1.3`, `save-prefix=~`→`~2.1.3`, `--save-exact`→`2.1.3`; pnpm/bun/nub-identity all keep `2.1.3`.

Refs: npm-add-save-prefix-convention thread.

### DIFF
--- a/crates/nub-cli/src/pm_engine/mod.rs
+++ b/crates/nub-cli/src/pm_engine/mod.rs
@@ -1316,6 +1316,11 @@ pub(crate) fn engine_brand_preflight() {
     // incumbent; under any other surface those Bun-named env vars are another
     // tool's state and must not be read (name-based policy, like `read_yarn_config`).
     let read_bun_config = read_bun_config_for_surface(&surface);
+    // npm save-prefix convention: only under an npm incumbent does a bare-exact
+    // `add pkg@1.2.3` get npm's `^` save-prefix (`"^1.2.3"`). pnpm/bun/nub-identity
+    // preserve the literal bare version — matching each PM's real behavior.
+    let npm_save_prefix_on_bare_exact =
+        matches!(surface, ConfigSurface::NonPnpmCompat { role: "npm", .. });
     aube_util::update_engine_context(|c| {
         c.read_branded_pnpm_config = read_branded_pnpm_config;
         // GLOBAL config is read PM-AGNOSTICALLY and UNGATED by cwd incumbency:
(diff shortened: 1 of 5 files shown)


colinhacks:
Shipped in v0.1.14: https://github.com/nubjs/nub/releases/tag/v0.1.14

PR #98, opened 2026-06-23
fix(runtime): recover from async-loader resolveSync stub on the fast tier (Turbopack/Tailwind crash)

colinhacks:
## Problem

`nub run build` crashes Turbopack in real Next.js apps that use **Tailwind v4**. The build aborts evaluating `app/globals.css` with:

```
Error evaluating Node.js code
Error: The resolveSync() method is not implemented
  at Hooks.resolveSync (node:internal/modules/esm/hooks)
  at #resolveAndMaybeBlockOnLoaderThread (node:internal/modules/esm/loader)
  at resolve (.../nub/runtime/preload-common.cjs:281)   ← nub's sync resolve hook
  at ModuleLoader.resolveSync -> ModuleLoader.resolve
```
followed by an esbuild `all goroutines are asleep - deadlock`.

## Root cause

A hook-tier interaction between nub's sync hooks and a user async loader:

1. On the fast tier (Node 22.15+) nub registers **sync** resolve/load hooks via `module.registerHooks`.
2. Tailwind v4's `@tailwindcss/node` calls `module.register(esm-cache.loader.mjs)` — an **async** ESM loader — when Turbopack evaluates `globals.css`.
3. Once sync hooks exist, Node routes **every** resolution (even `import`) through the synchronous chain. nub's sync `resolve` hook calls `nextResolve`, which reaches Node's default step `#resolveAndMaybeBlockOnLoaderThread`; with the async loader's `#customizations` set, that calls the async-hooks `Hooks.resolveSync()` — a stub that unconditionally throws `ERR_METHOD_NOT_IMPLEMENTED('resolveSync()')`. An async loader cannot service a sync resolve.

Verified against Node v24.3.0 source and a `module.register` spy on the real failing build. A fresh `create-next-app` (no Tailwind) does NOT reproduce — it never registers the async loader — which is why an earlier fresh-scaffold pass read as clean.

## Fix

nub's sync `resolve` hook now wraps `nextResolve` in try/catch. On `ERR_METHOD_NOT_IMPLEMENTED` **and** when a user async loader is active, it recovers by resolving the specifier through the parent's CommonJS resolver and short-circuiting. Narrow guard — only the impossible sync-into-async hop is bypassed; the user's async loader still runs for everything else. Fast-tier-only (the compat tier uses a separate worker-based resolve path that doesn't hit the sync stub). No Rust change.

## Verification

- Real Next 16 + Tailwind v4 + Turbopack app: pre-fix `nub run build` EXIT 1 (crash) → **post-fix EXIT 0**, full route table, no `resolveSync` error, no esbuild deadlock.
- Augmentation intact on normal runs (TS transpile, `import.meta.resolve`) — the recovery branch never fires without an async loader present.
- Adds an `#[ignore]`d real Next16+Tailwind4 Turbopack e2e regression test.


colinhacks:
Shipped in v0.1.14: https://github.com/nubjs/nub/releases/tag/v0.1.14

PR #99, opened 2026-06-23
worker: spec-faithful Worker — inline/Blob/data: workers, name, error location, classic mode

colinhacks:
Polishes the browser-shape `Worker` polyfill toward the WHATWG Web Workers spec (and Cloudflare semantics where applicable) — not Bun parity. Implemented over `node:worker_threads` via the existing preload; additive, feature-detected, brand-clean.

**HELD for the v0.3 minor release — do not merge yet.**

## Landed (all spec-correct)

- **Inline workers via `data:` and `blob:` URLs** — the WHATWG inline mechanisms, neither of which `node:worker_threads` exposes through a browser-shape constructor. `data:` runs directly; `blob:` (from `URL.createObjectURL`) captures the Blob source synchronously at mint time (new `runtime/worker-blob-url.cjs`, installed eagerly on the main thread so it is live before the URL is created — cold-start clean, pulls zero new builtins; shares its source registry with the lazily-loaded `worker-polyfill.mjs` via `createRequire` path-dedup).
- **`{name}` option** → `worker.name` (main) and `self.name` (worker), backed by `worker_threads` `name`/`threadName`. Previously dropped.
- **Real `ErrorEvent` source location** — `filename`/`lineno`/`colno` populated from the worker error's stack (were hardcoded empty/zero).
- **`execArgv` merge** — a user-supplied `execArgv` is merged with nub's preload-carrying parent flags instead of being clobbered (parent first, user wins); still strips `--harmony*`/`--experimental-shadow-realm`.
- **`terminate()` returns `undefined`** (the spec shape; Node's `Promise` is discarded).
- **Classic vs module workers** — `type: "classic"` exposes `importScripts(...urls)` (synchronous load + eval in scope); module workers throw on `importScripts`.
- **`messageerror`** is now a `MessageEvent` with `data: null` (dropped the non-spec `.data` carrying the error).
- **Removed the fabricated `exit` event** — not a web Worker event (it is a `node:worker_threads` concept).
- TS/JSX zero-build workers keep working (nub's transpile augmentation, unchanged).
- Docs (`site/content/docs/runtime/workers.mdx`) and `@nubjs/types` (`worker.name`) updated.

## Deferred / not added

- **`ref()` / `unref()`** — not added (non-spec Node/Bun extension).
- **`parallel.map` / worker-pool** — out of scope (neither spec nor Cloudflare; a separate new API-surface effort).

## Tests

Contract-named integration tests (`crates/nub-cli/tests/integration.rs`) plus fixtures: name (both sides), `data:` worker, `blob:` worker, `File instanceof Blob` regression, ErrorEvent source-location, no-`exit`-event, transferable detach, MessageChannel/MessagePort roundtrip, BroadcastChannel deliver-to-other-not-self, classic `importScripts`, module-worker `importScripts`-throws.

## Verification

- 14/14 worker integration tests green.
- Verified end-to-end on host Node 26.2, the 18.19 compat-tier floor, and 20.19.
- Two fresh-context adversarial self-reviews: review #1 surfaced a `File instanceof Blob` regression and an `ErrorEvent.filename` stack-prefix bug (both fixed + regression-tested); review #2 on the fix delta came back clean (prototype re-pointing verified across 18.19/20.19/22.15/26.2, duck-typed `FormData`/`Response` consumers survive the Blob wrap).

## Note for the maintainer

nub's **default worker `type` is `"module"`** (documented), a deliberate divergence from the WHATWG `"classic"` default — it aligns with nub's ESM-first main-entry posture. The added test pins this existing behavior; it does not decide the spec-default question, which is yours to make.

### DIFF
--- a/crates/nub-cli/tests/integration.rs
+++ b/crates/nub-cli/tests/integration.rs
@@ -2291,6 +2291,24 @@ fn worker_from_blob_url() {
 }
 
 #[test]
+fn blob_wrapping_preserves_file_instanceof_blob() {
+    // Regression: nub subclasses globalThis.Blob to support blob: workers. `File`
+    // is a bootstrap global extending the NATIVE Blob, so the swap must re-point
+    // File's prototype chain or `new File(...) instanceof Blob` silently turns
+    // false — an additivity violation vs vanilla Node (File IS-A Blob).
+    let (stdout, stderr, code) = run_nub("worker", "blob-file-instanceof.ts");
+    assert_eq!(code, 0, "blob/File fixture should run: {stderr}\n{stdout}");
+    assert!(
+        stdout.contains("file-instanceof-blob:true"),
+        "File must remain an instanceof Blob after the Blob wrap (matches plain Node): {stdout}"
+    );
+    assert!(
+        stdout.contains("blob-instanceof-blob:true") && stdout.contains("blob-size:5"),
+        "a Blob made through the wrapper must keep instanceof + the full Blob API: {stdout}"
+    );
+}
+
+#[test]
 fn worker_error_event_carries_source_location() {
     // WHATWG ErrorEvent must carry filename/lineno/colno from where the error was
     // raised. nub fills these from the worker error's stack (they were hardcoded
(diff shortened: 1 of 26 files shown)


colinhacks:

PR #100, opened 2026-06-23
test(ci): Node-version matrix smoke for the augmentation surface

colinhacks:
## What

Adds a CI workflow + fixtures that exercise nub's **version-banded augmentation** across a broad Node-version matrix — closing the gap that let the Turbopack/Tailwind `resolveSync()` crash ship uncaught. nub augments the user's Node through mechanisms that break **differently per Node version** (sync `module.registerHooks` fast tier ≥22.15, async `module.register` loader-worker compat tier 18.19–22.14, flag injection); the main CI matrix tested only a handful of versions and never set up the async-loader × sync-hooks collision.

## Scenarios

- **functional smoke** (every version): `hello.js`, `hello.ts` (transpile + non-erasable enum), ESM↔CJS, `import.meta.resolve`, Worker threads.
- **async-loader collision** (`tests/node-matrix/fixtures/async-loader-collision/`): a **bundler-free** reproduction of the `resolveSync()` stub crash — a bare `module.register(<async loader>)` while nub's fast-tier sync resolve hook is active. Verified to reproduce **identically to the real Turbopack build** through the actual nub binary.
- **real-app builds** (nightly / augmentation-surface PRs): **Next 16 + Tailwind v4 + Turbopack** (the in-the-wild manifestation) and a **Vite 6 SSR** build.

## Cost model (CI is free; optimize runner slots + latency)

- nub is built **once** and reused across every Node leg via an artifact — the Node version is a PATH/runtime choice (`actions/setup-node`), not a rebuild.
- per-PR: a lean tier-boundary subset of the lightweight smoke, gated by **path filters** (the full sweep + real apps run only when `runtime/`, `crates/nub-core/src/node/`, `crates/nub-cli/src/`, or this workflow change).
- push:main + nightly: the full version sweep + both real apps.
- `concurrency` group with `cancel-in-progress`; one always-green `node-matrix-gate` aggregator for branch protection.

## Validation (the non-negotiable proof)

Validated against a dev `nub` built from current `main` (pre-#98), toggling #98's `runtime/preload-common.cjs` fix in place, across nvm Node versions:

| Node | nub state | collision / real Next build |
|---|---|---|
| 22.16, 24.11 (Node-broken) | pre-#98 | **RED** — `resolveSync()` crash reproduced |
| 22.16, 24.11 (Node-broken) | **with #98 fix** | **STILL RED** — #98 does not recover the crash on broken bands |
| 24.12, 26 (Node-fixed) | either | green (trivially — Node has no bug there) |
| 18.19–22.12 (compat tier) | — | green (no sync hook → no stub) |

→ The matrix reproduces the bug on the affected versions and stays honest on the rest. Functional smoke passes on every tier; the Vite SSR build passes on every version.

## Note for review (the matrix already earned its keep)

Building this surfaced that **#98's fix does not actually resolve the crash on the Node-broken bands** — its `userAsyncLoaderActive()` gate is `false` at the throwing resolution (the loader module's own spec), so the recovery never fires; and the load hook has the same un-recovered `loadSync` stub. #98's "post-fix EXIT 0" was a false positive from validating on a Node-level-fixed host. So the collision guard **XFAILs on Node-broken legs** for now (documented in `tests/node-matrix/README.md`) and **must-pass** on Node-fixed / compat-tier legs; flip the broken legs to `--collision-must-pass` once nub recovers the sync-into-async hop. This matrix would have blocked #98 as a non-fix.

Refs #98

### DIFF
--- a/.github/workflows/node-matrix.yml
+++ b/.github/workflows/node-matrix.yml
@@ -1,0 +1,207 @@
+name: Node matrix
+
+# Augmentation smoke across a BIG Node-version matrix. nub augments the user's installed
+# Node via version-banded mechanisms (sync `module.registerHooks` fast tier >=22.15, async
+# `module.register` loader-worker compat tier 18.19-22.14, V8/experimental-flag injection)
+# that break DIFFERENTLY per Node version. The repo's main CI matrix tests only a handful
+# of versions and never exercised the async-loader x sync-hooks collision that shipped the
+# Turbopack/Tailwind `resolveSync()` crash (the nextjs-build-compat thread / PR #98) — this
+# workflow closes that gap.
+#
+# COST MODEL: CI is free (public repo); the real constraints are runner-SLOT contention and
+# dev-feedback latency. So:
+#   - nub is built ONCE (one job) and reused across every Node leg — the Node version is a
+#     PATH/runtime choice (actions/setup-node), NOT a rebuild.
+#   - per-PR: a LEAN tier-boundary subset of the lightweight smoke, gated by PATH FILTERS
+#     (the heavy real-app matrix and the broad version sweep run only when runtime/preload/
+#     spawn/flags code changes — the surfaces that can break version-banded augmentation).
+#   - push:main + nightly: the FULL version matrix + both real-app builds.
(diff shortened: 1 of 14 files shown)


colinhacks:

PR #17, opened 2026-06-18
perf: release-profile tuning, mimalloc, FxHashMap, and a bench harness

jdalton:
Low-risk performance and build changes, plus a benchmark harness to measure them.

## Changes
- **bench**: add a criterion harness (`cargo bench -p nub-core`) covering the cache-hash and workspace-filter/topo-sort hot paths, with a realistic TS fixture. Baseline-only; documents the transpile-bench gap (the oxc transform can't be benched in-process from the napi cdylib).
- **build**: tune `[profile.release]` — `lto = "thin"`, `codegen-units = 1`, `strip = true`. `panic` is left at the default `unwind` because `nub-native` is a cdylib loaded into Node's process and `panic` is a profile-global setting.
- **perf(cli)**: use mimalloc as the global allocator for the `nub` binary only (not the napi cdylib).
- **perf(core)**: use `FxHashMap` for internal, trusted-key workspace and registry maps (`workspace/filter.rs` `name_to_idx` + Kahn's `in_degree`, `pm/registry.rs` prefix grouping).

## Measured
On the workspace-filter benches (200-package synthetic graph):
- `workspace/topological_chunks/200`: 623µs → 311µs
- `workspace/build_dep_graph/200`: 55µs → 47µs

The release-profile and allocator changes do not appear in `cargo bench` (benches build in release already, and mimalloc governs only the `nub` binary) — they land in real binary startup/runtime.

## Notes
- Techniques drawn from surveying sibling Rust projects' optimization work.
- `env.rs`/`scripts.rs` env maps were left on std `HashMap`: they sit in public nub-core API signatures, are small, and are off the hot path.

Issue #11, opened 2026-06-18
pnpm lockfile round-trip broken around workspace peerDependencies (both directions reject)


AlexGodard:
nub's pnpm-compat lockfile round-trip is broken around a workspace library's **`peerDependencies`**, in **both** directions:

1. `nub install --frozen-lockfile` **refuses a valid pnpm-authored v9 lockfile**, claiming drift in a package whose only change is peer deps.
2. After `nub install --no-frozen-lockfile` rewrites the lockfile, **`pnpm install --frozen-lockfile` then rejects nub's output** — so the rewritten lockfile is not pnpm-compatible, contradicting the "reads and writes pnpm's own `pnpm-lock.yaml` ... accepted by pnpm 9+" claim.

This matters because a project may keep `packageManager: pnpm@…` (Vercel etc. still install with pnpm): if nub rewrites the lockfile, pnpm's frozen install on CI/deploy breaks; and if it doesn't, nub's own frozen install breaks.

## Setup

pnpm workspace, lockfile v9, `packageManager: pnpm@11.5.2`. One workspace library declares peer deps:

```jsonc
// packages/ui/package.json
{
  "peerDependencies": {
    "next": ">=16.0.0",
    "react": ">=19.0.0",
    "react-dom": ">=19.0.0"
  }
}
```

The committed `pnpm-lock.yaml` is validated by pnpm:

```
$ pnpm install --frozen-lockfile
Scope: all 11 workspace projects
Already up to date
Done in 248ms using pnpm v11.5.2
```

## Direction 1 — nub refuses pnpm's lockfile

```
$ nub install --frozen-lockfile
ERR_NUB_OUTDATED_LOCKFILE
  × lockfile is out of date with package.json: packages/ui: manifest removed next
  │ help: run without --frozen-lockfile to update the lockfile, ...
```

"manifest removed next" is wrong — `next` is a `peerDependency` of `packages/ui`, present in both the manifest and the lockfile; pnpm considers the lockfile up to date. nub's staleness comparison appears to treat the peer specifier as a removed dependency.

## Direction 2 — pnpm refuses nub's lockfile

```
$ nub install --no-frozen-lockfile         # rewrites pnpm-lock.yaml (~1165 +/ 2392 - on a ~6.8k-line file)
$ grep lockfileVersion pnpm-lock.yaml
lockfileVersion: '9.0'

$ pnpm install --frozen-lockfile
[...]

PR #104, opened 2026-06-24
Audit pnpm v11.9.0: fix inherited credential leak + multi-version minimumReleaseAgeExclude

colinhacks:
Audit of pnpm v11.9.0's 17 changelog items against nub's package manager (the vendored `aube` engine + `crates/nub-cli/src/pm_engine`). For each item: does nub inherit the bug, and is the enhancement worth adopting? Two inherited defects are fixed here; the rest are N/A or maintainer-gated (surfaced below, not landed).

## Per-item disposition

| # | pnpm v11.9.0 item | Disposition | Notes |
|---|-------------------|-------------|-------|
| 1 | Tarball integrity compute for checksumless registries | Recommend-only (C) | Round-trip already verbatim; hard-fail already matches. Computing integrity for checksumless registries changes emitted lockfile content + `strict-store-integrity` semantics — maintainer call. |
| 2 | `sbom --exclude-peers` | N/A | `sbom` is stubbed in nub. |
| 3 | `audit --fix` combined `minimumReleaseAgeExclude` | N/A | nub `audit --fix` writes overrides only; no writer for that setting yet. |
| 4 | Non-deterministic peer resolution | N/A | aube resolution is structurally deterministic (single-threaded FIFO BFS, all output `BTreeMap`s, sorted fallback). |
| 5 | Windows dlx EBUSY / MAX_PATH | N/A (design-note) | dlx uses an ephemeral tempdir; no persistent cache yet. Routed to the dlx-cache design. |
| 6 | Hang on non-retryable network error | N/A | aube's retry loop is bounded and always terminates; no hang. |
| 7 | Audit perf via Tarjan SCC | N/A (design-note) | No vuln-tree prune in nub's audit yet. Routed to the audit design. |
| 8 | Optional-dep update rewriting unrelated specs | N/A | Manifest write happens after a successful resolve, on a clone; each key rewrites from its own resolution. |
| 9 | GVS enable replaces stale symlinks | N/A (minor gap) | GVS implemented; stale-hoisted-symlink replacement on toggle absent. Low risk; routed to GVS-correctness. |
| 10 | `install --ignore-workspace` clobbers `allowBuilds` | N/A | No `--ignore-workspace` flag in nub; all config writes are read-modify-write (structurally immune). |
| **11** | **`minimumReleaseAgeExclude` multiple exact-version entries** | **Fixed** | Was name-only; version-pinned and `\|\|`-union entries were silently dropped. |
| 12 | Metadata cache registry-in-key | N/A | Cache key already includes the registry (CVE-2018-7167 class already handled). |
| 13 | `patch` dropped name for git single-version | N/A | Patch key is unconditionally `name@version`. |
| 14 | pnpr resolver endpoints | N/A | pnpm-internal service; no nub equivalent. |
| 15 | sbom CycloneDX issue-tracker refs | N/A | sbom stubbed. |
| 16 | `@pnpm/resolving.tarball-url` | N/A | pnpm-internal JS package; no nub equivalent. |
| **17** | **Registry-fetch failure leaks credentials** | **Fixed** | Error display surfaced the full credentialed registry URL. |

### #17 — Redact credentials from registry HTTP errors (security)
`aube-registry`'s `Error::Http` rendered reqwest's `Display`, which appends the full request URL via `url::Url`'s `Display` — including `user:pass@` userinfo and `?token=`/`_auth` query auth, unredacted. Command and install mappers format that into user-facing reports. aube already had `aube_util::url::redact_url` (used on the tarball path) but not on the metadata / error path.

- `aube-registry/src/lib.rs`: `Error::Http` now renders through `redact_http_error`, substituting `redact_url(e.url())` for the raw URL (redacted host preserved for debuggability).
- `aube/src/commands/install/fetch.rs`: wrapped a directly-interpolated tarball URL in `redact_url`.
- `crates/nub-cli/src/pm_engine/present.rs`: defense-in-depth `redact_credentials` pass at nub's output choke point, scrubbing any URL-like token through `redact_url` so a future un-redacted engine path cannot leak through nub.
- Tests: engine-level (builds a real connect error to `https://alice:s3cr3t@.../foo?token=abc123`, asserts the rendered text omits `s3cr3t`/`alice:`/`abc123` but keeps the host) + two `present.rs` unit tests.

### #11 — Multi-version `minimumReleaseAgeExclude` entries
`minimumReleaseAgeExclude` was name-only: a version-pinned entry like `axios@0.21.1` was stored as a literal string and compared against the bare package name, so it never matched and the age gate still fired. A rich union parser (`TrustExcludeRules`, supporting `name@v1 || v2`) already existed but was only wired to `trustPolicyExclude`.

- Unified `minimumReleaseAgeExclude` onto `TrustExcludeRules` (`aube-resolver/src/types.rs`, `trust.rs`, `settings.rs` via `parse_lossy` so a malformed entry doesn't drop the whole list).
- Per-version exclusion threaded through `pick_version`/`passes_cutoff` (`semver_util.rs`, `resolve/driver.rs`): name-only rules skip the gate for the whole package (original behavior preserved exactly); version-pinned rules exempt only the listed candidate versions.
- Tests: `name@v1 || v2` excludes both listed versions but not a third; name-only still excludes all versions.

Both fixes are default-preserving for standalone aube and cleanly upstreamable to `jdx/aube`.

## Verification
- `cargo clippy --all-targets --all-features -- -D warnings` (nub workspace) — clean
- `cargo fmt --check` — clean
- `cargo test -p aube-resolver` — 246 passed (2 new)
- `cargo test -p aube-registry` — 270 passed (1 new)
- HEAD compiles (`cargo build -p nub-cli`)
- A fresh-context adversarial review of the full diff returned clean (no blockers).

## Maintainer-gated recommendations (NOT landed — your call)
- **#1 — checksumless-registry integrity compute.** aube has the primitive (`aube_store::sha512_integrity`) and uses it on git/remote/local paths, but not for checksumless registry packages — they get no `integrity` in the lockfile. Adopting pnpm's behavior (compute from the downloaded tarball) would change emitted-lockfile content and interact with `strict-store-integrity`, so it's a behavior/default decision. Low incidence.
- **#6 (optional polish).** aube terminates on cert errors (no hang) but retries them twice first; classifying deterministic TLS errors as non-retryable to fail fast is minor polish.
- **Engine-source hardening for publish/login/unpublish error sites.** Those raw-URL error paths are covered by the new `present.rs` backstop at nub's boundary, but redacting them at the aube source too is a clean upstream win — separate scoped follow-up.

## Upstreamable to jdx/aube
The `aube-registry` redaction hunk (#17) and the entire `aube-resolver`/`aube` portion of #11 are default-preserving for standalone aube — both upstreamable as-is via `format-patch --relative=vendor/aube`. The `present.rs` backstop is nub-only.

### DIFF
--- a/crates/nub-cli/src/pm_engine/present.rs
+++ b/crates/nub-cli/src/pm_engine/present.rs
@@ -196,7 +196,8 @@ const RUNTIME_HINT_VOCAB: &[(&str, &str)] = &[(
 /// codes (so the `AUBE` inside `ERR_AUBE_*` never reaches the word pass), then
 /// per-line URL stripping, then the word-boundary `aube` → `nub` pass.
 pub(crate) fn rewrite(text: &str) -> String {
-    let mut text = text.to_string();
+    let text = redact_credentials(text);
+    let mut text = text;
     for (from, to) in RUNTIME_HINT_VOCAB {
         text = text.replace(from, to);
     }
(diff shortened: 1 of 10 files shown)


colinhacks:

PR #108, opened 2026-06-24
docker: official nub images (node:26 slim + alpine) + publish workflow

colinhacks:
Adds official, Bun-style Docker images for nub and a publish workflow, opened as one PR. Hardened to OCI/container best practices after a Bun-image audit.

## What's here

- **`docker/Dockerfile.slim`** — default variant, `FROM node:26-slim` (glibc, digest-pinned). Installs nub via `npm install -g @nubjs/nub`.
- **`docker/Dockerfile.alpine`** — `FROM node:26-alpine` (musl, digest-pinned). Adds `libgcc` + `libstdc++` for the native N-API addon.
- **`docker/docker-entrypoint.sh`** — bun-style arg dispatch (`docker run img script.ts` → `nub script.ts`; `docker run img sh` → shell), behind `tini` as PID 1.
- **`docker/smoke.sh`** — in-image smoke: `nub --version`, a TS run, a PM install + `require()`-load.
- **`.github/workflows/docker.yml`** — build + smoke BOTH variants on PR/main; multi-arch (`linux/amd64` + `linux/arm64`) push with provenance + SBOM on a release tag, gated behind a parameterized, maintainer-owned registry var (no publish target hardcoded). SHA-pinned actions; `permissions: packages: write` for GHCR.
- **`docker/PUBLISHING.md`** — copy-pasteable Docker Hub + GHCR publish runbooks.
- **`site/content/docs/docker.mdx`** + meta — docs page.

## Design decision: node-baked, not nub-provisioned

nub augments the user's installed Node — it is not a self-contained runtime like bun. So the images start from an official `node:*` base (Node present, cached, version-pinned at build time) and layer nub on top. A nub-only image relying on `nub node` to provision Node on first use would hit the network on first run, bloat the runtime layer, and diverge from every official Node-tool image. `npm install -g @nubjs/nub` (not the curl script) is the install path: npm is already present, its os/cpu filters resolve the correct per-platform binary including `-musl` on Alpine, and the package postinstall chmods the binary as the installer (root) — the install-as-root-then-drop-user container pattern. Default base is **node:26** (current line); 22.x is only the fast-tier floor, exposed by overriding `NODE_BASE`, never the default.

## Best-practice audit vs oven/bun + OCI (applied / left)

| Best practice | Applied? |
|---|---|
| Non-root user (`USER node`, uid 1000) | yes — base already ships it |
| PID-1 zombie reaping (`tini`) | yes — nub already forwards SIGINT/TERM/HUP to its Node child |
| Base digest-pin (slim + alpine) | yes |
| OCI `org.opencontainers.image.*` labels | yes (version/revision via build-args) |
| `--no-install-recommends` / cache cleanup | yes |
| Alpine native-dep libs (`libgcc` + `libstdc++`) | yes |
| Multi-arch amd64 + arm64 | yes (QEMU + buildx, push leg) |
| Provenance + SBOM attestation | yes (on push) |
| HEALTHCHECK | left — a CLI/runtime image has no long-lived service to probe |
| Distroless variant | left — nub needs Node present; distroless-node is niche, defer to demand |

## Verification (local, both variants green)

| Variant | Base | Image size | `npm i -g @nubjs/nub` | Smoke (non-root) |
|---|---|---|---|---|
| slim | node:26-slim (glibc) | ~367 MB | **yes** | pass |
| alpine | node:26-alpine (musl) | ~266 MB | **yes** | pass |

Each variant: `nub --version` = v0.1.14, a TypeScript file ran, `nub install` materialized a dep and it loaded via `require()` — all as the non-root `node` user. Alpine resolved `@nubjs/nub-linux-arm64-musl`. OCI labels populate. Signal handling verified end-to-end: `docker stop` → SIGTERM → tini → nub → Node child shut down cleanly (exit 0, not a 137 SIGKILL timeout). The headline docs example (`FROM nubjs/nub` + `COPY --chown=node:node . .` + `nub install`) was built downstream and confirmed to install + load a dependency as non-root.

**`npm install -g @nubjs/nub` works cleanly in a fresh node:26-slim AND node:26-alpine container — yes on both.**

Two adversarial L2 reviewer passes (fresh context, Opus) found no P0s and no leaked secrets; confirmed a PR can never push (every push step gated `if: github.event_name == 'release' && vars.DOCKER_IMAGE != ''`). Their P1s (SHA-pin convention, missing `packages: write`, the non-root docs example, bind-mount note) and P2s (cache-layer reorder, arm64 smoke-gap note) are all addressed.

## Maintainer-owned decisions (recommend-only — not decided here)

- **Registry:** Docker Hub `nubjs/nub` vs GHCR `ghcr.io/nubjs/nub` vs both. The workflow is parameterized via `vars.DOCKER_IMAGE` / `vars.DOCKER_REGISTRY`; the push leg no-ops until that var is set. `docker/PUBLISHING.md` has copy-paste steps for each.
- **Image name/namespace.** The docs + runbook show `nubjs/nub` as the proposed name — adjust on main if you pick another.
- **Variants + tag scheme at launch.** Proposed: slim = `latest`/`<ver>`/`slim`/`<ver>-slim`; alpine = `alpine`/`<ver>-alpine` (mirrors oven/bun).
- **Final ratification** of node-baked-vs-provisioned.

Do not merge or publish until those are signed off.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,0 +1,136 @@
+name: Docker image
+
+# Builds the official nub Docker images (Debian slim + Alpine), installing the
+# released @nubjs/nub npm package onto an official node:* base.
+#
+#   - On PR / push to main: build BOTH variants and run the in-image smoke test.
+#     Never pushes.
+#   - On a release tag (v*): build, smoke-test, then push to the configured
+#     registry.
+#
+# REGISTRY WIRING IS A MAINTAINER-OWNED TODO — see the `vars`/`secrets` below.
+# The registry choice (Docker Hub `nubjs/nub` vs GHCR `ghcr.io/nubjs/nub` vs both)
+# and the image name are deliberately NOT hardcoded. Set them at merge:
+#
+#   Repo variables (Settings → Secrets and variables → Actions → Variables):
+#     DOCKER_IMAGE   e.g. "nubjs/nub" (Docker Hub) or "ghcr.io/nubjs/nub" (GHCR)
+#     DOCKER_REGISTRY (optional) e.g. "ghcr.io" — omit for Docker Hub
+#   Secrets (only needed for the push leg):
+#     DOCKERHUB_TOKEN  (Docker Hub) — paired with the DOCKERHUB_USERNAME variable
+#     (GHCR uses the built-in GITHUB_TOKEN; no secret needed — see the login step.)
+#
+# Until DOCKER_IMAGE is set, the push leg is skipped (the `if` guards it), so this
+# workflow is safe to merge as build-and-smoke-only.
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - "docker/**"
+      - ".github/workflows/docker.yml"
+  pull_request:
+    branches: [main]
+    paths:
+      - "docker/**"
+      - ".github/workflows/docker.yml"
+  release:
+    types: [published]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build:
+    name: Build + smoke (${{ matrix.variant }})
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - variant: slim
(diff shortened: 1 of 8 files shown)


colinhacks:

PR #110, opened 2026-06-24
perf(cli): drop the dead config-tui (ratatui/crossterm) from the embedded engine

jdalton:
## What

Set `default-features = false` on nub-cli's `aube` path dependency, dropping aube's default feature set (`mimalloc`, `config-tui`). No `vendor/aube` changes — this is purely nub-cli's feature selection of its dependency.

Roadmap item **B2-tui**.

## Why these features are dead under nub

- **`config-tui`** enables `ratatui` + `crossterm`, which back only aube's interactive `aube config tui` subcommand. nub embeds aube as a library and runs its own CLI; aube's tool subcommands (including the config TUI) are never dispatched, and engine output is rebranded through `pm_engine::present`. The TUI code is unreachable.
- **`mimalloc`** (aube's feature) only swaps the global allocator inside aube's own `main.rs`, which is also dead under nub. nub-cli already sets `mimalloc::MiMalloc` as its global allocator directly in `crates/nub-cli/src/main.rs`, so the allocator is unchanged by dropping aube's feature.

## Measured impact (darwin-arm64, release profile: thin-LTO, codegen-units=1, strip)

- **Binary size: 43,173,392 → 42,870,304 bytes — 303,088 bytes (296 KiB) smaller** (same-tree before/after).
- `size -m`: `__text` section −233,764 bytes; `__TEXT` segment −294,912 bytes.
- **The ratatui/crossterm subtree is removed from the resolved graph** (`ratatui`, `ratatui-core`, `ratatui-widgets`, `ratatui-crossterm`, `crossterm`, `crossterm_winapi`, plus exclusive deps `palette`, `kasuari`, `time`, `derive_more`, `darling`, …) — a 368-line `Cargo.lock` reduction here.
- `cargo bloat` (baseline): ratatui_core 38.0 KiB, crossterm 26.5 KiB, ratatui_widgets 26.1 KiB, ratatui_crossterm 460 B of attributed `.text`; all absent after the change.

## Verification

- `cargo tree -p nub-cli -i ratatui` and `-i crossterm`: both **present before** (`ratatui/crossterm → aube → nub-cli`), **absent after**.
- `cargo build --release`: clean.
- `cargo test`: identical pass/fail set before and after the change (verified by running the suite in both an unchanged baseline tree and the changed tree — same result), so **no regressions** introduced.
- `cargo fmt --check`: clean.
- `cargo clippy --all-targets --all-features -- -D warnings`: clean.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -146,15 +146,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7f202df86484c868dbad7eaa557ef785d5c66295e41b460ef922eca0723b842c"
 
 [[package]]
-name = "approx"
-version = "0.5.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cab112f0a86d568ea0e627cc1d6be74a1e9cd55214684db5561995f6dad897c6"
-dependencies = [
- "num-traits",
-]
-
-[[package]]
 name = "arraydeque"
 version = "0.5.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -241,7 +232,6 @@ dependencies = [
  "clap_usage",
  "clx",
  "console",
- "crossterm",
  "demand",
  "diffy",
  "flate2",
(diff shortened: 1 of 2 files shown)


colinhacks:

PR #115, opened 2026-06-24
Write link: version for versionless workspace members in pnpm-lock.yaml

### LINKED ISSUE #113: `nub install` breaks `pnpm --frozen-lockfile` when a workspace package omits `version`
Hi! I've just started using nub and really like it so far. I ran into something around pnpm workspace handling, so I wanted to open an issue about it. Thanks for building such a great tool!

## What's wrong

When `nub install` writes `pnpm-lock.yaml` for a workspace where an internal package **omits the `version` field** in its `package.json`, nub serializes the `workspace:` dependency edge with the package's default version (`version: 0.0.0`) instead of pnpm's `version: link:<relative-path>` form.

Stock pnpm then rejects nub's own lockfile under a frozen install:

```
ERR_PNPM_LOCKFILE_MISSING_DEPENDENCY  Broken lockfile: no entry for 'foo@0.0.0' in pnpm-lock.yaml
```

This contradicts the documented guarantee at https://nubjs.com/docs/install/pnpm that "a lockfile Nub writes installs cleanly under pnpm" (including `--frozen-lockfile`), with `workspace:` listed as supported. A workspace package that declares an explicit `version` serializes correctly — only the **versionless** case is affected. Any CI on stock pnpm + frozen lockfile (Vercel, etc.) fails at install on every commit nub last wrote the lockfile.

## Reproduce

Minimal repro (clone, then `nub install`): https://github.com/morinokami/nub-pnpm-lockfile

```sh
git clone https://github.com/morinokami/nub-pnpm-lockfile && cd nub-pnpm-lockfile
nub install                      # writes pnpm-lock.yaml
grep -A2 'foo:' pnpm-lock.yaml   # edge shows `version: 0.0.0`  (expected: link:../../packages/foo)
pnpm install --frozen-lockfile   # fails: Broken lockfile: no entry for 'foo@0.0.0'
```

`packages/foo` has no `version` field; `apps/app` depends on it via `"foo": "workspace:*"`. Adding `"version": "0.0.0"` to `packages/foo/package.json` makes nub emit the correct `link:` form and frozen install passes.

## Environment

- nub `v0.1.14`
- pnpm `v10` / `v11` (used to validate the frozen install)
- `lockfileVersion: '9.0'`
- macOS (arm64), Node v24.15.0

---

Likely cause: in the vendored aube writer, `crates/aube-lockfile/src/pnpm/write.rs` only registers workspace members that have **both** `name` and `version`, while the resolver (`crates/aube/src/commands/install/workspace.rs`) defaults a missing version to `0.0.0`. A versionless member therefore misses the `name@version` lookup and falls through to the registry-version branch. Defaulting the missing version to `0.0.0` in the writer (mirroring the resolver) should fix it.

colinhacks:
## Problem

When `nub install` writes `pnpm-lock.yaml` for a pnpm workspace where an internal member's `package.json` omits the `version` field, nub serialized the consuming importer's `workspace:` edge as `version: 0.0.0` instead of pnpm's `version: link:<relative-path>` form. Stock pnpm then rejected nub's own lockfile under a frozen install:

```
ERR_PNPM_LOCKFILE_MISSING_DEPENDENCY  Broken lockfile: no entry for 'foo@0.0.0' in pnpm-lock.yaml
```

A workspace member with an explicit `version` serialized correctly — only the versionless case was affected. Reported with a minimal repro in #113.

## Root cause

The pnpm writer (`vendor/aube/crates/aube-lockfile/src/pnpm/write.rs`) recovers each workspace member's directory from a map keyed on the member manifest's `name@version`, used to render the importer edge as `link:<dir>`. That map only inserted members whose `package.json` declared **both** `name` and `version`. The resolver (`vendor/aube/crates/aube/src/commands/install/workspace.rs`) defaults a missing version to `0.0.0`, so a versionless member's edge keys as `name@0.0.0` — it missed the lookup and fell through to the bare-version branch.

## Fix

Default the writer's missing version to `0.0.0`, mirroring the resolver, so the `link:` target resolves. The change is a no-op for the versioned path (the `unwrap_or` only fires when `version` is absent), and the lookup is gated on the dep having no `packages:` entry, so it cannot collide with a real registry `name@0.0.0` package.

## Verification

- New regression test `versionless_workspace_member_writes_link_not_bare_version` — confirmed to fail on the pre-fix code and pass with the fix.
- Existing pnpm writer tests green (117/117), including the versioned-member round-trip.
- e2e against the reporter's repro: nub now writes `version: link:../../packages/foo`, and stock pnpm 10.15.1 `install --frozen-lockfile` passes.

Fixes #113

### DIFF
--- a/vendor/aube/crates/aube-lockfile/src/pnpm/tests.rs
+++ b/vendor/aube/crates/aube-lockfile/src/pnpm/tests.rs
@@ -2721,6 +2721,62 @@ fn workspace_deps_write_importer_relative_link_versions() {
     );
 }
 
+// A workspace member whose package.json omits `version` must still
+// serialize the consuming importer's edge as `link:<dir>`, not the bare
+// default version. The resolver defaults a missing version to "0.0.0"
+// (install::workspace), so the dep_path is `name@0.0.0`; the writer must
+// mirror that default when recovering member dirs or the lookup misses
+// and pnpm rejects the lockfile under `--frozen-lockfile`
+// (ERR_PNPM_LOCKFILE_MISSING_DEPENDENCY: no entry for `foo@0.0.0`).
+#[test]
+fn versionless_workspace_member_writes_link_not_bare_version() {
+    let dir = tempfile::tempdir().unwrap();
+    let lockfile_path = dir.path().join("pnpm-lock.yaml");
+    // `foo` has a name but no version — the regression case.
+    for (rel, body) in [
+        ("apps/app", r#"{"name": "app", "version": "0.0.0"}"#),
+        ("packages/foo", r#"{"name": "foo"}"#),
+    ] {
+        let pkg_dir = dir.path().join(rel);
+        std::fs::create_dir_all(&pkg_dir).unwrap();
+        std::fs::write(pkg_dir.join("package.json"), body).unwrap();
+    }
+
+    let mut importers = BTreeMap::new();
+    importers.insert(".".to_string(), vec![]);
+    importers.insert(
+        "apps/app".to_string(),
+        vec![DirectDep {
+            name: "foo".to_string(),
+            dep_path: "foo@0.0.0".to_string(),
+            dep_type: DepType::Production,
+            specifier: Some("workspace:*".to_string()),
+        }],
+    );
+    importers.insert("packages/foo".to_string(), vec![]);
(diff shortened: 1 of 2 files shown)


colinhacks:

PR #116, opened 2026-06-24
pm: fix packageManager URL-form crash and store-open without HOME

colinhacks:
Two pnpm-conformance fixes surfaced by the v10.15.1 conformance harness spike.

## `packageManager` URL form no longer crashes the PM shim

A project pinning a URL/git reference instead of a version — `packageManager: "pnpm@https://github.com/pnpm/pnpm"` — made nub's pnpm-identity shim feed the URL to the semver resolver and abort the run with `unexpected character 'h'`. Corepack and pnpm both treat a URL value as "no version pin" (corepack downloads the referenced artifact; pnpm does not version-switch). `parse_spec` now recognizes a URL/git reference in the version slot (`https://…`, `git+…`, `git@…`) and treats the field as name-only, falling through to the user's PM on PATH / a dynamic default — matching both tools. Normal exact pins, dist-tags, ranges, and the Corepack `+sha512.…` hash suffix are untouched.

The conformance test `install should not fail for packageManager field with url` now passes; its allowlist entry is removed.

## Store open no longer requires HOME when `storeDir` is configured

`Store::with_root` / `default_location` resolved the store's cache directory from `$XDG_CACHE_HOME`/`$HOME` via `dirs::cache_dir()` and hard-failed with `Error::NoHome` when neither is set — even when a concrete `storeDir` is configured. So `nub install` in an environment with HOME stripped from the env (a minimal container, or pnpm's own test harness, which keeps only `PATH`/`COLORTERM`/`APPDATA`) crashed with `failed to open store: HOME environment variable not set` after already scaffolding the manifest. The store open now resolves the cache dir through `resolved_cache_dir` and the store root through the configured `storeDir` (else the platform default, else a `$TMPDIR` fallback), via a new `Store::with_root_and_cache(root, cache_dir)` constructor.

This fixes two latent bugs in one. Besides removing the spurious `NoHome` abort, `resolved_cache_dir` honors a configured `cacheDir` (`.npmrc` / `AUBE_CACHE_DIR`), which `dirs::cache_dir()` ignored — so the store's packument/index cache now lands under a configured `cacheDir` instead of always under `~/.cache`. That aligns the store-open path with the `packument_cache_dir` helper, which already honored `cacheDir` (the two were split-brained). For a user with HOME set and **no** `cacheDir` configured, behavior is byte-identical.

## Scope

Both changes are isolated and default-preserving. The `vendor/aube` change is a justified latent-bug-fix: standalone aube's behavior is unchanged when HOME is set; only the previously-aborting no-HOME path now degrades gracefully, consistent with the rest of the engine. The sha256 fail-closed pin-hash path and the global-package layout are untouched.

Each fix carries a focused unit test. `cargo clippy --all-targets -- -D warnings` and `cargo fmt --check` are clean for the touched crates; the conformance run stays green (0 surprise, 0 stale-allow).

Refs the pnpm-conformance work.

### DIFF
--- a/crates/nub-core/src/pm/resolve.rs
+++ b/crates/nub-core/src/pm/resolve.rs
@@ -799,7 +799,28 @@ pub fn parse_spec(spec: &str) -> Result<PmPin> {
     let (name, version) = spec.split_once('@').with_context(|| {
         format!("packageManager \"{spec}\" must be in name@version form (e.g. pnpm@9.1.0)")
     })?;
+    // A URL/git reference in place of a version (`pnpm@https://…`, `pnpm@git+…`)
+    // is a Corepack-supported form meaning "fetch this artifact" — there is no
+    // semver to pin. nub can't provision an arbitrary tarball as a PM version,
+    // so it treats the field as NAME-only (no version pin) and falls through to
+    // the user's PM on PATH / a dynamic default — matching pnpm, which does not
+    // version-switch on a URL spec. Without this the URL reached the semver
+    // resolver and aborted the run (`unexpected character 'h'`).
+    if is_url_version_ref(version) {
+        return classify(name, None);
+    }
     classify(name, Some(version))
+}
+
+/// Whether a `packageManager` *version* slot holds a URL / git / file reference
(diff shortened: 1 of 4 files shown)


colinhacks:

PR #119, opened 2026-06-24
test(worker): WPT conformance harness + wpt-worker CI leg

colinhacks:
Production WPT conformance harness for nub's browser-shape `Worker` — Phase 2 of the worker-test-coverage effort (the v0.2 release gate). Builds on the Phase-1 research + prototype.

## What this adds

- **A production wpt-runner** (`tests/worker-wpt/harness/`) following Node's `test/common/wpt.js` model: a subprocess-per-file driver that loads the **real** `testharness.js` (not a shim) and runs each test **through nub's runtime**, so the polyfilled `Worker` / `MessageChannel` / `MessagePort` / `MessageEvent` / `structuredClone` are the globals under test. Worker-scope (`global=worker`) tests run **inside a real nub `Worker`**, so worker-scope conformance is genuinely exercised.
- **A pinned, vendored WPT slice** (`tests/worker-wpt/wpt/`, 24 `.any.js` files at WPT commit `80b7ba49`): the `webmessaging` battery, the structured-clone battery over a `MessageChannel`, and the worker-scope event tests, plus `testharness.js` and the includes. WPT's BSD license travels with it.
- **A per-file status file** (`status.json`): skip / expected-fail / pass, each documented.
- **A `wpt-worker` CI leg** (`.github/workflows/wpt-worker.yml`) — build nub once, run the matrix across the **compat tier** (Node 18.19, 20) and the **fast tier** (22.15, 24). Gated green.

## Measured conformance

**165 subtest pass, 21 expected-fail, 0 unexpected-fail, 3 files skipped** — identical on the floor (18.19), compat (20), and fast (26) tiers. The messaging and structured-clone core passes in full. The runner is hardened against false greens (a non-OK harness status, a watchdog non-completion, and a stale expected-fail name each turn the run red — verified with failure injection), across two rounds of adversarial self-review.

## Blob `instanceof` + Response fix (maintainer-directed)

The runner surfaced that a cloned `Blob` failed `instanceof Blob`, and that `new Response(clonedBlob).arrayBuffer()` returned the stringified `"[object Blob]"` (13 bytes) instead of the bytes — both diverging from native Node. Root cause: nub's blob: worker support replaced `globalThis.Blob` with a subclass, so `globalThis.Blob !== node:buffer.Blob`; the structured-clone deserializer returns a native Blob that failed both `instanceof` and undici's webidl brand check (which uses the ordinary `Symbol.hasInstance`, so a custom one can't fix it). Fixed by making `globalThis.Blob` a `Proxy` over the native Blob — identity (`.prototype`) is preserved, so every native Blob passes `instanceof` and the undici check exactly as on vanilla Node, while the construct trap still records parts for sync blob: worker assembly; `newTarget` is forwarded so user `class X extends Blob {}` keeps its prototype. This took WPT from 145 → **165** subtest pass. Existing worker integration tests (15) and the blob: worker fixtures pass.

## Maintainer-owned — held for sign-off

**Port auto-start** stays `expected-fail` (the `no-start` / `MessagePort_initial_disabled` divergence): adding a `message` listener without `start()` delivers — inherited from `node:worker_threads` (and shared by `workerd`). Document-vs-fix is the maintainer's call.

**FAQ + docs wording (public copy):** `site/content/docs/runtime/workers.mdx` is rewritten against the actual polyfill surface (classic mode, inline `data:`/`blob:` workers, `name`, populated `ErrorEvent` location) with the rough-edges FAQ folded in. Two FAQ claims were re-confirmed empirically in a real worker — `ErrorEvent` `filename`/`lineno`/`colno` **are** populated; `event.target` for a non-message event **is** a private `EventTarget`. The wording needs a maintainer pass before the v0.2 announce.

Refs the worker-test-coverage effort (v0.2 release gate). No GitHub issue — internal pre-release work.


colinhacks:

PR #121, opened 2026-06-24
ci: bump Homebrew tap formula on release

colinhacks:
Adds Homebrew distribution for nub via a custom tap (\`nubjs/homebrew-tap\`, already created + populated) and automates per-release formula bumps from the release workflow.

## What this PR does

- **New CI job \`bump-homebrew-tap\`** in \`.github/workflows/release.yml\`. After \`github-release\` completes on a real tag-push release, it checks out \`nubjs/homebrew-tap\`, regenerates \`Formula/nub.rb\` for the new version, commits, and pushes.
- **New script \`.github/scripts/gen-homebrew-formula.sh\`** — templated formula generator. It reads the four GNU/macOS \`.sha256\` sidecar assets from the release (\`darwin-arm64\`, \`darwin-x64\`, \`linux-arm64\`, \`linux-x64\`) rather than recomputing them, so the formula can never disagree with the shipped tarballs. musl tarballs and the win32 \`.zip\` are omitted (not Homebrew-installable).
- **Docs**: Homebrew install note (\`brew install nubjs/tap/nub\`) added to the README install block, the docs landing page, and the FAQ curl-install answer.

## Trigger gating

The job is gated \`if: github.event_name == 'push'\` so it runs ONLY on real \`v*\` tag releases, never the debug \`workflow_dispatch\` build-only path. It also \`needs: github-release\` so the release + its sidecar assets exist before the formula is generated.

## Manual step required (maintainer)

The default \`GITHUB_TOKEN\` is scoped to \`nubjs/nub\` and cannot push cross-repo to \`nubjs/homebrew-tap\`. **Add a repo secret \`HOMEBREW_TAP_TOKEN\`** — a PAT (or fine-grained token) with \`contents:write\` on \`nubjs/homebrew-tap\`. Until it is added, the job skips cleanly (logs a warning, does not fail the release); the tap can be bumped manually by running the generator.

## Verification

The formula was installed, tested, and audited locally on macOS arm64 against v0.1.14:
- \`brew install nubjs/tap/nub\` → \`brew test nub\` (exit 0) → \`brew audit --strict nub\` (exit 0).
- \`nub --version\` reports \`v0.1.14\`; \`nub hello.ts\` transpiles + runs (proves \`runtime/\` is installed via the libexec tree, not a bare binary).
- The generator script reproduces the hand-verified formula byte-for-byte.

The formula uses plain symlinks (\`bin.install_symlink\`), no wrapper script, so \`nub\`/\`nubx\` exec the native binary with zero per-call overhead; nub canonicalizes \`current_exe()\` to find \`runtime/\` beside the real binary.

Infra change, no associated issue.

### DIFF
--- a/.github/scripts/gen-homebrew-formula.sh
+++ b/.github/scripts/gen-homebrew-formula.sh
@@ -12,7 +12,7 @@ set -euo pipefail
 # Usage: gen-homebrew-formula.sh <version> <output-path>
 #   <version>     release version without the leading v (e.g. 0.1.14)
 #   <output-path> where to write nub.rb
-# Requires: gh (authed to read nubjs/nub releases), curl.
+# Requires: gh (authed to read nubjs/nub releases).
 
 VERSION="${1:?usage: gen-homebrew-formula.sh <version> <output-path>}"
 OUT="${2:?usage: gen-homebrew-formula.sh <version> <output-path>}"
--- a/.github/scripts/gen-homebrew-formula.sh
+++ b/.github/scripts/gen-homebrew-formula.sh
@@ -1,0 +1,103 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Regenerate the Homebrew formula for nubjs/homebrew-tap from a published release.
+#
+# Fills VERSION + the four GNU/macOS sha256s into Formula/nub.rb. The sha256s are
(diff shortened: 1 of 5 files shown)


colinhacks:

PR #122, opened 2026-06-24
githooks: add pre-commit guard blocking direct commits to the shared main tree

colinhacks:
Adds a `pre-commit` hook that blocks a commit made in the SHARED (non-worktree) checkout while on `main`, forcing tracked-file commits through worktrees.

## Why

The shared `main` working tree must stay always fast-forwardable to origin/main. A commit made directly in the shared tree on a stale base creates a local commit not on origin — so `git pull --ff-only` then fails (divergence, not just behind) and the tree silently rots. AGENTS.md already forbids direct shared-tree commits, but nothing enforced it structurally. This makes the rule structural.

## Mechanism

Detection: in the shared clone `git-dir` == `git-common-dir`; in a linked worktree they differ. The hook blocks only the shared-clone + on-`main` combination — a worktree, a deliberate non-main branch, and detached HEAD (rebases) all pass through. Escape hatch `NUB_ALLOW_SHARED_COMMIT=1` for a sanctioned emergency.

## Verification

- Shared tree on `main`: real `git commit` BLOCKED (exit 1), HEAD unchanged.
- `NUB_ALLOW_SHARED_COMMIT=1 git commit`: ALLOWED.
- Worktree commit: ALLOWED (proven by this PR's own commit, made in a worktree).
- Detached HEAD / non-main branch: ALLOWED (rebase-safe).

`core.hooksPath` is `.githooks` (shared across worktrees), so the hook is active everywhere with no per-clone setup.

Refs: AGENTS.md "Default to a PR flow (from a git WORKTREE)" and its no-direct-shared-commit corollary.

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -1,0 +1,51 @@
+#!/bin/sh
+# pre-commit hook: block direct commits to the SHARED main checkout.
+#
+# WHY: the shared `main` working tree must stay ALWAYS fast-forwardable to
+# origin/main. A commit made directly in the shared tree creates a local commit
+# that is NOT on origin; on a stale base this DIVERGES the shared tree from
+# origin/main, so `git pull --ff-only` then fails and the tree silently rots
+# (AGENTS.md "Default to a PR flow" + its no-direct-shared-commit corollary).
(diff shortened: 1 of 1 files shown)


colinhacks:

PR #124, opened 2026-06-24
githooks: only block AGENT commits to shared main (not the maintainer's terminal)

colinhacks:
PR #122's pre-commit guard blocked **every** direct commit to the shared main checkout — including the maintainer's legitimate one-off terminal commits.

This narrows the block to automation only. A Claude Code / dispatched-agent context has `CLAUDECODE` / `AI_AGENT` set in the environment; a bare human terminal has neither. The hook now blocks only when running under automation on the shared main tree:

- Agent commit to shared `main` → blocked; land it via a worktree.
- Maintainer's own terminal (no markers) → never blocked.
- Worktree, non-main branch, detached HEAD → always allowed.
- `NUB_ALLOW_SHARED_COMMIT=1` → sanctioned bypass with a stderr note.

Pure-env check, no fetch, works offline. The earlier stale-base approach on this branch was dropped in favor of the simpler, intent-matching automation marker.

Refs #122

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -1,20 +1,22 @@
 #!/bin/sh
-# pre-commit hook: block direct commits to the SHARED main checkout.
+# pre-commit hook: block an AGENT commit to the SHARED main checkout.
 #
 # WHY: the shared `main` working tree must stay ALWAYS fast-forwardable to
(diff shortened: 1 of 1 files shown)


colinhacks:

PR #128, opened 2026-06-24
hooks: guard gh-comment verbosity + sharpen issue-comment tone rule

colinhacks:
Deterministic floor against verbose GitHub comments posted by dispatched sub-agents (which don't inherit AGENTS.md/PROSE.md, so prompt-level rules fail).

- `.claude/hooks/gh-comment-guard.mjs` — PreToolUse hook on Bash. Blocks a `gh issue/pr comment` / `gh pr create` body over 700 chars with a reminder of the PROSE.md tone rule (initial ack = exactly "Investigating."; substantive comments terse). `NUB_ALLOW_LONG_COMMENT=1` overrides. Fail-open; scoped to the comment/create verbs — unrelated `gh` and non-`gh` calls pass untouched.
- Wired into `.claude/settings.json` as a PreToolUse hook.
- Sharpened the codified rule in AGENTS.md (maintainer-hygiene) and the `address-issue` skill.

Verified: long body blocked (`--body`/`-b`/`--body=`/heredoc), "Investigating." passes, override bypasses, `--body-file`/missing-body pass with reminder, unrelated `gh` + non-`gh` untouched, unparseable stdin fails open.

Refs: subagent-context-floor thread.

PR #129, opened 2026-06-24
hooks: gh-comment guard measures the longest of multiple --body flags

colinhacks:
Follow-up to #128. Adversarial review found a minor bypass: `extractBody` returned only the first `--body` occurrence, but `gh` uses the LAST one — so `--body hi --body <essay>` slipped a long body past the length check. Now collects every `--body`/`-b`/heredoc candidate and measures the longest.

Verified: multi-body short-then-long now blocks; single long still blocks; "Investigating." and `gh pr view` still pass.

Refs #128.


colinhacks:

PR #130, opened 2026-06-24
fix(pm): resolve yarn-berry catalogs instead of erroring

### LINKED ISSUE #118: Yarn Compatibility Issue
I have a Yarn monorepo with cataloged dependencies. These are recently supported by .yarnrc, but nub complains:

```
➜  repo git:(main) nub install
Error: nub: `catalog:` specifier (@roarr/cli: catalog:) is not supported — this project uses yarn, which doesn't implement catalogs (pnpm@9+ and bun@1.2+ do). Inline the version, or switch the project to a PM that supports catalogs (`nub pm use pnpm`).
```

A little unclear the model that connects nub to the underlying package manager, but this seems like it may be a versioning issue.

Really excited to try this out!

colinhacks:
yarn supports catalogs since 4.10.0, declared in `.yarnrc.yml` with the same `catalog:`/`catalog:<name>` protocol as pnpm — only the file differs. nub previously errored on a `catalog:` specifier under a yarn identity; it now discovers `.yarnrc.yml` catalogs and resolves the specifiers itself.

- Gated on the incumbent being yarn-berry (v2+), not a parsed `>=4.10` version: nub's declarative yarn-version detection can't reliably split 4.10 from earlier berry, and a `catalog:` specifier only appears when the project actually uses the feature. Classic yarn v1 still refuses (no catalogs there).
- No PM binary is invoked for version detection — yarn-berry is read purely from declarative on-disk signals (`packageManager`, `devEngines`, lockfile, `.yarnrc.yml`).
- yarn stays read-only: catalog resolution reads `.yarnrc.yml` and resolves for the install graph; it never writes `yarn.lock`.

Tests: `.yarnrc.yml` default + named catalog discovery (aube); the per-role gate (yarn-berry honors, classic v1 refuses). e2e: `nub install` on a yarn-berry catalog fixture resolves `catalog:` (the old unsupported-catalog error is gone); classic v1 still errors.

Closes #118

### DIFF
--- a/crates/nub-cli/src/pm_engine/mod.rs
+++ b/crates/nub-cli/src/pm_engine/mod.rs
@@ -804,15 +804,17 @@ fn apply_config_scope(
     Ok(())
 }
 
-/// Does the active PM honor `catalog:` specifiers? pnpm@9+ and bun@1.2+
-/// implement catalogs; npm and yarn do not. nub identity honors catalogs
-/// (an un-branded cross-tool field, like `workspaces`). aube resolves both
-/// dialects: pnpm's `pnpm.catalog(s)` / `pnpm-workspace.yaml` AND bun's
-/// `workspaces.catalog(s)` in `package.json` (see aube's `discover_catalogs`),
-/// so honoring bun here resolves the real catalog rather than mis-failing a
-/// project that works under bun.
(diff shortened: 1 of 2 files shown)


colinhacks:

PR #131, opened 2026-06-24
fix(pm): honor yarn npmMinimalAgeGate + drop fabricated bun trustedDependencies gate

colinhacks:
Two compat-correctness defects the PM feature-matrix re-audit confirmed.

## FIX 1 — yarn `npmMinimalAgeGate` silently dropped (security-honoring)

Yarn Berry has a supply-chain min-release-age gate `npmMinimalAgeGate` plus an allowlist `npmPreapprovedPackages` (berry `packages/plugin-npm/sources/index.ts`, `scopablePackageGateSettings`). nub already has the engine gate `minimumReleaseAge` and reads it for npm, pnpm, and bun — but the yarnrc reader (`vendor/aube/crates/aube-registry/src/config/yarnrc.rs`) never mapped yarn's spelling, so a yarn project's configured gate was ignored.

Map `npmMinimalAgeGate` → `minimumReleaseAge` and `npmPreapprovedPackages` → `minimumReleaseAgeExclude`, **only when the user set them** (yarn's `1d` default is yarn's, not nub's — an unset gate emits no entry).

Unit conversion: `npmMinimalAgeGate` is a `DURATION` declared in `MINUTES` (berry `Configuration.ts` + `miscUtils.parseDuration`). The engine's `minimumReleaseAge` is also minutes, so:
- a **bare number** (`1440`) is already minutes — pass through;
- a **unit string** (`7d`) is converted via berry's ms table then ÷ minutes (`7d` = 10080 min);
- **sub-minute** values (`30s`) round **up** to 1, so a non-zero gate never collapses to 0 and silently disables the control (same rationale as the bunfig reader).

Scope-level `npmScopes.<s>.npmMinimalAgeGate` is **not** mapped — the engine has no per-scope min-age setting; top-level only.

This honors a **user-configured supply-chain security control**, consistent with how nub already treats npm/pnpm/bun. No posture change — it only stops dropping config the user wrote. Hold on review if you want it gated differently.

## FIX 2 — fabricated bun `trustedDependencies` drop (latent)

`crates/nub-cli/src/pm_engine/config_scope.rs` claimed bun@10 dropped `trustedDependencies` via a pnpm-style approve-builds migration. Fabricated: there is no bun major 10; bun 1.3.x still uses `trustedDependencies` (bun.com/docs/install/lifecycle; bun source `install/lifecycle.zig` + `lockfile.zig`), and `onlyBuiltDependencies`/`approve-builds` have zero hits in bun source. Dropped the `<10` major gate so bun honors `trustedDependencies` unconditionally; rewrote the comment factually; replaced the false `bun_10_dropped_trusted_dependencies` test.

## Verified
- `cargo build -p nub-cli -p aube-registry` — clean
- `cargo clippy --all-targets --all-features -- -D warnings` — clean
- `cargo fmt --check` — clean
- new test `yarnrc_translates_min_age_gate_to_minimum_release_age` (7d→10080, bare-number→minutes, 30s→1 round-up, unset→no entry) + all 133 aube-registry config tests pass
- new test `only_bun_honors_trusted_dependencies` + `pm_two_mode` suite pass

Refs: pm-feature-matrix-reaudit audit context.

### DIFF
--- a/crates/nub-cli/src/pm_engine/config_scope.rs
+++ b/crates/nub-cli/src/pm_engine/config_scope.rs
@@ -238,12 +238,13 @@ fn honors_namespaced_overrides(role: Role) -> bool {
 }
 
 /// Does the active PM honor Bun's top-level `trustedDependencies` (the
-/// build-script allowlist)? Only bun — and bun@10 DROPPED it (pnpm@10's
-/// approve-builds migration also dropped `trustedDependencies` +
-/// `onlyBuiltDependencies`, and bun followed). `major` is bun's declared
-/// major; undeclared bun is assumed current (honors).
-pub(crate) fn honors_trusted_dependencies(role: Role, major: Option<u64>) -> bool {
-    role == Role::Bun && major.is_none_or(|m| m < 10)
+/// build-script allowlist)? Only bun — at every major. Bun's current line
+/// (1.3.x) still uses `trustedDependencies` for lifecycle-script approval
+/// (bun.com/docs/install/lifecycle; bun source `install/lifecycle.zig` +
+/// `lockfile.zig`), and ships no `onlyBuiltDependencies` / `approve-builds`
+/// migration (zero hits in bun source). No version gate applies.
+pub(crate) fn honors_trusted_dependencies(role: Role) -> bool {
+    role == Role::Bun
 }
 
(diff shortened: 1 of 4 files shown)


colinhacks:

PR #136, opened 2026-06-25
fix(ci): guard thread::spawn against EAGAIN + cgroup-aware concurrency cap so install degrades instead of aborting (panic=abort)

colinhacks:
## The bug

`nub ci` intermittently exited **101** on resource-constrained CI (observed on Vercel) at the install tail. 101 is the Rust panic-runtime exit code — it pins the cause to a panic inside nub's own process, not a mirrored child (a failing child mirrors as the child's code, e.g. 1; a SIGABRT is 134).

Root cause: a thread/process spawn fails with **EAGAIN** under peak PID/thread pressure. Two flavors:

1. An unguarded `std::thread::spawn` whose OS thread-create fails — `thread::spawn` panics on failure.
2. tokio's **internal** blocking-pool thread growth (`spawn_blocking` for CAS save/restore during the build-script phase, concurrent with the parallel native postinstalls) panicking when `clone(2)` returns EAGAIN.

v0.2 ships `panic = "abort"` for the CLI (#112), so either panic now **aborts the whole install** rather than unwinding one thread — a latent whole-process crash in the shipped release.

## The fix (two parts)

**1. Guard unguarded `thread::spawn` against EAGAIN.** Each switches to `thread::Builder::new().spawn(...)` (returns `io::Result`) and degrades gracefully:

| Site | Class | Degraded behavior |
|---|---|---|
| `vendor/aube/.../progress/ci.rs` heartbeat | cosmetic | skip the ticker (`stop()` already tolerates a `None` handle) |
| `nub-core/src/node/spawn.rs` signal-forward | best-effort | don't install the forwarder (same as the pre-existing `Signals::new`-fails path) |
| `cli.rs` script stdout/stderr drains | load-bearing | drain inline (a missing reader would deadlock the child on a full pipe) |
| `cli.rs` workspace `-r` worker pool | load-bearing | survivors drain the shared queue; if **all** workers fail, drain inline — work is never lost, only parallelism |
| `pm_engine/mod.rs` fd-capture drain | load-bearing | run `f` then drain inline (the captured output is a few bytes) |

**2. Prevent the exhaustion the panic rides on.** We cannot guard inside tokio or rayon, and `catch_unwind` cannot save a `panic = "abort"` process — so the only in-process fix is to keep the peak thread+process count under the box's ceiling. New `pm_engine/resource_limits.rs` detects the effective ceiling (cgroup **v1 + v2** `pids.max`, `RLIMIT_NPROC`) and, **only under a detected constraint**, caps — under ONE shared headroom budget (`split_budget`, so the shares SUM within the ceiling) — the four concurrent consumers:

- tokio worker pool + tokio blocking pool (`build_runtime`),
- the **rayon global pool** (`cap_rayon_global_pool`) — the engine fans CAS writes / delta / fetch over rayon's implicit global pool, whose lazy init otherwise spins up `available_parallelism()` threads (CPU-quota-bound, **not** PID-bound) and panics on thread-create EAGAIN: the same exit-101 abort, relocated from tokio to rayon. This closes the second abort vector,
- aube's parallel build-script concurrency (`apply_constrained_child_concurrency`, via the neutral `npm_config_child_concurrency`).

On an unconstrained box every detector returns `None` and the full-speed defaults stand — **no normal-box perf regression**.

Also adds `spawn_with_eagain_retry`: a bounded (5×, ~155ms) EAGAIN/ENOMEM retry-with-backoff on nub's load-bearing child spawns, so a transient `fork` refusal retries instead of surfacing a spurious failure. Non-transient errors propagate immediately.

## Why panic=abort makes this load-bearing

Under `panic = "unwind"` (0.1.x) a panic in the cosmetic heartbeat thread killed only that thread. Under `panic = "abort"` (v0.2) the same EAGAIN aborts the whole install. Every guarded site is now strictly safer; the concurrency cap removes the tokio- AND rayon-internal panics that no guard can catch.

## Known limitations

- The script-pipe inline-drain multiplex uses `poll(2)` and so is Unix-only; on Windows the fallback drains sequentially, with a vanishing large-output deadlock window. Windows never exhibited the thread-exhaustion abort this guards, so this is acceptable.
- The constraint detector is Linux-only (cgroup + `RLIMIT_NPROC`); macOS/Windows are treated as unconstrained (the abort was Linux-CI-only).
- The poll-based inline drain decodes UTF-8 per chunk (`from_utf8_lossy`), so a multi-byte char split across a read boundary on the degraded path can render replacement chars; the normal threaded path (`BufReader::lines`) is unaffected. Degraded-path-only, cosmetic.

## Repro

Linux container (`node:22`, arm64) with a tightened cgroup `pids.max`, running the fixed binary:

`--pids-limit 200` (constraint detected) — the cap engages and the blocking pool is cut from 128 to fit the ceiling:

```
DEBUG constrained box detected: capping install runtime pools + build-script concurrency under the PID/thread ceiling headroom=135 workers=8 blocking=67 child_concurrency=5
```

No `--pids-limit` (unconstrained) — the detector returns `None`, no cap is applied, defaults stand (no log line, no concurrency change). This confirms the no-regression property: the cap engages only under a detected constraint.

The blocking-pool cut (128 → 67 under a 135-PID headroom) is the exact lever that keeps the tail thread+process peak under the ceiling, removing the tokio internal-spawn panic. `split_budget` keeps the shares summing within budget (workers 8 + blocking 67 = 75 ≤ 135).

The spawn-guard layer (`Builder::spawn` + graceful degrade) and the EAGAIN retry are validated by the unit tests + clippy and by the fixed binary completing workspace `-r` runs under pressure. A fully-deterministic "old exits 101 / new exits 0" side-by-side is hard to pin in a container because Node's own thread needs choke before nub's spawn sites under a tight `pids.max`; the panic mechanism itself is established by the diagnosis (101 = panic-runtime exit, the unguarded `thread::spawn` + tokio internal growth) and the guards are proven by construction + the cap repro above.

## Fork-discipline (vendor/aube hunk)

The only `vendor/aube` change is the heartbeat guard. It is **default-preserving**: standalone aube's happy path is byte-identical (same thread, same work); only the EAGAIN path changes from panic to "no ticker." A justified latent-bug fix, upstreamable to `jdx/aube`.

## What is NOT changed

No change to concurrency on unconstrained boxes. The cap engages only when a tight cgroup `pids.max` / low `RLIMIT_NPROC` is detected. **This is a default-behavior change gated on detected constraint — flagging it for explicit maintainer review.**

## Review

Four fresh-context adversarial reviewers covered: (i) concurrency-correctness + cgroup/`RLIMIT_NPROC` detection, (ii) the `panic = "abort"` interaction + remaining panic sites, (iii) the default-change blast radius, (iv) the EAGAIN-retry + inline-drain correctness. Findings incorporated in the second commit:

- **cgroup v1** is now detected alongside v2 (the constrained CI hosts are often v1/hybrid; the v2-only probe was blind there).
- **Budget over-subscription** fixed: one headroom budget is split across the three concurrent consumers via `split_budget`, so their sum stays under the ceiling instead of each independently clamping to `min(headroom, …)`.
- **Brand**: the constrained build-script count is set via the neutral `npm_config_child_concurrency` (which aube honors), not `AUBE_CHILD_CONCURRENCY` — no engine-brand leak even in internal plumbing.
- **Inline-drain deadlock** removed: the script pipe drain now uses a `poll(2)`-based concurrent inline fallback (`drain_both_inline`) instead of a probe that could leave a both-inline sequential drain able to deadlock a chatty child; stdout is `dup`ed so a failed thread-spawn leaves it recoverable.
- **EAGAIN-retry** confirmed correct (bounded 5×/~155ms, only `WouldBlock`/`EAGAIN`/`ENOMEM`, sync call sites only, fails fast on persistent/real errors).
- Pure headroom math extracted into unit-tested helpers.

Reviewers confirmed: no abort-capable `thread::spawn` remains on the hot path (the one untouched site is test-only); the `set_var` before the engine's env-snapshot is correctly ordered and single-threaded; tokio's internal blocking-pool growth can only be addressed by capping (no panic hook exists), which this does.

Refs #112

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2618,6 +2618,7 @@ dependencies = [
  "miette",
  "mimalloc",
  "nub-core",
+ "rayon",
  "rustc-hash",
  "semver",
  "serde_json",
--- a/crates/nub-cli/Cargo.toml
+++ b/crates/nub-cli/Cargo.toml
@@ -84,6 +84,13 @@ toml.workspace = true
 tokio = { version = "1", features = ["rt-multi-thread"] }
 tracing-subscriber.workspace = true
 url = "2.5"
+# Only to SIZE the rayon GLOBAL pool under a detected PID/thread constraint
+# (pm_engine::build_runtime): the embedded engine fans CAS writes / delta / fetch
+# out over rayon's implicit global pool, whose lazy init spins up
+# available_parallelism() threads (CPU-quota-bound, NOT PID-bound) and PANICS on
+# thread-create EAGAIN — the same exit-101 abort the tokio cap closes. We cap it
+# under the same headroom budget; unconstrained boxes keep rayon's default.
+rayon = "1"
 
 [target.'cfg(unix)'.dependencies]
 # fd-level capture of the engine's raw println/eprintln hint lines so the
--- a/crates/nub-cli/src/cli.rs
+++ b/crates/nub-cli/src/cli.rs
@@ -2693,75 +2693,82 @@ fn run_workspace_target(
             let (tx, rx) = mpsc::channel::<usize>();
             let rx = Arc::new(std::sync::Mutex::new(rx));
 
-            let num_workers = concurrency.min(chunk.len());
-            let workers: Vec<_> = (0..num_workers)
-                .map(|_| {
-                    let rx = Arc::clone(&rx);
-                    let failed = Arc::clone(&failed);
-                    let ran = Arc::clone(&ran);
-                    // Clone the selected members so they cross the thread boundary
-                    // (the borrowed `&members` slice can't be `move`d); paired with
-                    // their original index for the prefix color. `WorkspacePackage`
-                    // derives Clone, so this replaces the prior 4-field tuple
-                    // snapshot with one structured clone.
-                    let members_snapshot: Vec<(
-                        usize,
-                        nub_core::workspace::filter::WorkspacePackage,
-                    )> = chunk
+            // The per-worker loop body, factored out so it backs BOTH a spawned
+            // worker thread and the inline fallback (when thread creation fails
+            // under resource pressure). Each invocation pulls indices from the
+            // shared channel until it drains, so any number of live workers — even
+            // one, even the calling thread — completes ALL the work; thread-create
+            // EAGAIN only costs parallelism, never correctness.
+            let run_worker = |rx: Arc<std::sync::Mutex<mpsc::Receiver<usize>>>,
+                              failed: Arc<AtomicUsize>,
+                              ran: Arc<AtomicUsize>| {
+                let members_snapshot: Vec<(usize, nub_core::workspace::filter::WorkspacePackage)> =
+                    chunk
                         .iter()
                         .map(|&idx| (idx, members[idx].clone()))
                         .collect();
-                    let ws_root_buf = ws_root.to_path_buf();
-                    // Owned target + per-script knobs so they cross the thread
-                    // boundary; reconstituted into the borrowed forms inside the
-                    // worker (the borrowed forms can't be `move`d).
-                    let target = OwnedTarget::from(target);
-                    let ignore_scripts = exec.ignore_scripts;
-                    let script_shell = ws.script_shell.clone();
+                let ws_root_buf = ws_root.to_path_buf();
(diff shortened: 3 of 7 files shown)


colinhacks:

PR #137, opened 2026-06-25
perf(pm): read each config file once per command via an mtime-validated cache

jdalton:
The `pm_engine` re-read and re-parsed the same config files several times per command — root `package.json` (3x via aube's parser, 3x via serde), `.yarnrc.yml` (once per key), and the `.npmrc` ancestor chain (re-walked per key). This caches each per-invocation so every file is read once.

A per-process cache in `nub-core` (`config_cache::MtimeCache`), path-keyed and validated on a `(mtime, size)` stamp. The serde `package.json`, the aube-parsed manifest, and the yarnrc/npmrc/bunfig readers route through it.

Output-preserving with no stale read: every lookup re-stats the file and serves the cached value only when the stamp is unchanged, so a mid-command rewrite re-reads; a missing/unparseable file is never cached, matching the prior `.ok()` paths exactly.

Verified: reads of each existing config file drop 3x to 1x on an install fixture; cached vs uncached `install`/`list`/`why`/`outdated` give byte-identical stdout, stderr, and exit codes. New `config_cache` unit tests cover hit, mtime invalidation, size invalidation, and not-cached-on-missing. `clippy --all-targets --all-features -D warnings`, `fmt --check`, and `test -p nub-cli -p nub-core` pass.

### DIFF
--- a/crates/nub-cli/src/pm_engine/mod.rs
+++ b/crates/nub-cli/src/pm_engine/mod.rs
@@ -730,6 +730,29 @@ pub(crate) fn session_role_root(
     Some((role, detected.dir.clone()))
 }
 
+/// Per-process, mtime-validated cache of parsed `aube_manifest::PackageJson`
+/// keyed by file path. The PM-engine config phase parses the root manifest
+/// through aube's parser several times per command — `apply_config_scope`, the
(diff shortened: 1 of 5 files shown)


colinhacks:

PR #139, opened 2026-06-25
perf(cache): collapse warm-hit body copies from 3 to 1

jdalton:
Every warm transpile-cache hit copied the cached output three times: `read_to_string`, then `body.to_string()`, then the caller's `body[1..].to_string()`. `cache_get` now returns `(format_byte, code)` — it verifies integrity over the body slice, then drains the integrity prefix + format byte off the read buffer in place, reusing that one allocation as the final `code`. Three body copies become one (the file read).

Byte-identical: old `code = raw[16..][1..] = raw[17..]`; new `drain(..17)` leaves the same `raw[17..]`. The on-disk format and the integrity self-heal check are untouched. The length guard tightens `< 16` to `< 17`, rejecting only a never-written empty-body entry the old path would have panicked on.

Bench (`crates/nub-core/benches/cache_body_copy.rs`, mirroring `cache_hash.rs` since nub-native can't link a bench): isolated extraction work drops ~35 µs to ~11 µs on a 100 KB entry — the two eliminated body-sized allocations.

Gates green: clippy + fmt on both workspaces, `cargo test -p nub-cache-key -p nub-core -p nub-cli` (incl. `corrupt_cache_entry_self_heals`, `transpile_cache_writes_atomically`), nub-native release + fast builds. Cold-then-warm `.ts` run gives identical output.

Pushed from the `jdalton/nub` fork (no direct push access to nubjs/nub).

### DIFF
--- a/crates/nub-core/Cargo.toml
+++ b/crates/nub-core/Cargo.toml
@@ -58,5 +58,9 @@ harness = false
 name = "cache_hash"
 harness = false
 
+[[bench]]
+name = "cache_body_copy"
+harness = false
+
 [lints]
 workspace = true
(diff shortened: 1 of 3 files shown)


colinhacks:

PR #140, opened 2026-06-25
feat(install): size CPU-bound pools to the cgroup CPU quota, not host cores

colinhacks:
Generalizes the PID-only `resource_limits.rs` from #136 with a CPU-budget axis: the install's CPU-bound pools (tokio workers, the rayon global pool, aube's linker pool) now size to the box's cgroup CFS CPU quota instead of the over-reported host core count. On a quota-constrained container the pools no longer over-subscribe the quota — preventing the CFS-throttle latency cliff and the thread/PID growth that #136 caps reactively.

## Crate vs. port

Uses `num_cpus = "1"` (`num_cpus::get()`), not a hand-rolled cgroup parser. It reads the CFS quota for both cgroup v1 (`cpu.cfs_quota_us`/`cfs_period_us`) and v2 (`cpu.max`) via `/proc/self/mountinfo` and returns `min(ceil(quota/period), logical_cores)` — the automaxprocs algorithm with the chosen `max(1, ceil)` rounding. Maintained (v1.17.0), 2 deps. Replaced a first-draft ~80-line port.

## Finding

On a cgroup-v2 host, Rust std's `available_parallelism()` already reads `cpu.max` (since 1.61), so under `docker --cpus=0.5` it returns 1 on its own. `cpu_budget()` returns `None` there — its net value is the two gaps std misses: cgroup-v1 quota, and the affinity-blocked sandbox path (std drops the v2 quota when `sched_getaffinity` fails). `num_cpus` reads the files directly, covering both.

## Composition + override

- The rayon global pool is sized to `min(cpu-budget-derived, pid-headroom-derived)` and capped exactly once (a double `build_global` no-ops the second cap).
- A dedicated `NUB_CPU_BUDGET` env var overrides auto-detection (distinct from the fetch-concurrency `NUB_CONCURRENCY`). Explicit-wins, clamped `[1, host_logical_cpus]` (quota-ignoring) so it can raise above a detected quota, not only lower.

## Fork-discipline (vendor/aube)

aube's CPU-bound pools route through a new `Embedder.cpu_budget` hook + `aube_util::concurrency::effective_cpu_cap()`. Standalone aube installs no hook, so `effective_cpu_cap() == available_parallelism()` — byte-for-byte unchanged. Default-preserving, upstreamable (extends the #862 embedder profile).

## Stress matrix (Docker, cgroup v2, 10-core host) — all INSTALL_OK

| case | cpu_budget | workers / blocking / rayon |
|---|---|---|
| `--cpus=0.5` | None (std to 1) | 1 / 128 / 1 |
| `--cpus=2` | None (std to 2) | 2 / 128 / 2 |
| `--cpuset-cpus=0-1` | None (affinity to 2) | 2 / 128 / 2 |
| `--cpus=0.5 --pids-limit=80` | None min PID | 1 / 5 / 1 |
| unconstrained | None | 8 / 128 / 10 |
| `NUB_CPU_BUDGET=1` (free box) | Some(1) | 1 / 128 / 1 |
| `NUB_CPU_BUDGET=3` + `--cpus=2` | Some(3) override wins | 3 / 128 / 3 |

No EAGAIN/panic in any case; unconstrained sizing unchanged.

## Tests + gates

5 new unit tests in `resource_limits.rs` + 1 in `aube-util`. `clippy --all-targets --all-features -- -D warnings` and `fmt --check` clean; nub-cli bin suite 193 passing.

Refs #136

https://claude.ai/code/session_012YwF7yWjrXcj3bVJ1TziHK

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2618,6 +2618,7 @@ dependencies = [
  "miette",
  "mimalloc",
  "nub-core",
+ "num_cpus",
  "rayon",
  "rustc-hash",
  "semver",
@@ -2732,6 +2733,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "071dfc062690e90b734c0b2273ce72ad0ffa95f0c74596bc250dcfd960262841"
 dependencies = [
  "autocfg",
+]
+
+[[package]]
+name = "num_cpus"
+version = "1.17.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "91df4bbde75afed763b708b7eee1e8e7651e02d97f6d5dd763e89367e957b23b"
+dependencies = [
+ "hermit-abi",
+ "libc",
 ]
 
 [[package]]
--- a/crates/nub-cli/Cargo.toml
+++ b/crates/nub-cli/Cargo.toml
@@ -91,6 +91,12 @@ url = "2.5"
 # thread-create EAGAIN — the same exit-101 abort the tokio cap closes. We cap it
 # under the same headroom budget; unconstrained boxes keep rayon's default.
 rayon = "1"
+# CPU-budget detection (resource_limits::cpu_budget): num_cpus::get() reads the
+# cgroup CFS quota (cgroup v1 `cpu.cfs_quota_us`/`cfs_period_us` AND v2 `cpu.max`,
+# via /proc/self/mountinfo) and returns min(ceil(quota/period), logical_cores) —
+# exactly the automaxprocs algorithm, maintained, 2-dep. Closes the v1-quota +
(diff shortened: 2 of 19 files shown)


colinhacks:
Daily-driver failure is unrelated to this change. The `Run daily-driver smoke` step (the real install) passed; the failing step is `Post Run actions/setup-node` — a cache-save path-validation error (`Path(s) specified in the action for caching do(es) not exist`), which is #138's target ("drop pnpm cache from daily-driver setup-node"). It fires on a branch cache miss, not on `main` (where the restore hits and no save is attempted). Touches no install/pool/budget code. Merging #138 first should clear it.

colinhacks:

PR #152, opened 2026-06-25
feat(runtime): transpile project-source plain JS (.js/.mjs/.cjs)

colinhacks:
Extends nub's transpile pipeline to project-source plain JavaScript (`.js`/`.mjs`/`.cjs`) so transformable syntax lowers uniformly regardless of file extension — closing the JS/TS transform divergence where identical source behaved differently by extension (a project `.js` using `using` broke on the floor while the same code in `.ts` worked).

## What changed

- **`.js`/`.mjs`/`.cjs` added to `TRANSPILE_EXTS`.** Project plain JS now routes through the same pipeline as `.ts`, lowering the three syntaxes oxc transforms at nub's `es2022` target.
- **Native `transformable_syntax` skip-gate.** A verdict riding the existing `detect_module_info` parse flags a plain-JS file as needing transpile iff it contains any of the three triggers: `using`/`await using` (es2026), `v`-flag unicode-sets RegExp literals (es2024), or decorators. The set is pinned to oxc `0.132.0`'s es2022 lowering set with a re-derive-on-bump comment.
- **Verbatim no-op path.** When the verdict is false, the raw source is returned byte-for-byte — no codegen, no sourcemap footer, no cache write. A plain `.js` with nothing to lower is exactly the file you wrote (oxc otherwise reformats quotes/semicolons/whitespace).
- **node_modules excluded at all four dispatch sites.** Added the missing node_modules bails to the two compat-tier sites (the classic `require.extensions` shim and the async-hooks load); the two fast-tier sites were already gated. Without these, every dependency `.js` would be transpiled.
- **`.mjs`→module / `.cjs`→commonjs format branches** added to `moduleFormatFor`.

JSX-in-`.js` is out of scope (use `.jsx`).

## Tests

New `project-js` fixture suite, verified on both tiers (fast Node 22.15, compat 18.19): `using` down-levels; the `v`-flag RegExp lowers (the trigger the original design missed); decorator-in-`.js` routes to the Stage-3 diagnostic; no-op `.js` is byte-identical and resolves the correct format; `.mjs`/`.cjs` format correctness; node_modules deps (a plain dep + a pnpm `.pnpm`-store dep) are never transpiled. Full integration suite green (160 passed); root + nub-native clippy (`--all-targets --all-features -D warnings`) and fmt clean.

Refs the worker-relative-specifier-resolution thread — extending the Worker raw-string rewrite to `.js`/`.mjs`/`.cjs` is the trivial follow-up now that project JS transpiles.

This is a runtime-behavior change touching every project's JS handling; holding for maintainer review.

### DIFF
--- a/crates/nub-cli/tests/integration.rs
+++ b/crates/nub-cli/tests/integration.rs
@@ -296,6 +296,161 @@ fn legacy_decorators_require_experimental_flag() {
     );
 }
 
+// ── Project-source plain JS (.js/.mjs/.cjs) transpile ───────────────
+// nub transpiles PROJECT plain JS through the same pipeline as `.ts`, so
+// transformable syntax (`using`/`await using`, `v`-flag RegExp, decorators) lowers
+// uniformly — identical source must not behave differently by extension. A native
+// `transformableSyntax` verdict (riding the existing detect parse) gates a verbatim
+// skip-return for no-op JS so byte-parity is preserved; node_modules is excluded at
+// every dispatch site. See wiki/runtime/typescript.md + the transpile-project-js
+// design thread.
+
+#[test]
+fn project_js_using_down_levels() {
+    // `using` is a SyntaxError on every supported Node's V8; this project `.js`
+    // must be down-leveled (the transformableSyntax verdict flags it) and run, just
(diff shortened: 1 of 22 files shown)


colinhacks:

Issue #120, opened 2026-06-24
`pnpm` overrides and friends


innermatrix-cityblock:
According to the documentation at https://nubjs.com/docs/install#config-it-reads, the following `pnpm-workspace.yaml` fields are unsupported:

* `overrides`
* `packageExtensions`
* `patchedDependencies`

The more detailed docs at https://nubjs.com/docs/install/pnpm claim `overrides` is supported, which I reckon means the other two really aren't? This is blocking me from adopting `nub` at work.

colinhacks:
Fixed in v0.2.0 (now published): https://github.com/nubjs/nub/releases/tag/v0.2.0

PR #154, opened 2026-06-25
fix(pm): link yarn workspace members into the consumer's node_modules

colinhacks:
## What

Fix workspace-member linking under a yarn incumbent. With a classic `yarn.lock` present, `nub install` printed `✓ Already up to date` and never symlinked a local workspace member into `node_modules`, so a consumer member couldn't resolve it. pnpm and npm incumbents already linked the member correctly — the gap was yarn-incumbent-specific.

## Why

A classic `yarn.lock` is a flat resolution list with no importer structure and never records workspace members (they aren't registry packages). The classic reader (`vendor/aube/crates/aube-lockfile/src/yarn/classic.rs`) reconstructs member importers from the root `workspaces` globs, but built each importer's direct deps only by matching declared ranges against the flat resolution list. A member-to-member dependency has no `yarn.lock` entry, so it was silently dropped — leaving the linker with nothing to symlink.

The reader now also recognizes a dep whose name is a workspace sibling and emits a workspace-link `DirectDep` (a `dep_path` with no `packages` entry — the convention the linker keys on, mirroring the resolver's `try_workspace_link`). A real `yarn.lock` entry, when one exists, still wins. `yarn.lock` stays read-only (byte-identical).

## Tests (differential — fail on `main`, pass with the fix)

- `aube-lockfile` `yarn/tests.rs`: `test_classic_member_links_workspace_sibling`, `test_classic_workspace_protocol_and_range_gating` (parse layer — verifies the `DirectDep`/`dep_path` convention; covers `workspace:`/`link:`/satisfiable caret/unsatisfiable range).
- `nub-cli` `install_engine.rs`: `install_links_yarn_workspace_member_into_consumer` (end-to-end, no network — runs the real `nub install`, then `node require.resolve` from the consumer, and asserts `yarn.lock` is unchanged).

## aube fork-discipline

The change is default-preserving (it only adds importer edges the classic yarn reader was dropping; standalone aube on a non-workspace or pnpm/npm/bun path is untouched) — a latent-bug fix, upstreamable to `jdx/aube`.

Refs the #120 PM-field-support audit (`verify-pm-field-support`).

### DIFF
--- a/crates/nub-cli/tests/install_engine.rs
+++ b/crates/nub-cli/tests/install_engine.rs
@@ -468,3 +468,81 @@ fn prefer_offline_does_not_trip_the_yarn_offline_mirror_fatal() {
     // by the yarn gate / install path, never the mirror preflight.
     let _ = code_prefer;
 }
+
+/// Workspace-member linking under a YARN incumbent. A classic yarn.lock never
+/// records workspace members (they aren't registry packages), so a member that
+/// depends on a SIBLING member has no resolution entry. nub must still symlink
+/// the sibling into the consumer's node_modules so it resolves — matching what
+/// reference yarn does (and what nub already does under pnpm/npm incumbents).
+///
+/// No network: every dep is a local member, the yarn.lock is empty, and the
+/// install is a frozen read (yarn.lock left byte-identical). Before the fix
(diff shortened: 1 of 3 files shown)


colinhacks:

PR #155, opened 2026-06-25
fix(pm): re-read dist-tag specs live when the primer served the pick

### LINKED ISSUE #135: Discrepency between nub install and nub update
### Environment
* nub version: 0.2.0 (same issue happens with 0.1.14)
* nodejs version: 26.3.1
* Platform: Windows 11 25H2

### Steps to reproduce:
1. Create empty directory with this package.json:
```json
{
    "name": "test",
    "version": "1.0.0",
    "private": true,
    "type": "module",
    "dependencies": {
        "vite": "latest"
    }
}
```
2. Run `nub update`, notice vite 8.0.16 is installed. Output:
```
Updating all dependencies...
  vite: (new) 8.0.16
Resolved 46 packages
Wrote lock.yaml
dependencies:
+ vite@8.0.16

nub 0.2.0 · ✓ installed 15 packages in 333ms
```
4. Run `nub outdated`, result:
```
Package  Current  Wanted  Latest
vite     8.0.16   8.0.16  8.1.0
```
4. Run `nub install`, result: `✓ Already up to date`. User is stuck at the non-latest version.
5. Repeat from step 1, but instead of `nub update` first, do `nub install` first.
6. Notice that this time vite 8.1.0 is correctly installed first. Output:
```
dependencies:
+ vite@8.1.0

nub 0.2.0 · ✓ installed 15 packages in 880ms
```
8. Follow up with `nub update`, the 8.1.0 version is downgraded to 8.0.16.
### Expected result
Both `nub install` and `nub update` should install vite 8.1.0 per package.json specification and current npmjs version table.

### Actual result
`nub update` is installing an outdated version.

colinhacks:
A dist-tag (`latest`/`next`) is a mutable pointer, but the metadata primer bakes its value at build time. The freshness gate keyed refetch on `classify_regime` (version position in history), which can't see the pointer moved: a stale baked `latest` is the primer frontier, classifies `Current`, and absent a cutoff was served offline. So a 0.2.0 binary kept resolving `"vite": "latest"` to 8.0.16 after the registry moved it to 8.1.0; `update` served the stale pick while `install` refetched, so they diverged and `update` downgraded.

Fix: refetch any primer-seeded pick whose range is a dist-tag (matching npm/pnpm). Frozen range picks still serve offline.

Closes #135

### DIFF
--- a/vendor/aube/crates/aube-resolver/src/resolve/driver.rs
+++ b/vendor/aube/crates/aube-resolver/src/resolve/driver.rs
@@ -26,7 +26,10 @@ use crate::local_source::{
 use crate::package_ext::{
     apply_package_extensions, apply_package_extensions_to_deps, pick_override_spec,
 };
-use crate::semver_util::{PickResult, Regime, classify_regime, pick_version, version_satisfies};
+use crate::semver_util::{
+    PickResult, Regime, classify_regime, pick_version, range_resolves_via_dist_tag,
+    version_satisfies,
+};
 use crate::{
     Error, ExoticSubdepDetails, FxHashMap, FxHashSet, ResolutionMode, ResolveTask, ResolvedPackage,
     Resolver, error, is_deprecation_allowed, is_supported,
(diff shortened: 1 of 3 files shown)


colinhacks:

PR #156, opened 2026-06-25
feat(install): link directories.bin as a fallback when top-level bin is absent

colinhacks:
## Gap

When a package declares no top-level `bin` but sets `directories.bin`, `nub install` linked nothing. The dominant reference PMs all link it; nub matched only the outlier (berry).

| PM | links `directories.bin`? |
|----|--------------------------|
| npm 11.13 | yes |
| pnpm 10.15 | yes |
| yarn classic 1.13 | yes |
| bun 1.3.14 | yes |
| yarn berry 4.9 | no |
| nub 0.2.1 (before) | no |

## Fix

Mirrors pnpm's `getBinsFromPackageManifest`: when `bin` is absent and `directories.bin` present, every regular file under that dir (recursive, files only, symlinks not followed) links as a bin named `basename(file)`. `bin` wins; `directories.bin` is the fallback only. Wired into all three bin sites (direct, workspace, bundled dep).

Containment: the bin dir must canonicalize strictly inside the package root (rejects `../escape`, absolute, symlink-elsewhere, and `.`, matching pnpm's `isSubdir`); each basename passes `validate_bin_name`; the walk does not follow symlinks.

Tests fail on pre-fix main, pass after. Pure additive parity fix; upstreamable to jdx/aube as-is.

### DIFF
--- a/vendor/aube/crates/aube/src/commands/install/bin_linking.rs
+++ b/vendor/aube/crates/aube/src/commands/install/bin_linking.rs
@@ -537,7 +537,11 @@ fn link_dir_bins(
     ) else {
         return Ok(());
     };
-    if !canon_bins.starts_with(&canon_root) {
+    // Reject both an escape (`../x`, absolute, symlink-elsewhere) AND
+    // the package root itself: pnpm's `isSubdir(pkgPath, binDir)` is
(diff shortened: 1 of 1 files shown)


colinhacks:

PR #159, opened 2026-06-25
perf(runtime): drop redundant work on the JS transpile path

jdalton:
Removes two pieces of provably-redundant work on the JS warm transpile path (runtime/transform-core.mjs).

One detectModuleInfo parse instead of two. For an ambiguous-extension .ts/.tsx/.jsx (no explicit package.json#type) whose source contains @, loadTranspile parsed the source twice: once via moduleFormatFor (reading hasValueEsmSyntax) and again via hasDecoratorSyntax (reading hasDecorators). The native detectModuleInfo already returns both fields from a single oxc parse (crates/nub-native/src/detect.rs). loadTranspile now parses once via a moduleFormatWithInfo helper and reuses that result's hasDecorators for the Stage-3 guard. The no-parse short-circuits (.mts/.cts/explicit type) are unchanged — there the guard still does its own single parse, so it is never more than one detect per file. moduleFormatFor's public contract is untouched.

O(1) builtin check. Three module.builtinModules.includes(...) (a linear ~80-name scan) become module.isBuiltin(...), matching the idiom already used in preload-common.cjs. Pure swap, no semantic change.

This is a redundancy cleanup on the warm path, secondary in scope. It is microsecond-scale and does NOT move nub's launcher-dominated wall-clock — no speedup is claimed.

Verification: the transpilation / module-format / decorator integration suite passes unchanged (no behavior change), plus a standalone harness confirms detectModuleInfo is invoked at most once per file on the load path — including the ambiguous-ext + @ case that previously parsed twice.

### DIFF
--- a/runtime/transform-core.mjs
+++ b/runtime/transform-core.mjs
@@ -342,7 +342,7 @@ export function resolveSpec(specifier, parentURL) {
   // from a nub-dependency ESM entry) delegated to a strict user loader is exactly
   // the R11 leak. See isNubInternalParent.
   if (isNubInternalParent(parentURL)) {
-    if (specifier.startsWith("node:") || module.builtinModules.includes(specifier)) {
+    if (specifier.startsWith("node:") || module.isBuiltin(specifier)) {
       const url = specifier.startsWith("node:") ? specifier : `node:${specifier}`;
       return { url, shortCircuit: true };
     }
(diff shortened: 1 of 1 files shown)


colinhacks (approved):

colinhacks:

PR #158, opened 2026-06-25
test(wpt): expand WPT slice to urlpattern + web-locks + reporterror

colinhacks:
Adds 18 WPT files covering nub's polyfilled surface: URLPattern (3 active + 2 tentative-skip), Web Locks (12 active + 1 non-secure-context skip), reportError (skip).

`run-file.mjs`: `location`, `isSecureContext=false`, and `fetch`→local-file shims. The `location` shim also fixes two stale "Untitled" expected-fail names in existing webmessaging entries.

`status.json` tier attributions are honest against the real floor legs — #163 now pins each matrix leg to its own Node, and #164 backfills `navigator` + spec-faithful Web Locks to the 18.19 floor:

- URLPattern is native only on Node 24+; on 18/20/22 nub loads urlpattern-polyfill, whose 22 regex-divergences fail identically across those polyfill tiers (one versioned band, Node ≤ 23). The earlier Node-24 native bugs were fixed in 24.17, the version the floating `24` leg installs, so those bands are removed.
- Web Locks passes fully on every tier — including `signal`/`steal` on Node 22 — via #164's navigator-locks. Cross-worker `query()` subtests stay expected-fail (single-process locks); reportError stays skipped (async dispatch, no global ErrorEvent).

Verified on real Node 18.19 / 20.19 / 22.15 / 24.17: 0 unexpected-fail and 0 unexpected-pass on every leg.

PR #163, opened 2026-06-25
test(wpt): pin the WPT Worker matrix to the floor Node it claims to test

colinhacks:
The WPT harness spawns nub with cwd = repo root, so nub reads the root `engines.node >=22.15.0` and, on the floor legs (18.19, 20), silently provisions a >=22.15 Node — the "18.19 leg" actually ran a fast-tier Node, making the compat-floor coverage vacuous.

Fix: pin `NODE_EXECUTABLE=process.execPath` on the spawned nub children. `NODE_EXECUTABLE` is nub's existing override (bypasses pin-file/engines resolution; the floor check still applies); the harness runs under the leg's Node, so the children now run that exact Node.

Reproduced: floor 18.19 + `engines >=22.15.0` ran v26.2.0 without the pin, v18.19.0 with it.

Refs #158

### DIFF
--- a/runtime/polyfills.cjs
+++ b/runtime/polyfills.cjs
@@ -84,6 +84,63 @@ function installSyncPolyfills(preloaded) {
     });
   }
 
+  // ── File (global on Node 20+, missing on the 18.x compat floor) ─────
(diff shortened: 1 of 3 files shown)


colinhacks:

PR #169, opened 2026-06-26
feat(nix): add prebuilt-binary flake (nix run github:nubjs/nub)

### LINKED ISSUE #167: feat: Add Nix package manager flake support
## Add Nix flake support

**Proposal**: Add Nix flake support to enable installation via `nix run github:nubjs/nub` or `nix profile install github:nubjs/nub`.

## Motivation

Nix flakes provide a reproducible, declarative way to manage software installations. Adding flake support would:

- Enable Nix users to install nub with a single command: `nix run github:nubjs/nub`
- Provide reproducible builds across different systems
- Support NixOS users and development environments
- Align with modern Rust project conventions (many Rust CLIs now ship with flake.nix)

## Implementation

I've implemented a minimal flake.nix that:

- Uses `rustPlatform.buildRustPackage` for standard Rust builds
- Includes proper macOS framework dependencies (Security, CoreFoundation)
- Provides both package and app outputs for `nix run` and `nix profile install`
- References the upstream repository correctly for installation commands
- Includes a dev shell for development

The implementation has been tested with `nix flake check` and passes validation.

## Changes

- **Added**: `flake.nix` - Nix flake configuration
- **Added**: `flake.lock` - Lock file for reproducible builds
- **Modified**: `README.md` - Added Nix installation instructions

## Installation

Users can install nub via Nix with:

```bash
nix run github:nubjs/nub
nub profile install github:nubjs/nub
```

## Testing
The flake has been validated with nix flake check which confirms:

- Proper flake structure
- Valid derivation evaluation
- Correct outputs (packages, apps, devShells)

## Alternatives Considered
- Devbox: The project already uses devbox.json, but Nix flakes are more widely adopted and provide better integration with the Nix ecosystem
- No Nix support: Would exclude Nix/NixOS users from easy installation
Additional Notes

This is a minimal implementation that can be extended later if needed (e.g., cross-compilation, caching optimizations, etc.). The current setup prioritizes simplicity and correctness.

colinhacks:
Adds a Nix flake so Nix/NixOS users can install nub via `nix run github:nubjs/nub` (or `nix profile install github:nubjs/nub`).

## Build model: prebuilt binary, not from-source

The flake `fetchurl`s nub's existing per-platform GitHub release tarballs and lays them out — it does not compile from source. Every other nub channel (curl installer, Homebrew, npm) ships these prebuilt tarballs, and a from-source `rustPlatform.buildRustPackage` would omit the vendored `runtime/` tree (preload + `node_modules` + the `nub-native.node` N-API addon) that nub resolves beside its binary at run time — a bare cargo binary passes `--version` but fails real TypeScript workloads. The prebuilt tarball already carries the correct `bin/` + `runtime/` layout.

## What's here

- **`flake.nix`** — `packages.<system>.default` + `apps.<system>.default` across the four Nix-supported targets (`x86_64`/`aarch64` × `linux`/`darwin`). Each `fetchurl`s the matching v0.2.2 tarball (sha256 pinned from the release asset digests) and lays out a real `bin/nub` (+ `bin/nubx`) with the `runtime/` sibling exactly as shipped — no wrapper or symlinked entrypoint, since nub canonicalizes `current_exe()` and walks up to find `runtime/preload.mjs`. Linux uses `autoPatchelfHook` (+ `stdenv.cc.cc.lib` for `libgcc_s`) to rewrite the glibc interpreter/rpath; Darwin needs no patching.
- **`flake.lock`** — pins nixpkgs (`nixpkgs-unstable`).
- **`.github/workflows/nix.yml`** — path-filtered CI (fires only on flake/`*.nix` changes) that runs `nix flake check --all-systems`, `nix build .#default`, and `nix run .#default -- --version` on an Ubuntu runner via `DeterminateSystems/nix-installer-action`. There is no Nix on the dev host, so this job is the flake's validation gate (it exercises `fetchurl` + `autoPatchelf` + exec end to end).
- **README** — a `nix run github:nubjs/nub` line in the Install block.

The win32/musl targets are excluded — Windows isn't a Nix target, and the glibc tarballs cover the Linux Nix systems.

## Fast-follow (not in this PR)

A prebuilt flake pins version + sha per release, so a future change should auto-bump `flake.nix`'s `version` + the four sha256s from the published `.sha256` sidecars on each release (the shape of the existing `bump-homebrew-tap` job, but committing in-repo). Deferred here because it mutates `main` post-publish and can't be exercised by this PR's CI; the flake is correct pinned to v0.2.2 in the meantime.

Closes #167

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -1,0 +1,60 @@
+name: Nix flake
+
+# Validates the prebuilt-binary flake (flake.nix) that lets Nix users install via
+# `nix run github:nubjs/nub`. There is no Nix on the dev host, so THIS job is the
+# flake's only validation: it evaluates every system's outputs, builds the package
+# for the runner's own system (fetchurl + autoPatchelf), and smoke-runs the binary.
+#
+# Path-filtered to the flake files so it fires only when they change — not on every
+# Rust or docs commit.
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - "flake.nix"
+      - "flake.lock"
+      - "**/*.nix"
+      - ".github/workflows/nix.yml"
+  pull_request:
+    branches: [main]
+    paths:
+      - "flake.nix"
+      - "flake.lock"
+      - "**/*.nix"
+      - ".github/workflows/nix.yml"
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  check:
+    name: nix flake check
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    steps:
+      - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5
+
+      - name: Install Nix
+        uses: DeterminateSystems/nix-installer-action@v16
+
+      - name: Cache Nix store
+        uses: DeterminateSystems/magic-nix-cache-action@v8
+
+      # Evaluate every system's outputs (the locked flake must evaluate for all
+      # four targets, including the Darwin ones that cannot build on this runner).
+      - name: nix flake check
+        run: nix flake check --all-systems --print-build-logs
+
+      # Build the package for x86_64-linux: fetchurl the release tarball, run
+      # autoPatchelfHook against the glibc binary + nub-native.node, lay out the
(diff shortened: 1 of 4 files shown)


colinhacks:
Shipped in v0.2.3: https://github.com/nubjs/nub/releases/tag/v0.2.3

PR #170, opened 2026-06-26
fix(runtime): neutralized localStorage is absent, not present-undefined

### LINKED ISSUE #166: `localStorage` is undefined in React project with vitest using `nub run test`
Script cmd: `"test": "vitest run --config ./vitest.config.ts"`
`nub run test` results in tests failing due to localStorage being undefined. `npm run test` doesn't have this issue

> TypeError: Cannot read properties of undefined (reading 'getItem')
 ❯ Object.get src/utils/authorization/resources.ts:11:38
      9|
     10|   export function get(): Resource[] {
     11|     const currentJson = localStorage.getItem('resources');
       |                                      ^
     12|
     13|     if (currentJson === lastResourcesJson) return lastResources;

vitest.config.ts
```ts
export default defineConfig({
    test: {
        isolate: true,
        environment: 'happy-dom',
        restoreMocks: true,
        clearMocks: false,
        mockReset: false,
        testTimeout: 10000,
        setupFiles: ['__mocks__/set-env-vars.ts', '__mocks__/setup-tests.tsx'],
        css: false,
        include: ['src/**/*.test.{js,jsx,ts,tsx}'],
        exclude: ['node_modules', 'dist', 'coverage', '__mocks__']
    },
    resolve: {
        tsconfigPaths: true,
    },
})
```

colinhacks:
On the Node 22.4–24 band, nub neutralized the throwing `localStorage` getter by redefining it present-but-undefined, leaving `'localStorage' in globalThis === true`. Libraries gating on that check (vitest happy-dom) skipped installing their own store, so user code read nub's `undefined` and crashed. Vanilla Node 24 has it absent.

Fix: delete the getter inside the existing `__NUB_NEUTRALIZE_LOCALSTORAGE` gate so the global is absent, matching vanilla Node 24. `sessionStorage` still works; the `--localstorage-file` path and Node 25/26 are unaffected. The env var is retained to scope neutralization to what nub causes.

Adds a band-gated regression test.

Closes #166

### DIFF
--- a/crates/nub-cli/tests/integration.rs
+++ b/crates/nub-cli/tests/integration.rs
@@ -1608,6 +1608,78 @@ fn sessionstorage_works_by_default_localstorage_needs_user_file() {
     let _ = std::fs::remove_dir_all(&dir);
 }
 
+/// On the flag-needed band (22.4–24.x), the neutralized `localStorage` must be
+/// ABSENT (`'localStorage' in globalThis === false`), matching vanilla Node 24 —
+/// NOT present-but-undefined. A present-but-undefined global broke isomorphic
+/// libraries that gate on `'localStorage' in window/globalThis` (vitest's happy-dom
+/// `getWindowKeys`): the present property made them SKIP installing their own
+/// store, so user code read nub's `undefined` and crashed (#166). nub now `delete`s
+/// the throwing getter inside the neutralize gate, restoring the absent shape so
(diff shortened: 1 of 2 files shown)


colinhacks:
Shipped in v0.2.3: https://github.com/nubjs/nub/releases/tag/v0.2.3

PR #175, opened 2026-06-26
Single self-contained binary (embed runtime, JIT-extract on first run)

colinhacks:
Ships nub as one file: the runtime/ tree (preload + vendored node_modules + native addon) is embedded as a zstd-19 blob and JIT-extracted once, on first run, to ~/.cache/nub/runtime-<ver>-<hash>/. Steady state unchanged (one stat).

Release-only `embed-runtime` cargo feature, off by default — the dev loop walks to the in-repo runtime/ as before. build.rs makes the blob (include_bytes!); runtime_cache.rs extracts it (atomic tmp+rename, RO-FS to $TMPDIR fallback, age-based GC). find_preload routes through it; downstream resolution unchanged.

Tradeoff: binary +~2.1-2.7 MB/platform; npm package -~8 MB net; first run +~35-90 ms once.

Open item: Windows first-run DLL write (SmartScreen/AV), verifiable only on the Windows CI leg.

Held for maintainer review — architectural, not part of the patch release. Full detail in the commit message.

### DIFF
--- a/.github/scripts/gen-homebrew-formula.sh
+++ b/.github/scripts/gen-homebrew-formula.sh
@@ -78,26 +78,18 @@ class Nub < Formula
   end
 
   def install
-    # The release archive is a tree (bin/nub, bin/nubx, runtime/), not a bare
-    # binary: nub loads runtime/ (preload + vendored polyfills + native addon)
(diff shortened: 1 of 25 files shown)


colinhacks:

PR #180, opened 2026-06-26
fix(pm): reject registry bin paths that escape the package dir

### LINKED ISSUE #172: 🔒 Potential path traversal via unvalidated `dist.bin_subpath`
🔒 **Security** · 🟠 High · Confidence: 96%

**File:** `crates/nub-core/src/pm/provision.rs`
**Location:** `install`

---

### What's wrong

The code constructs the executable path by directly joining a user‑provided value: `let bin = final_dir.join("package").join(&dist.bin_subpath);`. If `dist.bin_subpath` contains `..` or an absolute path, the resulting `bin` could escape the intended package directory, allowing execution of arbitrary files extracted from a malicious tarball.

### Suggested fix

Validate `dist.bin_subpath` before joining, ensuring it is a relative path without parent components. Example:
```rust
let bin_subpath = Path::new(&dist.bin_subpath);
if bin_subpath.is_absolute() || bin_subpath.components().any(|c| matches!(c, std::path::Component::ParentDir)) {
    bail!("invalid bin subpath in dist: {}", dist.bin_subpath);
}
let bin = final_dir.join("package").join(bin_subpath);
```

---

<details>
<summary>About this report</summary>

This finding was generated by an automated audit tool using Llama 3.3 70B + verification passes.
Only findings with ≥92% confidence that passed both LLM self-verification and line reference
verification are reported. False positives are still possible — please verify before acting.

</details>

colinhacks:
The runnable PM bin is built as `<store>/<version>/package/<bin_subpath>` and executed. Since `Path::join` discards the base on an absolute component, an absolute (`/bin/sh`) or `..`-laden registry `bin` escaped the package dir, gated only by `is_file()`.

`registry::bin_subpath` and `named_bin_subpath` now reject a path with a `Prefix`/`RootDir`/`ParentDir` component, covering all three execution consumers and the shim with one guard. The component scan also catches Windows drive-/root-relative entries `is_absolute()` misses. Rejected bins surface as the existing error — no panic, no rewrite.

Defense-in-depth: requires a compromised/MITM'd registry.

Closes #172

### DIFF
--- a/crates/nub-core/src/pm/registry.rs
+++ b/crates/nub-core/src/pm/registry.rs
@@ -10,7 +10,7 @@
 //! `dist.shasum` (sha1) is the fail-closed fallback for ancient publishes that
 //! predate `integrity`.
 
-use std::path::PathBuf;
+use std::path::{Component, PathBuf};
 
 use anyhow::{Context, Result, bail};
 use serde_json::Value;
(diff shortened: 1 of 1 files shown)


colinhacks:

PR #181, opened 2026-06-26
fix(cache): fold filename into transpile cache key (#171)

### LINKED ISSUE #171: 🐛 Cache key omits filename leading to incorrect sourceURL on cache hit
🐛 **Bug** · 🟠 High · Confidence: 96%

**File:** `crates/nub-native/src/cache.rs`
**Location:** `transform_cached`

---

### What's wrong

The cache key is derived without the `filename` component: `let key = cache_key(&source, &ext, &tsconfig_hash, &pkg_type);`. Because the source URL comment (`//# sourceURL={filename}`) is appended after transformation, two different files with identical source content will share the same cache entry, causing the cached code to embed the wrong filename in the sourceURL comment on subsequent loads.

### Suggested fix

Include `filename` in the cache key generation and update the `cache_key` function signature accordingly. Example:
```rust
// In transform_cached
let key = cache_key(&filename, &source, &ext, &tsconfig_hash, &pkg_type);
```
And modify the helper:
```rust
fn cache_key(filename: &str, source: &str, ext: &str, tsconfig_hash: &str, pkg_type: &str) -> String {
    nub_cache_key::cache_key(
        NUB_VERSION,
        CACHE_SCHEMA,
        BUILD_ID,
        filename,
        source,
        ext,
        tsconfig_hash,
        pkg_type,
    )
}
```
Adjust the `nub-cache-key` crate or its wrapper to accept the additional `filename` parameter, ensuring the hash incorporates the file path and prevents cross‑file cache collisions.

---

<details>
<summary>About this report</summary>

This finding was generated by an automated audit tool using Llama 3.3 70B + verification passes.
Only findings with ≥92% confidence that passed both LLM self-verification and line reference
verification are reported. False positives are still possible — please verify before acting.

</details>

colinhacks:
The transpile cache key omitted `filename`, but the cached body bakes in a per-file `//# sourceURL`. Two byte-identical sources in one directory collided on one entry, so the second was served the first's `sourceURL` — misattributing V8 stack frames, even cold within one process.

Fold `filename` into the cache-key preimage and bump `CACHE_SCHEMA` 5 → 6 (one-time cold-cache miss on upgrade). Adds a regression test; verified e2e that identical files now get distinct entries + sourceURLs cold and warm.

Closes #171

### DIFF
--- a/crates/nub-cache-key/src/lib.rs
+++ b/crates/nub-cache-key/src/lib.rs
@@ -3,18 +3,26 @@
 //! test can't live in the cdylib `nub-native`).
 //!
 //! The key preimage is NUL-separated and order-fixed (no trailing separator):
-//!   `nub_version \0 schema \0 build_id \0 source \0 ext \0 tsconfig_hash \0 pkg_type`
+//!   `nub_version \0 schema \0 build_id \0 source \0 ext \0 tsconfig_hash \0 pkg_type \0 filename`
 //! → blake3 → 64-hex lowercase → the cache FILENAME. Every component is hashed
 //! IN, so a change to ANY of them (a nub release, a CACHE_SCHEMA bump, a rebuild
 //! at a new build-id) yields a disjoint filename — old on-disk entries are
 //! silently ignored (a miss), never mis-read. `nub-native` calls `cache_key` with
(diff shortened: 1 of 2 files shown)


colinhacks:

PR #182, opened 2026-06-26
feat(embed): verify-on-load integrity + safe-extraction hardening (R1+R2)

colinhacks:
Hardens the single-binary JIT-extracted runtime (the `embed-runtime` feature) against a planted, corrupt, or tampered on-disk cache. Stacked on `single-binary-embed` — the diff here is only the integrity change.

The embed feature moves the trusted-on-disk runtime from a sidecar `runtime/` to a versioned cache dir that the child `node` `--require`s and `dlopen`s. The cache path is public and precomputable from the released binary, so a world-writable base or an unverified warm load is a local code-load surface. This closes both.

## R1 — safe extraction location/permissions (enforced, recover-never-brick)

- The cache base is created `0700`; before a pre-existing base is adopted it is validated (unix): owner == euid, no group/world write bit, not a symlink. An unsafe base is skipped — not used, not destroyed — and extraction recovers to the next safe candidate. It never bricks.
- The `$TMPDIR` fallback is now per-user (`$TMPDIR/nub-<uid>`, `0700`), closing the one genuinely-new world-writable `/tmp` code-load surface the embed feature added.
- Windows relies on the per-user profile ACL (`%USERPROFILE%\.cache`, `%TEMP%`); the unix stat checks are skipped there.

## R2 — verify the loaded entrypoints against a baked-in digest (canary-first)

- `build.rs` bakes the BLAKE3 digest of the three directly-loaded entrypoints (`preload.mjs`, `preload.cjs`, `addons/nub-native.node`) into the binary. On the load path — once per process via `OnceLock` — the extracted files are re-hashed against those consts.
- On mismatch: self-heal — re-extract the trusted in-binary blob over the dir and re-verify. A persistent post-re-extract mismatch is a non-fatal warning that proceeds (canary), gated by `const INTEGRITY_ENFORCE: bool = false` (a one-line flip to fail-closed once the wild mismatch rate is confirmed ~0). The maintainer-chosen posture: a verify-on-load bug must not brick nub on day one.
- The digests live inside the (signed) binary, so a tampered on-disk file can't swap its own expected hash alongside it.

### Why BLAKE3, not SHA-256

The verify re-hashes the ~5–9 MB addon on every warm load. Software SHA-256 measured ~28 ms on aarch64 (Apple Silicon has no auto HW-SHA in the `sha2` crate, and the `asm` path won't build on win-arm64/musl) versus ~6 ms for BLAKE3's portable SIMD — a per-invocation startup cost on the primary dev platform. BLAKE3 is collision/preimage-resistant (adequate for integrity), builds identically on every target, and is already a dependency. The 32-bit cache key stays SHA-256 (not on the hot path).

Integrity notices use `eprintln!` rather than `tracing::warn!`: nub's default subscriber only enables `aube*` targets, so a `nub_core` tracing warning is swallowed without `RUST_LOG` — which would defeat the canary's "watch for real-world warning reports" purpose.

## Tests + CI

- `runtime_cache` unit/integration tests against the real embedded blob: clean extraction verifies (zero false-positive), a tampered entrypoint is detected and self-healed, a persistent mismatch is canary by default and refuses under `INTEGRITY_ENFORCE` (test override), the build.rs hash is deterministic (baked == extracted), and the R1 0700/owner/world-writable/symlink validation + recovery.
- The `embed-runtime` CI job becomes a Windows/macOS/linux-gnu matrix, plus a linux-musl leg, all running the verify suite against the real blob — a verify false-positive would brick nub on that platform, so the matrix is ci-gate-required.
- The Windows and a new macOS round-trip job add R2 assertions (zero-false-positive on a clean run; tamper → self-heal → restore) on the real addon `dlopen` path.

Verified end-to-end on a real release binary (real addon embedded): cold extract → verify → dlopen → execute, warm reuse, self-heal on a tampered addon, and R1 recovery from a deliberately world-writable base.

Refs #175.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -180,45 +180,97 @@ jobs:
         run: tests/brand-lint/check-env-reads.sh
 
   # The single-binary `embed-runtime` feature is OFF in every default build, so the
-  # feature-off `test` job never RUNS the runtime_cache extraction/race/GC unit
-  # tests, and build.rs's tar+zstd blob arm + the tar/zstd/sha2 build-dep wiring go
+  # feature-off `test` job never RUNS the runtime_cache extraction/race/GC/INTEGRITY
+  # tests, and build.rs's tar+zstd blob arm + per-entrypoint hash baking go
   # unexercised on PRs (they would first break at a release tag). Build + test the
-  # feature-on path here so an embed/extract regression fails a PR. One platform
-  # suffices for the logic; the tracked runtime/*.mjs satisfy build.rs's staging
-  # probe with no extra assembly.
+  # feature-on path here so an embed/extract/verify regression fails a PR.
+  #
+  # CRITICAL: the R2 verify-on-load suite (zero-false-positive on a clean extraction,
+  # tamper→self-heal, canary/enforce decision, R1 0700/owner/symlink validation,
+  # build.rs hash determinism) runs against the REAL embedded blob — and a verify
+  # FALSE-POSITIVE would BRICK nub on that platform. So it runs on every hosted OS
+  # (Windows path semantics + macOS + linux all differ), is ci-gate-required, and the
+  # musl divergence is covered by the sibling `embed-runtime-musl` job. The addon is
+  # gitignored (built locally / staged in release.yml), so a placeholder is staged so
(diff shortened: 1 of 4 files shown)


colinhacks:

PR #185, opened 2026-06-26
feat(cli): add shell completion generation

kalicyh:
Adds `nub completion <shell>` as the source command for install channels to generate shell completions.

Homebrew can then wire this up with `generate_completions_from_executable`, while manual installs can still redirect the generated script if needed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -804,6 +804,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "clap_complete"
+version = "4.6.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e0a7a9bfdb35811f9e59832f0f05975114d2251b415fb534108e6f34060fd772"
+dependencies = [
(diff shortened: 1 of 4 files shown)


colinhacks:
Thanks for this — we do want shell completions in nub. We're going to land them against a fuller design we've already sketched: the load-bearing piece is *dynamic* completion of `package.json` script names for `nub run <TAB>` and bare `nub <TAB>` (workspace-aware, computed live per-TAB), plus `.bin` names for `nubx` — which static `clap_complete` output can't do on its own. So this'll take a little longer to do right rather than ship static-only. Keeping it open and crediting your work as we build on it.

kalicyh:
Sounds good, that makes sense. Dynamic `package.json` script and `.bin` completion is the part users will actually feel, so I agree static-only is not the final shape.

I’m really looking forward to that version.

colinhacks:
Closing after all, but I'll leave a note here when our official version lands.

PR #186, opened 2026-06-26
build(nix): from-source flake with embed-runtime (drop prebuilt tarball)

colinhacks:
#175 embeds the `runtime/` tree into the binary (JIT-extracted on first run), so a from-source build is self-contained — the flake no longer needs the prebuilt tarball.

`flake.nix` builds from source: a derivation builds the N-API addon (`crates/nub-native`), a `runCommand` lays out the runtime `node_modules` (4 deps + transitive `jsbi`, versions/SRI pinned to `pnpm-lock.yaml`), and `buildRustPackage` stages both into `runtime/` before building `nub-cli --features embed-runtime`. `nubx` is a symlink; `version` reads from `Cargo.toml`. `nix.yml` gains a cold-run `.ts` smoke.

No Nix on the dev host: verified by this PR's Nix CI.

Refs #175

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -1,9 +1,14 @@
 name: Nix flake
 
-# Validates the prebuilt-binary flake (flake.nix) that lets Nix users install via
-# `nix run github:nubjs/nub`. There is no Nix on the dev host, so THIS job is the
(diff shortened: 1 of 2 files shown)


colinhacks:

PR #187, opened 2026-06-26
fix(install): silence the unactionable GVS-incompatible auto-fallback notice

colinhacks:
On `nub i`, an incompatible dep (Next.js, Nuxt, Parcel, SvelteKit) forces per-project materialization and emitted a loud `WARN_NUB_GVS_INCOMPATIBLE` every run. The fix is upstream in the package, so it is unactionable; this silences it by default.

A new embedder-fixed `gvs_incompatible_warning` toggle defaults `true` (standalone aube unchanged); the nub profile sets it `false`, demoting the notice to `debug` — hidden under nub's `aube=warn` filter, reachable via `RUST_LOG=aube=debug`. Per-project fallback behavior is unchanged; only the log level differs.


colinhacks:

PR #189, opened 2026-06-26
fix(shim): don't refuse a global install on a cross-PM project pin

colinhacks:
In a project pinning a different PM (`packageManager: "pnpm@…"`), the shims refuse a foreign-PM **global** install (`npm install -g …`): "would write a competing lockfile and node_modules". A global install writes the global prefix, not the project — that rationale doesn't apply.

Fix: a `global` signal lets a global mismatch fall through to the system PM (like the transparent/nested escapes); the local refusal is unchanged. `is_global_invocation` matches bare `-g`/`--global` (npm/pnpm) or yarn's `global` subcommand — never `--global=false` (an explicit local install).

corepack has no cross-PM refusal to mirror. Tests: unit + `pm_shim` integration cover global-passes-through and local-still-refuses.

### DIFF
--- a/crates/nub-cli/src/cli.rs
+++ b/crates/nub-cli/src/cli.rs
@@ -6569,11 +6569,17 @@ fn shim_plan(
     // top-level invocation (no marker) keeps full strict behavior.
     let nesting = Nesting::from_env(|k| env::var(k).ok());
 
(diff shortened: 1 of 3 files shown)


colinhacks:

PR #183, opened 2026-06-26
feat(install): forward pnpm reporter/silent/loglevel flags to quiet the install

colinhacks:
Closes #179

Adds pnpm's output flags to the install family (install/ci/add/remove/update/import/dedupe), mapped to the engine's text-mode renderers:

- `--silent` / `-s` / `--reporter=silent` / `--loglevel=silent` → nothing on stderr but a fatal error (matches `pnpm install --silent`).
- `--reporter=append-only` → plain output, no progress object.
- `--loglevel <error|warn|info|debug>` → retune verbosity. Default unchanged.

The silencer drops before the error path, so `--silent` still prints errors. `--reporter=ndjson` not accepted (separate feature). Additive, default-preserving aube change. Grammar + network tests; docs updated.

PR #191, opened 2026-06-26
fix(install): honor --silent/--reporter/--loglevel before the verb too

colinhacks:
## What

#183 wired the PM output flags (`--reporter=<default|append-only|silent>`, `--silent`/`-s`, `--loglevel`) onto every install-family verb in the per-verb position (`nub install --silent`). The same flags **before** the verb were broken:

- `nub --silent install` ran noisily (the global `--silent` only drove `nub run`'s preamble).
- `nub --reporter=silent install` fell through the dispatch scan to the file runner → `node: bad option: --reporter=silent`, exit 9, install never ran.

The originally-reported per-verb P0 was already fixed by #183 (that repro was on v0.2.3, which predates it). This closes the global-position gap.

## How

`cli::dispatch`'s pre-verb scan now recognizes `--reporter`/`--loglevel` (captured, not leaked to Node) and records them — plus `--silent`/`-s` — as process defaults (atomics in `pm_engine::output`), mirroring the existing `SILENT` atomic. `OutputFlags` merges the per-verb flag over the default (per-verb wins). Invalid values give a clean usage error. `run` keeps its own per-run `--reporter`; the PM defaults are inert there.

## Maintainer-owned bit

run and PM have different `--reporter` vocabularies (`default/silent/ndjson` vs `default/append-only/silent`), so this deliberately does **not** add a single unified global `--reporter` ValueEnum. Unifying them (pnpm's full set is `default/append-only/ndjson/silent`) is a larger follow-up you own.

## Verified

Dev binary, byte-measured stderr: per-verb unchanged; `nub --silent install` / `-s install` / `--reporter=silent install` / `--loglevel=silent install` / `--silent add …` → empty stderr, dep linked, exit 0; `pnpm install --silent` parity = 0 bytes; global silent + install error still prints the error; append-only drops the footer keeps the dep list; `nub run` unaffected. Gates: clippy `--all-targets --all-features -D warnings` + fmt clean; output/cli/install_engine/grammar-parity tests pass. Non-network regression test `pre_verb_output_flags_reach_install_not_node` guards the misroute.

Refs #179

### DIFF
--- a/crates/nub-cli/src/cli.rs
+++ b/crates/nub-cli/src/cli.rs
@@ -1140,6 +1140,14 @@ fn run_nub() -> Result<i32> {
     let mut help_verbose = false;
     let mut show_warnings = false;
     let mut silent = false;
+    // Pre-verb PM output flags (`nub --reporter=silent install`,
+    // `nub --loglevel=error add foo`): captured here and recorded as process
+    // defaults below so the per-verb `OutputFlags` resolution can fall back to
+    // them (per-verb always wins). Without these arms a leading `--reporter`/
+    // `--loglevel` falls through to the file-run path and is shipped to Node
+    // (`node: bad option`). `--silent`/`-s` is already captured above.
+    let mut reporter_val: Option<String> = None;
+    let mut loglevel_val: Option<String> = None;
     // Top-level `--node`: provision the project's Node (version management stays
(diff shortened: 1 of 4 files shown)


colinhacks:

PR #192, opened 2026-06-26
fix(spawn): use `as _` for the TIOCSCTTY ioctl request cast (musl compile)

colinhacks:
Two PTY test workers in `spawn.rs` cast `libc::TIOCSCTTY as libc::c_ulong`. libc types the `ioctl` request as `c_int` on musl vs `c_ulong` on glibc, so it fails to compile for `x86_64-unknown-linux-musl` (`E0308: expected i32, found u64`). Surfaced by #182s dropped `embed-runtime-musl` leg.

Not release-blocking: both sites are inside `#[cfg(test)] mod tests`, excluded from release builds — only `cargo test`/`check --tests` for musl hit it.

Fix: `as _`, so inference picks the per-target type. Default-preserving on glibc.

Verified: musl Docker repro shows old cast errors, `as _` compiles on musl + glibc; local glibc clippy/fmt/test green.


colinhacks:

PR #193, opened 2026-06-26
fix(embed): degrade to un-augmented when self-heal leaves no live runtime dir

colinhacks:
`verify_or_heal`'s terminal canary branch returned `Some(target)` unconditionally, but a failed `swap_extract` (publish fails, stale-restore loses a race to `gc_stale`) leaves `target` absent. `find_preload`'s embed branch did no existence check, so a ghost dir hard-errored the child `node` instead of running un-augmented — violating the canary contract (always a live dir or `None`).

Fix: guard the terminal return with `target.is_dir()`; add a defense-in-depth `is_file()` gate in `find_preload`. Default-preserving. Test: `canary_never_returns_a_nonexistent_dir`.

Refs #182


colinhacks:

PR #190, opened 2026-06-26
Harden self-provisioning to aube-parity security (5 #172-class findings)

colinhacks:
Brings nub's PM/Node self-provisioning (`crates/nub-core/src/pm/*`, `version_management/*`) up to the guards the embedded aube engine already enforces. Five findings from `wiki/research/security-audit-172class.md` (#172 / F0a was fixed in #180); each fix mirrors an existing aube guard.

- **F0c** (`pm/registry.rs`): validate a registry `dist.version` (reject `/ \ NUL` ctrl `.` `..`) in `dist_from_meta` before it joins the executed `<store>/pm/<pm>/<version>/…` path. Mirrors `aube_store::validate_version`. Covers dist-tag + range; exact-pin was already semver-gated.
- **N1a** (`version_management/download.rs`): https→http downgrade-stop redirect policy (mirror of aube-registry), so the registry `_authToken` can't follow a redirect into cleartext.
- **N1b** (`pm/registry.rs`, `pm/provision.rs`, `cli.rs`): attach the registry credential to the tarball download only when `dist.tarball`'s origin matches the registry's. Packument fetch keeps full auth.
- **N2** (`pm/extract.rs`, `version_management/extract.rs`, `aube-runtime/extract.rs`): `CappedReader` on the gz/xz decoders + `take`-cap on zip entries (mirror of aube-store). Prod cap 1 GiB; bomb tests inject a tiny cap via `_capped` seams so real-tarball e2e is unaffected.
- **F0b** (`pm/extract.rs`): reject symlink/hardlink entries in the PM-provision extractor (CVE-2021-37701 class). pnpm/npm/yarn publishes ship none (verified live). The Node extractor still allows Node's intra-tree symlinks.
- **N3** (`vendor/aube/.../install/delta.rs`): contain the `exec:` change-hash `fs::read` through aube's own `resolve_exec_script_path`, so a committed lockfile's `exec:/abs` / `exec:../escape` / `exec:/dev/zero` can't drive an out-of-project or unbounded read. Default-preserving for standalone aube. `link:`/`file:`/`portal:` left unrestricted (legitimate monorepo semantics).

Verification: clippy `--all-targets --all-features -D warnings` + `fmt --check` clean; nub-core (306) + aube delta (33) + aube-runtime extract (7) + nub-cli PM integration green; each finding has a malicious-fixture test; real pnpm@10.0.0 provisioned through the rewritten extractor and run under Node.

[...]

PR #194, opened 2026-06-26
fix(pm): close the Windows drive-letter store-path escape in PM provisioning (F0c residual)

colinhacks:
Post-merge review of #190 found one residual in the F0c guard: `validate_version` didn't block `:`, so on Windows a drive-prefixed version (`C:foo`) makes `Path::join` discard the base and escape the store — a malicious-registry dist-tag could relocate the executed PM bin. Windows-only, drive-relative, registry-gated.

Fix: block `:` in nub-core's `validate_version` (always a concrete published semver; aube-store stays permissive for its git/alias version slots — no parity break) plus a `store_version_dir` containment backstop.

The other four lenses and the bot reviews were clean. Gates green: clippy --all-targets --all-features, fmt, nub-core lib 313, real pnpm/Node e2e.

Refs #172; follow-up to #190.

### DIFF
--- a/crates/nub-core/src/pm/provision.rs
+++ b/crates/nub-core/src/pm/provision.rs
@@ -31,6 +31,30 @@ pub struct ProvisionedPm {
     pub version: String,
 }
 
+/// Join a registry-resolved version onto the PM store dir, asserting the result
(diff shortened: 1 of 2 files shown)


colinhacks:

PR #195, opened 2026-06-26
fix(install): add --loglevel/--reporter to VALUE_FLAGS to prevent misroute to add

colinhacks:
`nub install --loglevel silent` (space form) misrouted to `add` and exited 1. Root cause: `install_to_add_args` uses a `VALUE_FLAGS` list to skip over a flag's value token; `--loglevel` and `--reporter` were absent, so `silent` was read as a package operand. Fix: add both to `VALUE_FLAGS`. The `Install` struct already accepts them via `#[command(flatten)] output: OutputFlags`.

Verified: space and equals forms both produce a silent exit-0 install; `--loglevel info` emits normal output; 198 unit tests pass; fmt and clippy clean.

### DIFF
--- a/crates/nub-cli/src/cli.rs
+++ b/crates/nub-cli/src/cli.rs
@@ -961,6 +961,13 @@ fn install_to_add_args(rest: &[String]) -> Option<Vec<String>> {
             "--dir",
             "--registry",
             "--node-linker",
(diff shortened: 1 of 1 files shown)


colinhacks:

PR #197, opened 2026-06-26
fix(pm): scope lockfile-ambiguity guard to the mutating install family

colinhacks:
`nubx cowsay hi` (and `nub dlx`/`nub create`) errored with `ERR_NUB_LOCKFILE_AMBIGUOUS` ("multiple lockfiles found: …") in a project that has multiple lockfiles. npm/pnpm/bun do not error on `npx`/`dlx`/`bunx` in this situation.

Root cause: the transient fetch-and-run verbs shared the same identity-resolution preflight (`engine_session` → `resolve_identity_walk_up` → `identity_error`) as the mutating install family that actually writes a lockfile. A dlx/nubx/create run never reads or writes the project lockfile, so lockfile ambiguity is irrelevant to it.

The fix adds an identity-strictness axis to the engine session:

- `engine_session_transient` (Lenient) — `dlx`, `nubx`, `create`: an ambiguity / declaration-contradiction degrades to no-identity (the same well-trodden fresh-project path) instead of hard-erroring. The transient run then proceeds (and fails downstream only for real reasons, e.g. a registry error).
- `engine_session` / `engine_session_quiet` (Strict) — `install`/`ci`/`add`/`remove`/`update`/`dedupe` and the project-reading families keep the hard error, since they operate on the project's lockfile and cannot proceed without a resolved identity.

`run`/`watch`/`nub <file>` never hit this path.

Regression test `transient_runs_do_not_error_on_multi_lockfile_projects` (`tests/pm_identity.rs`): a multi-lockfile fixture runs `nub dlx` and a `nubx` argv0 symlink without the ambiguity error (reaching the registry instead); the existing `undeclared_multi_lockfile_projects_error_as_ambiguous` continues to prove the mutating guard.

Verified locally: `cargo build`, `cargo clippy --all-targets --all-features -- -D warnings`, `cargo fmt --check`, `cargo test -p nub-cli --test pm_identity` (7/7). End-to-end on a 3-lockfile fixture: `nubx`/`dlx`/`run`/`nub <file>` succeed; `install`/`add`/`remove`/`update`/`dedupe`/`ci` still raise the ambiguity error.

### DIFF
--- a/crates/nub-cli/src/pm_engine/install_family.rs
+++ b/crates/nub-cli/src/pm_engine/install_family.rs
@@ -527,7 +527,10 @@ fn run_dlx(typed: &str, args: &[String]) -> Result<i32> {
         println!("{}", present::rewrite_help(help.to_string().trim_end()));
         return Ok(0);
     }
-    let session = super::engine_session(globals.dir.as_deref())?;
+    // Transient fetch-and-run (see `engine_session_transient`): a dlx never
+    // touches the CWD project's lockfile, so a multi-lockfile project must not
+    // hard-error on identity ambiguity — it degrades to no-identity here.
+    let session = super::engine_session_transient(globals.dir.as_deref())?;
     // NOTE: on child failure the engine propagates the child's exit code via
     // std::process::exit — control does not return here on that path. Output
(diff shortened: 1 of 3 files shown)


colinhacks:

PR #206, opened 2026-06-27
fix(install): suppress IWR progress bar on Windows

brequet:
`Invoke-WebRequest` re-renders a progress bar on every received chunk. In PowerShell this dominates the total download time.

Fix: set `$ProgressPreference = 'SilentlyContinue'` around the `Invoke-WebRequest` call, restore it after.

Applied to both `install.ps1` and `site/public/install.ps1`.

**Before / after on the same machine:**

pwsh-install-before-fix

pwsh-install-after-fix

| | |
|---|---|
| Before | 1m 39s |
| After | 6s |


colinhacks:

PR #212, opened 2026-06-27
feat(lockfile): read legacy npm lockfiles (lockfileVersion 1 + pre-2017 shrinkwrap)

colinhacks:
Reads `package-lock.json` lockfileVersion 1 (npm 5/6) and pre-2017 `npm-shrinkwrap.json` (nested `dependencies`, no `lockfileVersion`) — both previously hard-failed the install. The legacy nested tree is lifted into the existing flat v2/v3 `packages` representation, then reuses the unchanged reader pipeline (nested resolution, hoist, importers).

Default-preserving: the legacy branch fires only for lockfileVersion < 2 / absent, which standalone aube rejected, so the v2/v3 path is byte-for-byte unchanged — a clean additive upstream candidate for jdx/aube.

- The `lockfile_version` field becomes `Option` (old shrinkwraps omit it); direct deps come from package.json (yarn-classic model), so `npm::parse` now takes the manifest, threaded at the reader and the pm-use conversion path.
- Edges come from `requires` plus nesting (pre-npm-5 shrinkwraps omit `requires`).
- No-integrity / sha1 entries (npm 3 / npm 5.0) reuse the unchanged verify path: a missing pin warns and recovers a fresh sha512, or refuses under `strict-store-integrity`. The v2/v3 trust posture is untouched (reader-only diff).
- A fully-hoisted shrinkwrap that omits a dependency edge entirely (the transitive is unreachable and would silently drop) now warns (`WARN_NUB_LOCKFILE_LEGACY_INCOMPLETE_GRAPH`) instead of shipping a broken tree.

Verified with real `nub install`: npm-6 v1 (2-pkg, a 13-pkg scoped/nested/dev fixture, and lodash/lodash's 673-entry v1 lockfile), npm-3 oldshrink (warns + installs), `--frozen-lockfile`, and the strict-store-integrity refusal. Gates green (clippy/fmt/464 aube-lockfile tests). Five fresh-context self-review passes (correctness, impact, format-fidelity, security, follow-up) all clean.

Known follow-up: v1 git/`file:` deps encode the source marker in `version` (not `resolved`), so they aren't routed yet and fail loudly; registry deps — the dominant v1 population — work fully.

### DIFF
--- a/crates/nub-cli/src/pm_engine/use_align.rs
+++ b/crates/nub-cli/src/pm_engine/use_align.rs
@@ -287,7 +287,9 @@ fn parse_source_graph(
 ) -> Result<aube_lockfile::LockfileGraph> {
     match from_kind {
         LockfileKind::Pnpm | LockfileKind::Aube => aube_lockfile::pnpm::parse(from),
-        LockfileKind::Npm | LockfileKind::NpmShrinkwrap => aube_lockfile::npm::parse(from),
+        LockfileKind::Npm | LockfileKind::NpmShrinkwrap => {
+            aube_lockfile::npm::parse(from, manifest)
+        }
         LockfileKind::Yarn | LockfileKind::YarnBerry => aube_lockfile::yarn::parse(from, manifest),
         LockfileKind::Bun => aube_lockfile::bun::parse(from),
     }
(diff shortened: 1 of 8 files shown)


colinhacks:
Shipped in v0.2.6: https://github.com/nubjs/nub/releases/tag/v0.2.6

PR #213, opened 2026-06-27
feat(dist): winget manifest + release-time winget-pkgs submission

colinhacks:
Makes nub installable via `winget install` on Windows, with a high-confidence test path that does not require publishing to `microsoft/winget-pkgs` first.

## What's here

- **Manifest** — `winget/manifests/n/Nubjs/Nub/0.2.5/` (version, installer, en-US locale; schema 1.6.0). Installer type `zip` + `NestedInstallerType: portable`, one `Installer` per arch (x64, arm64) with the real release-asset URLs and SHA256s (read from the published `.sha256` sidecars). Exposes both `nub` and `nubx` as portable command aliases. This committed set is both the first-submission payload and the fixture the test workflow installs from.
- **Test workflow** — `.github/workflows/winget-validate.yml` (windows-latest): `winget validate` then `winget install --manifest` the local manifest, then asserts `nub --version` and `nubx --version`. Path-filtered to `winget/**` + the workflow file, plus `workflow_dispatch`. No winget-pkgs publication needed.
- **Release submission** — a `submit-winget` job in `.github/workflows/release.yml` that submits new versions to winget-pkgs via `vedantmgoyal9/winget-releaser` (pinned by SHA). Gated exactly like the existing `bump-homebrew-tap` job: it is a **no-op when the `WINGET_PAT` secret is absent**, so merging this fires nothing externally.
- **`winget/README.md`** — the test paths, the confidence chain, and the one-time maintainer enablement.

## Why the test is high-confidence

`winget install --manifest` performs the same operation the winget-pkgs validation bot performs — download the zip, verify its SHA256, extract, register the portable binaries. So:

1. Local/CI `winget install --manifest` green means the manifest is installable.
2. Submission to `microsoft/winget-pkgs` opens a PR; their bot re-runs that same install-in-sandbox check and blocks merge on failure.
3. A broken manifest therefore cannot reach users — worst case is "the PR doesn't merge."

The highest-fidelity local check (Windows Sandbox via winget-pkgs's `Tools/SandboxTest.ps1`) is documented in `winget/README.md`; it mirrors the bot 1:1 but needs nested virtualization, so it is referenced rather than run in CI.

## Held for maintainer sign-off

No external submission fires from this PR. To enable real submission later (per `winget/README.md`):

1. Bootstrap the first `Nubjs.Nub` version manually from the committed manifest (winget-releaser only updates an existing package).
2. Create a PAT (classic `public_repo`, or fine-grained contents+pull-requests:write on the winget-pkgs fork) under the account that forks winget-pkgs.
3. Add it as the `WINGET_PAT` repo secret. After that, each tagged release auto-opens a winget-pkgs PR.

## Notes

- Package id casing (`Nubjs.Nub`), moniker (`nub`), and publisher (`nub contributors`) follow winget convention and can be adjusted before the first submission.
- No originating issue exists, so no closing keyword.
- User-facing "winget install nub" docs are intentionally deferred until the package is live in winget-pkgs, to avoid documenting an install path that doesn't yet resolve.
- Windows is nub's least-exercised platform; winget exposure invites more Windows installs. Flagging, not gating.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1064,3 +1064,51 @@ jobs:
           git add Formula/nub.rb
           git commit -m "nub: update formula to v$VERSION"
           git push
+
+  submit-winget:
+    name: Submit winget manifest
+    # Open/refresh the Nubjs.Nub manifest in microsoft/winget-pkgs for this release.
+    # Runs ONLY on a real tag-push release (never the debug workflow_dispatch path)
+    # and AFTER github-release, because the manifest references the published
+    # release-asset URLs + their SHA256 sidecars this run created.
+    #
+    # HELD FOR MAINTAINER ENABLEMENT — this is an EXTERNAL, effectively-irreversible
+    # publish under the nubjs publisher identity (winget-releaser opens a PR on
+    # microsoft/winget-pkgs from the token owner's fork). It is a NO-OP until a
+    # WINGET_PAT secret is present: the gate step below reads the secret and skips
+    # the submission when it is absent (empty string), so merging this workflow
+    # fires NOTHING. winget-releaser also UPDATES an existing package — the FIRST
+    # version (Nubjs.Nub v1) is bootstrapped manually from the committed
+    # winget/manifests/... set. Both are documented in winget/README.md.
+    if: github.event_name == 'push'
+    needs: github-release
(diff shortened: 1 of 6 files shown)


colinhacks:

PR #214, opened 2026-06-27
fix(watch): re-read .env across restarts (#207)

### LINKED ISSUE #207: nub watch freezes .env values across restarts — changed vars never picked up
# `nub watch` freezes `.env` values across restarts — changed vars never picked up

## Summary

Under `nub watch`, a `.env` change triggers a restart but the **new value is not applied** — the process keeps the value `.env` had when `nub watch` first started. Plain `node --watch --env-file=.env` (the same flags nub passes to node) does live-update across the restart, so the regression is in how nub provisions the environment, not in Node.

The cause is that nub loads `.env` into the **spawned process environment** at `nub watch` startup *and* passes `--env-file` to node. Node's documented rule is that `--env-file` never overrides a variable already present in the environment, so the value nub baked in at startup always wins. `node --watch` restarts only its child (which inherits nub's stale environment), and the long-lived `nub watch` process — the one that actually read `.env` — never restarts. The value is therefore frozen until `nub watch` itself is restarted.

## Environment

- nub: `0.2.3` (installed via mise `github-nubjs-nub`)
- node: `v22.23.1`
- OS: macOS 26.5.1 (arm64)

## Minimal reproduction

```sh
mkdir repro && cd repro
echo '{ "name": "repro", "version": "1.0.0" }' > package.json
printf 'TEST=value_ONE\n' > .env
cat > app.mjs <<'EOF'
console.log("[BOOT] pid=" + process.pid + " TEST=" + process.env.TEST);
setInterval(() => {}, 1000);
EOF

nub watch app.mjs
# -> [BOOT] pid=... TEST=value_ONE

# in another shell, change the value:
printf 'TEST=value_TWO\n' > .env
# nub logs: Restarting 'app.mjs'
# -> [BOOT] pid=... TEST=value_ONE   <-- still ONE (BUG)
```

## Expected

After the restart, `TEST=value_TWO` — matching plain Node:

```sh
node --watch --env-file=.env app.mjs
# change .env to value_TWO -> Restarting -> [BOOT] TEST=value_TWO   (correct)
```

## Actual

`TEST` stays `value_ONE` across every watch-restart. Only stopping and re-running `nub watch` picks up the new value.

## Root cause detail

1. nub auto-detects `.env` and launches node with `--env-file=<abs>/.env` (confirmed in the process args).
2. nub also seeds `.env`'s values into the child's environment at `nub watch` startup.
3. Node: a variable already in `process.env` takes precedence over `--env-file`, so the seeded value is never overridden. Verified:
   ```sh
   printf 'TEST=from_file\n' > .env
   node --env-file=.env -e 'console.log(process.env.TEST)'              # from_file
   TEST=from_shell node --env-file=.env -e 'console.log(process.env.TEST)'  # from_shell
   ```
4. `node --watch` restarts only the child; it inherits the seeded (stale) environment. `nub watch`, which read `.env`, does not restart.

## Suggested fix

Either:
- don't seed `.env` into the spawned environment when `--env-file` is also passed (let Node's `--env-file` own it, so it re-applies on each watch-restart), or
- re-read `.env` and re-provision the child environment on every watch-restart.

colinhacks:
`nub watch` froze `.env` values across restarts; `node --watch --env-file` live-updates, so this was a nub regression.

Cause: `run_watch` passes `.env*` via `--env-file` (re-read on restart) AND injects pre-expanded values via `Command::env`, which Node's `--env-file` can't override — the startup value froze.

Fix: inject only vars whose `${VAR}` expansion changed the raw value; plain vars flow through `--env-file` and live-reload. Expansion-dependent vars stay injected (documented trade-off); explicit `--env-file` unchanged. Verified differentially; unit tests added.

Closes #207

### DIFF
--- a/crates/nub-cli/src/cli.rs
+++ b/crates/nub-cli/src/cli.rs
@@ -4179,24 +4179,15 @@ fn run_watch(file: &str, args: &[String]) -> Result<i32> {
             .map(|p| nub_core::workspace::env::discover_env_files(&p.root))
             .unwrap_or_default()
     };
-    // Pre-expand env vars (same path as the direct file runner). Auto `.env*` is
-    // suppressed when `--env-file` is present; `merge_child_env` applies the gate
-    // and overlays the explicit `--env-file` vars (shell still wins).
-    let auto_env = project
-        .as_ref()
-        .map(|p| nub_core::workspace::env::load_env_files(&p.root))
-        .unwrap_or_default();
-    let env_vars = merge_child_env(
-        auto_env,
-        env_file_present,
-        ENV_FILE_VARS.get().unwrap_or(&HashMap::new()),
-    );
-    // The RAW (unexpanded) values Node's `--env-file` args deliver, so the inject
-    // loop below can skip any var whose value expansion didn't change — leaving it
-    // to Node's `--env-file` to live-reload across restarts (#207). Empty when the
-    // user passed `--env-file`: auto-discovery is suppressed and Node gets no
-    // `--env-file` args, so injection is the only delivery channel — inject all.
+    // Load the `.env*` files ONCE: `raw_env` is the unexpanded merge (the values
+    // Node's `--env-file` args deliver), `auto_env` is the same map expanded. A
+    // single read (rather than `load_env_files` + a second `load_env_files_raw`)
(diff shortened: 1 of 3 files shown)


colinhacks:

PR #215, opened 2026-06-27
test(pm): isolate vocab-precedence cases against the manifest cache

colinhacks:
Fixes a Windows flake in `vocab_precedence_keys_on_the_declared_pnpm_major`.

The test rewrites one `package.json` path three times, reading it back through `ROOT_MANIFEST_CACHE` (keyed on `(mtime, size)`). The `pnpm@11.9.0`→`pnpm@9.15.9` rewrite is byte-identical in length, so `size` cannot disambiguate it; when both writes share an mtime tick (Windows resolution is coarse), the third read serves the stale parse of the second → major 11 → `YamlWins`, failing the pnpm-9 case.

Each case now gets its own temp dir, keying the cache on a distinct path. Production `pnpm_vocab_precedence` untouched.

Refs #214 (its Windows leg hit this flake).

### DIFF
--- a/crates/nub-cli/src/pm_engine/use_nub.rs
+++ b/crates/nub-cli/src/pm_engine/use_nub.rs
@@ -1433,10 +1433,18 @@ mod tests {
 
     #[test]
     fn vocab_precedence_keys_on_the_declared_pnpm_major() {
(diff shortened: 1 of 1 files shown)


colinhacks:

PR #217, opened 2026-06-27
pm: abort install on unresolvable yarn.lock sources and Yarn PnP

colinhacks:
## What

`nub install` / `nub ci` now refuse, before any `node_modules` write, a Yarn lockfile they can't honor faithfully — rather than produce a tree that diverges from what Yarn installs.

**Unresolvable lockfile source.** A `yarn.lock` (Classic or Berry) entry whose dependency source the reader can't resolve — a `git`/`jsr`/unknown protocol — was reclassified to a registry `name@version` (a 404 at fetch, past the abort window) or, for Berry, dropped. A non-optional such entry now aborts at parse time, naming the entry and protocol; an optional one warns and is skipped, matching Yarn's tolerance of a missing optional.

```console
$ nub install
ERR_NUB_LOCKFILE_UNSUPPORTED_SOURCE
  × failed to parse lockfile
  ╰─▶ lockfile yarn.lock entry `leftpad@git+https://github.com/user/leftpad.git` uses the unsupported `git+https` source
  help: nub can't resolve this source from the lockfile — install with the package manager that wrote it, or remove the dependency
```

**Yarn Plug'n'Play.** A PnP project (`nodeLinker: pnp`, or Yarn Berry's default) was silently downgraded to a `node_modules` install — a different on-disk layout than Yarn's. It now aborts with `ERR_NUB_PNP_UNSUPPORTED`.

## Fork discipline

The strict behavior is gated behind a new compile-time embedder toggle `strict_unsupported_source` — `false` in the standalone `aube` profile, `true` in nub's — so standalone aube keeps its prior reclassify/drop default byte-for-byte. The optional-skip reuses the existing `skipped_optional_dependencies` graph field, so a frozen (`ci`) drift check tolerates the deliberately-absent optional.

## Tests + docs

- `vendor/aube/crates/aube-lockfile/tests/unsupported_source.rs` — strict-profile reader behavior (Berry jsr fatal, Berry git still resolves, Classic git/github-shorthand fatal, optional warn+skip, clean lockfile no false-positive).
- `vendor/aube/crates/aube-lockfile/src/yarn/tests.rs` — toggle-off anti-regression, the `ClassicSource` classifier, a property test that a registry-shaped range never classifies unsupported.
- `crates/nub-cli/tests/abort_eagerly.rs` — end-to-end: git-dep and PnP fatal (non-zero exit, branded code, no `node_modules`), and the optional carve-out.
- `site/content/docs/install/yarn.mdx` — PnP-abort and git-source rows.

Full aube workspace suite (2429 tests) and the nub PM suite are green; `clippy --all-targets --all-features` and `fmt --check` clean.

https://claude.ai/code/session_01TuvdYzBgmnyJQJDXheWmF9

### DIFF
--- a/crates/nub-cli/src/pm_engine/identity.rs
+++ b/crates/nub-cli/src/pm_engine/identity.rs
@@ -158,6 +158,11 @@ pub(crate) const NUB: aube_util::Embedder = aube_util::Embedder {
     // The animated single-line install bar is nub's first-class TTY UX; the
     // engine still falls back to append-only for CI / piped / non-TTY output.
     tty_progress: true,
+    // nub's guarantee is "the installed tree equals the incumbent PM's, or an
+    // eager precise refusal" — so a lockfile source nub can't resolve (a
+    // git/jsr/unknown protocol in a yarn.lock or bun.lock) aborts at plan time
+    // for a non-optional dep, instead of reclassify→404 / silent drop.
+    strict_unsupported_source: true,
 };
 
 /// Register [`NUB`] as the active embedder profile. Idempotent (the engine's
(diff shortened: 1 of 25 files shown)


colinhacks:

PR #218, opened 2026-06-27
test: harden four shared-state flake hazards in the nub-cli suite

colinhacks:
Test-only hermeticity fixes for four shared-state flake hazards in `crates/nub-cli`. No production or behavior change.

- **pm_shim watchdog** (`empty_path_entry_with_cwd_at_the_shim_does_not_loop`): raise the exec-loop watchdog 10s->60s. The reproduced flake — it starves only under load when sibling pm_shim tests run concurrently, a load-correlated false-fail of a finite run, not the infinite exec-loop the watchdog guards. A real loop is unbounded, so 60s still trips it.
- **unsupported_config HOME race**: serialize the two tests that mutate the process-global `$HOME` under a poison-recovering `HOME_LOCK`, held across the whole set->use->restore window. The lib/bin test binary runs multi-threaded, so an unguarded `set_var("HOME")` is a latent hazard to any concurrent `home_dir()` reader; the prior `// SAFETY: single-threaded test` note was false. Hardening, not an observed failure.
- **workspace_parallel_timing /tmp collision**: qualify the stamp dir by PID in addition to subsec_nanos and pre-clean it, so two CI matrix jobs sharing `/tmp` can't collide on the same nanosecond and a stale dir can't poison a stamp read.
- **polyfill tests** (`polyfills_available`, `base64_and_resource_management_polyfills`): write the temp script into a per-test unique tmpdir instead of the checked-in `vanilla-ts` fixture, and isolate `XDG_CACHE_HOME`, so a panic can't strand a file in the repo and concurrent runs don't share the real `~/.cache/nub`.

Verified hermetic under repeated high-thread runs: `HOME_LOCK` 40/40 and the three integration tests 20/20 at `--test-threads 1`; the whole `pm_shim` binary 30/30 at `--test-threads=16`. A fresh-context self-review pass (correctness / order-independence / test-only / impact-analysis) approved the diff.

Refs #215

### DIFF
--- a/crates/nub-cli/src/pm_engine/unsupported_config.rs
+++ b/crates/nub-cli/src/pm_engine/unsupported_config.rs
@@ -658,6 +658,14 @@ mod tests {
         tempfile::tempdir().unwrap()
     }
 
+    /// Serializes the two tests that mutate the process-global `$HOME`. This lib
+    /// test binary runs MULTI-THREADED (nothing pins `--test-threads=1`), so the
+    /// set→use→restore window must be held under a lock or it races sibling tests
+    /// (and each other) — every `scan_unsupported_config` call reads
+    /// `dirs_next::home_dir()`. Poison-recovering, matching the crate's other
+    /// process-global seams (`RELEASE_ENV_LOCK`, `CWD_LOCK`).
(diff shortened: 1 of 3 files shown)


colinhacks:

PR #220, opened 2026-06-28
fix(aube): warm installs reuse the store for no-integrity (legacy v1 lockfile) packages

colinhacks:
## Problem

A project with a legacy `lockfileVersion: 1` lockfile (or a pre-2017 `npm-shrinkwrap.json`) re-downloads every integrity-less package on every install, even when the global store is warm. On a 519-package fixture (CoffeeScript 2.0.1), 104 packages re-fetched on each warm install, and a warm install failed outright offline.

## Cause

The store index is keyed by integrity: `Some` → `index/<hex>/<name>@<ver>.json`, `None` → `index/<name>@<ver>.json` (root). The stream-eligible tarball import wrote the index under the *computed* sha512 (the hex key), but the warm-install classifier reads with the *lockfile* integrity — `None` for legacy entries — which maps to the root key. The two keys never met, so the warm read always missed and re-fetched. The buffered import path already wrote the root key; only the stream-eligible path was wrong.

## Fix

Factor the dual-save into `persist_pkg_index`, shared by the buffered and stream-eligible import paths. For a no-integrity entry, also write the index under the root/`None` key (the warm-read key). Integrity-bearing packages are unchanged, and the computed-hex write is kept for the self-heal case (a non-frozen install rewriting the lockfile to v3 with the computed integrity).

Before / after on the 519-package fixture, warm (`node_modules` purged, store primed):

| | re-downloaded | warm offline install |
| --- | --- | --- |
| before | 104 | fails |
| after | 0 | succeeds |

## Tests

A hermetic unit test (`no_integrity_import_is_retrievable_by_warm_none_key`) drives `persist_pkg_index` over a tempdir store: a no-integrity import is retrievable by the `None` warm-read key, while an integrity-bearing import is not dual-saved (no cross-source aliasing). It fails if the dual-save is reverted or the `is_none()` guard is dropped.

## Notes

- vendor/aube change landing as a normal nub PR. It harmonizes the stream-eligible path with the buffered path's existing behavior — an upstream candidate for jdx/aube.
- `cargo clippy -p aube --all-targets` fails from the nub workspace root on an unresolved `predicates` dev-dep in aube's `tests/e2e.rs`. This is pre-existing on `main` (identical failure on a pristine checkout) and unrelated to this change; it passes from the aube workspace. `vendor/aube/crates/aube/src/progress/mod.rs` also carries a pre-existing fmt diff, left untouched.

Refs #212

### DIFF
--- a/vendor/aube/crates/aube/src/commands/install/lifecycle.rs
+++ b/vendor/aube/crates/aube/src/commands/install/lifecycle.rs
@@ -746,6 +746,55 @@ pub(crate) async fn run_dep_lifecycle_scripts(
     Ok(ran)
 }
 
+/// Persist a freshly imported package index under the store key(s) a
+/// later warm install will read it back by. Shared between the buffered
+/// and streaming import paths so both key the cache identically.
+///
+/// `lockfile_integrity` is the SRI the resolver carried from the
+/// lockfile (`None` for a v1/legacy entry or an integrity-stripping
+/// proxy); `computed_integrity` is the sha512 the streaming path
+/// derived from the tarball bytes when the lockfile carried none. The
+/// index is always saved under the effective key
+/// (`lockfile_integrity.or(computed_integrity)`).
+///
+/// The legacy-cache fix: when there is no lockfile integrity but the
+/// effective key is a *computed* sha512, the index lands in a hex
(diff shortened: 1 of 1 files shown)


colinhacks:

PR #221, opened 2026-06-28
fix(aube): content-address the no-integrity warm lookup per project

colinhacks:
A no-integrity package's store index was also written under a content-free `index/<name>@<version>.json` root key. In the per-user shared store, that key lets one project's bytes serve another's warm install of the same coordinate, across registries, unverified.

Now content-addressed: `persist_pkg_index` drops the root-key write (hex write kept); a per-project `<state-dir>/no-integrity-index.json` binds `<registry-name>@<version>` to the sha512 this project computed. Classifier and linker fallback resolve through it; no binding → re-fetch.

Integrity-bearing locks and warm perf unchanged. Changes standalone aube on-disk behavior.

Refs #212. Hardens #220.


colinhacks:

PR #222, opened 2026-06-28
perf(linker): whole-dir clonefile + parallel fill for the hoisted layout

colinhacks:
The hoisted layout (npm/yarn/bun-lockfile projects) filled node_modules with a single-threaded per-file copy loop, skipping the whole-dir `clonefile(2)` fast path isolated mode uses.

This parallelizes the fill across a rayon pool (one depth level at a time, preserving bundled-`node_modules` ordering) and clones each placement via the isolated linker's `try_clonedir_fill` — each nested duplicate becomes a CoW clone of one shared `trees/` entry. Non-macOS/non-APFS/cross-volume fall back to the per-file path.

webpack@3 fixture (76,036 files, warm, `--offline`): `phase:link` 11.3s -> 1.5s, byte-identical output. 103 linker tests pass. jdx/aube upstream candidate.

### DIFF
--- a/vendor/aube/crates/aube-linker/src/hoisted.rs
+++ b/vendor/aube/crates/aube-linker/src/hoisted.rs
@@ -41,8 +41,9 @@
 //! dependency lifecycle scripts can locate a package's on-disk
 //! directory without recomputing the tree.
 
+use crate::pool::with_link_pool;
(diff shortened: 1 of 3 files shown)


colinhacks:

PR #223, opened 2026-06-28
fix(upgrade): rescue sidecar-era nub upgrade onto bin-only releases

colinhacks:
`nub upgrade` from sidecar-era v0.1.x fails on bin-only v0.2.x releases: `swapped bin/ but failed to swap runtime/`. #175 dropped the `runtime/` sidecar, but the v0.1.0–v0.1.14 upgrader still blindly renames the archive's `runtime/` into place, aborting and half-updating the install.

Fix: ship a vestigial empty `runtime/` in each release archive so the old rename completes; the embed-runtime binary ignores `~/.nub/runtime`. Also: `nubx` alias now best-effort; Homebrew installs by `bin/` path (two top-level entries now).

Verified: repro + fix via a faithful v0.1.10 swap replica; new test covers both archive shapes; clippy/fmt clean; impact-analysis cleared all consumers. No GH issue.

### DIFF
--- a/.github/scripts/gen-homebrew-formula.sh
+++ b/.github/scripts/gen-homebrew-formula.sh
@@ -80,12 +80,14 @@ class Nub < Formula
   def install
     # nub is a single self-contained binary: it embeds its runtime (preload +
     # vendored polyfills + native addon) and JIT-extracts it to ~/.cache/nub on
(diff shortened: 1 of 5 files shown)


colinhacks:

PR #224, opened 2026-06-28
feat(nubx): unified runner with a consent-gated registry tier

colinhacks:
Make `nubx` a unified runner and split it from `nub dlx`. `nubx <thing>` now resolves a local **file → script → bin** before reaching the registry, and that implicit registry tier is gated so `nubx` never executes remote code silently.

## What changed

- **`nubx` resolves locally first.** File (a path, or a verbatim-existing file) → `package.json` script → `node_modules/.bin` bin → registry. The order matches `run`/`mux`; scripts beat bins, a file beats both. A local match never touches the network or prompts. (`-p`/`--package` still forces the registry, as on npx.)
- **The registry tier is consent-gated.** On a full local miss `nubx` reaches the registry, but: **CI → fail closed; non-TTY → fail closed; interactive TTY → prompt `y/N` on the first fetch of a spec, then run without re-prompting once consent is recorded.** `-y`/`--yes` is the explicit escape hatch (lets CI / non-TTY through, skips the prompt).
- **`nub dlx` / `nub x` split off — the explicit, ungated downloader.** They are no longer aliases of `nubx`. Invoking `nub dlx` *is* the consent: no prompt, runs in CI. (Unchanged engine verbs.) `nub exec` stays the no-network local-bin runner.
- **Consent is persisted** in a per-user ledger (`<cache>/dlx/consent.json`, atomic write) keyed on the canonical input spec. Pinned (exact `1.2.3`) is remembered forever; floating (`latest`, a tag, a range) honors a 24h TTL, then re-prompts (a moved tag = new code = fresh consent).
- **Docs rewritten** (`site/content/docs/nubx.mdx`): resolution order, the fetch gate, and a "Coming from npx" migration section.

Implementation is **nub-cli only — no `vendor/aube` change.** New module `crates/nub-cli/src/nubx_consent.rs`; the gate wraps the existing `run_dlx_for_nubx` call in `run_exec_with_dlx`.

## Decisions I made (for your review)

All reversible; recorded in `.fray/universal-nubx.md` under Decisions.

1. **`-p` forces the registry tier** (skips local resolution; npx-identical). The gate keys on the canonical install set (the `-p` packages, else the bare bin), the same identity the engine resolves.
2. **`nubx` takes one bare token, not `mux`'s multi-token chain.** No `$ ` shell prefix, no verb-leading — those are `mux`'s surface. `nubx` shares the file>script>bin *order*, not the token grammar.
3. **`run`/`exec` do NOT gain file-resolution.** It stays scoped to `nubx` + the bare file-runner (+ `mux` when built). `run` stays byte-for-byte pnpm-compatible; `exec` stays bins-only. "Universal" = the order is consistent where file-resolution applies.
4. **npx migration documented:** a local script now wins over a download (npx is bins-only), and CI/non-TTY now fails closed where npx downloads. Both called out with the concrete break cases.
5. **Scope: consent LEDGER now, byte-level TREE cache deferred.** v1 ships the full security thesis (fail-closed) + consent UX (prompt-once-then-no-prompt) via the ledger, but NOT the per-resolution tree cache that would let a consented hit *skip the re-fetch*. The ledger is pure nub-cli (zero engine risk, small diff) and delivers every observable consent behavior; the tree cache is a perf optimization needing a default-preserving aube `dlx` extension (it's tempdir-only today) and keys off the SAME ledger entries, so it layers on without changing the contract. Not a regression — today's `nubx` already re-fetches into a fresh tempdir every run; this keeps that cost and adds the gate.

## Self-review

Two fresh adversarial reviewers (impact-analysis + correctness/security) ran on the diff. The security pass found two real holes, both fixed in the second commit:

- **Consent was recorded on a failed fetch.** The engine maps a 404 / install failure to `Ok(nonzero)`, so the old `result.is_ok()` guard recorded consent for a package that never resolved — a one-time `y` on a not-yet-published spec would become a permanent silent run-grant. `run_dlx_for_nubx` now returns `(exit_code, fetched_ok)`; the ledger write gates on `fetched_ok`.
- **The ledger was checked before the CI / non-TTY gate.** A restored/shared `consent.json` could let a fetch run silently in CI, contradicting "CI → always fail closed." The gate now fails closed in CI / non-TTY unconditionally; the ledger is consulted only in an interactive terminal. (`-y` still proceeds anywhere.)

A regression test asserts a failed fetch leaves the ledger untouched.

## Testing

`cargo clippy --all-targets --all-features -- -D warnings` + `cargo fmt --check` clean; scoped `cargo test -p nub-cli` green. 4 unit tests (spec classification, ledger key, TTL/pinned roundtrip) + 7 e2e tests (CI fail-closed, non-TTY fail-closed, `-y` bypass, failed-fetch-records-nothing, the dlx/nubx split, the script tier, file>script precedence). e2e tests isolate `XDG_CACHE_HOME` so the consent ledger is hermetic.

Note: `exec_runs_node_and_non_node_bins` fails locally on a bare `cargo test` debug binary (the NODE_OPTIONS preload isn't staged for it) — confirmed pre-existing on `origin/main`, unrelated to this change; it passes under CI's embedded-runtime build.

Refs the `universal-nubx` design thread (feature work, no tracking issue).


colinhacks:

PR #234, opened 2026-06-28
perf(install): short-circuit a warm-satisfied install under no-downgrade

### LINKED ISSUE #2: Bash Installation failure on MacOS - api_auth[@]: unbound variable
```bash
version=$(curl -fsSL "${api_auth[@]}" "https://api.github.com/repos/nubjs/nub/releases/latest" | grep '"tag_name"' | sed -E 's/.*"v(.*)".*/\1/')
```

The issue is `"${api_auth[@]}"` when `api_auth` is an empty array, `set -euo pipefail` causes `u` (nounset) to treat an unset/empty array expansion as an error in some bash versions (particularly bash 3.x, which is the default on macOS).

We can fix the issue if we only expand the array if it's non-empty, i.e. `${api_auth[@]+"${api_auth[@]}"}`

```bash
version=$(curl -fsSL ${api_auth[@]+"${api_auth[@]}"} "https://api.github.com/repos/nubjs/nub/releases/latest" | grep '"tag_name"' | sed -E 's/.*"v(.*)".*/\1/')
```

I'm opening a PR to fix

colinhacks:
A repeat `nub install` on an unchanged, fully-satisfied tree re-ran the whole resolve/fetch/link pipeline instead of exiting "Already up to date". The warm short-circuit was gated off before it could run: under the default `trustPolicy=no-downgrade`, an online install forced trust re-validation in `install_fast_path_eligible` and bailed before `check_needs_install` was ever consulted. npm, pnpm, and bun all exit instantly on a satisfied tree, as do nub's own `--offline` path and `nub run` auto-install — only the explicit online `install` re-validated.

## Change

A fully-satisfied no-op now short-circuits regardless of trust policy; the trust gate fires only when there is real work.

- New embedder toggle `warm_trust_revalidate` on `aube_util::Embedder`. Standalone aube sets it `true` (re-validates online exactly as before — byte-for-byte unchanged); nub's profile sets it `false`.
- In `install_fast_path_eligible`, the pre-`check_needs_install` trust bail is now gated on the toggle. The short-circuit is reachable only when `check_needs_install` returns `None` (lockfile, manifest, settings, and layout all match the on-disk tree — zero resolve, fetch, or link). Any install that does real work returns `Some(reason)`, falls through to the full pipeline, and trips the trust check during resolution.

## Security

The trust gate still fires on every install that installs anything. The short-circuit only skips redundant re-validation of an already-on-disk, already-validated tree, and a no-downgrade-violating package can never be in a satisfied tree because installing it aborts first.

Verified end-to-end:

- Warm-satisfied online install under the default `no-downgrade` exits "Already up to date" with zero network and no re-link (wall ~0.02s, down from a full re-pipeline).
- A fresh install of `node-gyp@10.3.0` (whose picked version dropped the provenance attestation `10.3.1` carried) still aborts with exit 23 / `ERR_NUB_TRUST_DOWNGRADE`, linking nothing.

Both directions are covered by committed regression tests in `crates/nub-cli/tests/install_engine.rs`.

Refs: install-warm-fastpath-trust-gate (perf fix #2 of the warm-install workstream).

### DIFF
--- a/crates/nub-cli/src/pm_engine/identity.rs
+++ b/crates/nub-cli/src/pm_engine/identity.rs
@@ -163,6 +163,14 @@ pub(crate) const NUB: aube_util::Embedder = aube_util::Embedder {
     // git/jsr/unknown protocol in a yarn.lock or bun.lock) aborts at plan time
     // for a non-optional dep, instead of reclassify→404 / silent drop.
     strict_unsupported_source: true,
+    // A fully-satisfied warm install (node_modules present, lockfile/manifest/
+    // settings/layout all match) short-circuits to an instant "Already up to
+    // date" regardless of `trustPolicy`, matching npm/pnpm/bun and aube's own
+    // offline + `aube run` auto-install paths. The trust gate is a resolve-time
+    // downgrade defense with nothing to validate on a no-op (zero resolve/fetch/
+    // link); any install that does REAL work misses the short-circuit and still
+    // trips the gate during resolution. Standalone aube keeps the re-validation.
+    warm_trust_revalidate: false,
 };
 
 /// Register [`NUB`] as the active embedder profile. Idempotent (the engine's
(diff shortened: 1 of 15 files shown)


colinhacks:

PR #235, opened 2026-06-28
perf(install): overlap the default isolated materialize with fetch

colinhacks:
## What

The fetch-pipelined materializer (`spawn_gvs_prewarm`) hides the per-package materialize under the download tail — but until now only for the GVS sub-mode (`node-linker=isolated` + `hoist=false`). Under the default isolated layout (`hoist=true`) the prewarm got the raw, hoist-unaware global-virtual-store override, so it populated the shared global store while `link_all` — which recurses to per-project materialization when `use_global_virtual_store && hoist` — re-materialized every package into `.aube/<dep_path>` serially after fetch. The prewarm work was wasted; the materialize ran on the critical path.

This routes the prewarm by the link phase's effective GVS decision, so the per-project copy lands where `link_all` reads it and overlaps fetch. `gvs::prewarm_global_virtual_store_override` is the single home for that decision.

## Scope

Isolated linker only. The hoisted layout keeps the raw override — it discards `.aube` regardless, so forcing its prewarm per-project would only strand orphan `.aube/<dep_path>` dirs the hoisted sweep doesn't reclaim. GVS (`hoist=false`) is unchanged: `effective_gvs` is `true` off-CI, so the prewarm stays on the shared store.

## Behavior

The `node_modules` tree is byte-identical; only the timing of the materialize moves. Measured on a 440-package isolated cold install (`coffeescript@2.0.1` closure):

| mode | link phase before | link phase after |
| --- | --- | --- |
| default isolated (`hoist=true`) | ~0.5s (materializes every file) | tens of ms (0 files, overlaps fetch) |
| GVS (`hoist=false`) | ~30ms | ~30ms (unchanged) |
| hoisted `node-linker` | unchanged | unchanged |

Verified cold + warm (idempotent) installs at rc=0; the cold tree is byte-identical before/after (paths + symlink targets); the global virtual store is no longer populated under default isolated.

## Tests

`gvs::prewarm_global_virtual_store_override` carries unit tests for the isolated-mirrors-effective and hoisted-passes-raw cases.

### DIFF
--- a/vendor/aube/crates/aube/src/commands/install/gvs.rs
+++ b/vendor/aube/crates/aube/src/commands/install/gvs.rs
@@ -80,6 +80,34 @@ pub(super) fn effective_global_virtual_store(
     planned_gvs && !hoist && matches!(node_linker, aube_linker::NodeLinker::Isolated)
 }
 
+/// The global-virtual-store decision the fetch-pipelined materializer
+/// (`spawn_gvs_prewarm`) must use so its per-package materialize lands
+/// exactly where the link phase later reads it.
+///
+/// For the isolated linker this is the *effective* mode
+/// ([`effective_global_virtual_store`]), which folds in `hoist`: under the
+/// default `hoist=true` layout `Linker::link_all` recurses to
+/// `without_global_virtual_store` and materializes per-project, so the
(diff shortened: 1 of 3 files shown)


colinhacks:

PR #236, opened 2026-06-28
aube: make no-work warm install phases legible in phase diagnostics

colinhacks:
Annotate aube's `RUST_LOG=debug` `phase:*` install lines so a zero-work warm phase isn't misread as real work.

- Every `phase:*` line carries its work count; the two `phase:resolve` lines had none.
- A zero-work phase gets an explicit, greppable ` [no-work: warm/up-to-date]` marker.

    - phase:fetch 31.5ms (0 packages)
    + phase:fetch 31.5ms (0 packages) [no-work: warm/up-to-date]

Debug-log content only; default output unchanged (emits only under `RUST_LOG=debug`).

Refs the startup-overhead diagnostics work.


colinhacks:

PR #237, opened 2026-06-28
ci(release): publish Docker images on every release via reusable workflow

colinhacks:
`ghcr.io/nubjs/nub` is frozen at v0.2.0: `docker.yml` fires on `release: published`, which never triggers for an automated release because `release.yml` cuts the Release with the default `GITHUB_TOKEN` (GitHub does not cascade `GITHUB_TOKEN`-raised events).

Fix: make `docker.yml` a reusable workflow (`workflow_call`, `nub_version` input) called from a `release.yml` `docker` job (`needs: publish-npm`), so build+push runs in the same run after npm publish. Push keys on `inputs.nub_version` presence (not `github.event_name`), so PR/push builds stay build-and-smoke-only. Adds `workflow_dispatch` for backfill. slim+alpine dual-tag design unchanged; actionlint clean.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -4,9 +4,20 @@ name: Docker image
 # released @nubjs/nub npm package onto an official, digest-pinned node:* base.
 #
 #   - On PR / push to main: build BOTH variants for the runner arch and run the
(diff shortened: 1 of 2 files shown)


colinhacks:

PR #239, opened 2026-06-29
cli: keep `--` on `nub watch` + durable target-passthrough test

colinhacks:
nub keeps the post-target `--` verbatim across `nub <file>`, `run`, `exec`, and `watch` (maintainer decision, Option A): the target token ends runner parsing and everything after it forwards byte-for-byte — `--` neither required nor special-cased. Matches `node` and `pnpm 10`.

Removes the first-`--`-strip from `split_subcommand_argv` (uniform; no per-subcommand carve-out). The leading `--` (end of runner options, before the target) is still consumed, as in node/pnpm. Also fixes the prior `nub watch` subcommand-vs-flag inconsistency, now subsumed by the uniform keep.

Adds an oracle-grounded table-driven test (`target_passthrough.rs` + fixture) pinning target termination + verbatim passthrough across file/run/exec/watch, plus a split-level unit test. Goldens from node 26.2.0 / pnpm 10.15.1.

Refs: target-token-passthrough-hardening thread.


colinhacks:

PR #242, opened 2026-06-29
fix(pm): close brand-boundary leak on aube's user/project config-file path

colinhacks:
## Problem

Under the NUB embedder profile, the vendored aube engine read an `aube`-branded user-authored config file during settings resolution — `~/.config/aube/config.toml` and `<cwd>/.config/aube/config.toml` — and could author one via `nub config set`. That violates the symmetric, public-surface brand boundary: nub must not read another tool's branded config paths, just as it never reads a pnpm-named file off-incumbent.

The asymmetry was the bug: the *managed/system* config path already routes through the embedder profile (`/etc/<dir>/managed.toml`, `{config_env_prefix}_MANAGED_CONFIG_PATH`), but `user_aube_config_path()` / `project_aube_config_path()` hardcoded `aube` with no `embedder()` lookup.

### Empirical verdicts

- **Config-file leak — REAL.** `nub config get minimumReleaseAge` returned a planted `~/.config/aube/config.toml` (4321) and `<cwd>/.config/aube/config.toml` (7777, project outranks user).
- **No nub-branded config file is read today.** Planted `.config/nub/config.{toml,yaml}` → ignored. Nothing to remove.
- **The env-var side is already shut.** `nub store path`: the neutral `NPM_CONFIG_STORE_DIR` redirects the store; the branded `AUBE_STORE_DIR` is ignored. The gate (`branded_env_alias_enabled`) is applied at resolve time for every `sources.env` alias, and `branded_settings_env_gate.rs` exercises that resolve path. No change needed there.

## Fix

A new embedder field `config_namespace: Option<&'static str>` — the XDG leaf dir for the tool's own user/project config file, mirroring the existing `managed_config_system_dir: Option<&'static str>` precedent.

- Standalone aube: `Some("aube")` — `~/.config/aube/config.toml` byte-for-byte as before.
- nub: `None` — the branded user/project config file is disabled entirely. nub reads it nowhere and writes it nowhere; a nub project's config surface stays `.npmrc` (neutral) plus the sanctioned `NUB_*` env knobs. nub does **not** substitute a `~/.config/nub/` home of its own.

`user_aube_config_path()` now returns `Result<Option<PathBuf>>` and `project_aube_config_path()` returns `Option<PathBuf>`. Read loaders return empty on `None`; write paths refuse on `None` with a new coded error `ERR_AUBE_NO_BRANDED_CONFIG_FILE`. Unknown and npm-shared keys still route to the neutral `.npmrc` unchanged.

## Default-preserving

Standalone aube (no profile registered → `AUBE`, `Some("aube")`) is byte-identical: same paths, same write routing, same `config set` workspace-yaml fallback, same `missing key` messages. Verified by the in-crate default-profile unit tests.

## Verification

- Re-ran the empirical repro on the fixed binary: the leak is closed (`config get` → `undefined`), a neutral `.npmrc` value is still honored, and `config set` writes no `.config/aube/` file.
- New nub-cli e2e gate test `nub_profile_reads_no_branded_user_or_project_config_file` (plants both scopes for both brands; asserts none are read).
- New default-profile unit test pins the AUBE path; the NUB `None` is pinned by a compile-time assertion in `pm_engine::identity`.
- Green: aube-codes/util/settings/lib-config/lockfile/manifest/scripts/registry suites, nub-cli `pm_identity`, `cargo clippy --all-targets --all-features -D warnings`, `cargo fmt --check`.

## Impact analysis

Every caller of the two path functions (the read loaders in `aube_config.rs`, and the `config set`/`config delete` write paths) and the `FileSources`/`ResolveCtx` precedence chain were traced: under `None` only the branded-file source is removed from resolution; no other setting source changes. The new error code is registered in the `ALL` array (self-tests gate duplicates/typos).

Refs the brand-boundary work.

## Write-side routing (note)

Under nub, real `config set` keys (npm-shared, pnpm-only, and unknown alike) route to the neutral `.npmrc` — verified empirically; none author a branded config file. The `None` refusal is a defensive guard on the branded-file write destination; the e2e test asserts a `config set` leaves the planted branded files untouched and lands on `.npmrc`.

One deliberate, maintainer-aligned choice worth flagging: for the narrow case of a non-npm-shared known setting written at project scope with an existing `pnpm-workspace.yaml`, the `Some`-path workspace-yaml fallback is preserved for standalone aube but is refused under nub (no branded config file). This does not manifest for any common key (they route to `.npmrc`), and refusing is consistent with nub's stated config surface (`.npmrc` + `NUB_*`). Flagging it for an explicit call rather than burying it.

### DIFF
--- a/crates/nub-cli/src/pm_engine/identity.rs
+++ b/crates/nub-cli/src/pm_engine/identity.rs
@@ -87,6 +87,15 @@
 ///   aube cannot make nub silently inherit its `/etc/aube` policy: the brand
 ///   boundary holds on the system path the same way it already held on the env
 ///   override (`NUB_MANAGED_CONFIG_PATH`, via `config_env_prefix`).
+/// - `config_namespace` = `None` — nub has NO branded user/project config file.
+///   The engine never reads `~/.config/aube/config.toml` or
+///   `<cwd>/.config/aube/config.toml` (the leak this profile closes), and nub
+///   does NOT substitute a `~/.config/nub/` home of its own: a nub project's
+///   config surface is the neutral `.npmrc` + the sanctioned `NUB_*` env knobs,
+///   so there is no bespoke nub config file to author or read. Mirrors how
+///   `managed_config_system_dir = None` would skip the system read — here the
+///   user/project branded-file read/write is skipped entirely. Standalone aube
+///   keeps `Some("aube")`, so its `~/.config/aube/config.toml` path is unchanged.
 /// - `canonical_lockfile_always_wins` = `false` — `lock.yaml` never silently
 ///   outranks a foreign lockfile beside it; that state is the loud
 ///   ambiguity/contradiction error (was
@@ -155,6 +164,10 @@ pub(crate) const NUB: aube_util::Embedder = aube_util::Embedder {
     cache_namespace: "nub/pm",
     data_namespace: "nub",
     managed_config_system_dir: Some("nub"),
+    // No branded user/project config file: nub never reads `~/.config/aube/`
+    // (or `<cwd>/.config/aube/`) and authors no `~/.config/nub/` of its own —
+    // a nub project's config surface is `.npmrc` + the `NUB_*` env knobs.
+    config_namespace: None,
     canonical_lockfile_always_wins: false,
     runtime_switching: false,
     self_engines_check: false,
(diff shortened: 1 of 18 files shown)


colinhacks:

PR #244, opened 2026-06-29
runtime: make an unhandled worker 'error' fatal, matching node:worker_threads

colinhacks:
A worker created with no `'error'`/`onerror` listener that failed to load — or threw at top level — exited 0 in silence. nub's browser-shape global `Worker` is an `EventTarget`, and `dispatchEvent` drops an event with no listener, so the worker failure vanished. The oracle, `node:worker_threads.Worker`, is an `EventEmitter` that re-throws an unhandled `'error'` on the main thread (`process.nextTick(() => { throw err })`) → uncaughtException → exit 1 with the stack printed.

## Fix

The main-thread `Worker` now tracks its `'error'` listeners (a registry keyed on EventTarget's `(listener, capture)` identity, maintained by `addEventListener`/`removeEventListener`; the `onerror` setter routes through `addEventListener`, so it is counted too). When the underlying worker emits `'error'` with none registered, the wrapper re-throws on a fresh tick — mirroring the oracle exactly: visible, fatal (exit 1), and still interceptable by a user `process.on('uncaughtException')`. The listener-present path is unchanged.

The registry keys on capture (not just the function reference) so a capture-phase `removeEventListener` cannot empty it while a live bubble-phase listener remains (which would be a false fatality). `{once}`/`{signal}` auto-removal stays un-mirrored, but a stale entry there errs toward delivery, never toward a false fatality.

## Verification

- New regression tests: `worker_unhandled_error_is_fatal` and `worker_error_listener_capture_identity_not_a_false_fatality`, plus fixtures.
- Verified e2e on Node 26 (fast tier), 20.19 and 18.19 (compat tier + floor): the no-listener resolve-failure exits 1 with the error on stderr; the throwing worker with `onerror` exits 0 (parent survives).
- Worker integration suite (17 tests) and the WPT worker gate (606 pass, 0 unexpected-fail) green; `clippy --all-targets --all-features -D warnings` and `fmt --check` clean.

## Conformance sweep

Ran Node's own worker suite (141 files: `test/parallel/test-worker-*`, `test-messageport-hasref`, `test-structuredClone-*`) under augmented nub vs plain `node`. 140 match. The only divergence is `test-worker-memory.js`, a heuristic asserting a worker-memory ratio below 5×. nub forwards its transpile/polyfill preload into each worker — load-bearing, it is how `Worker(new URL("./x.ts"))` transpiles its own TS entry — so each augmented worker has a larger footprint. The ratio crosses 5× only at high worker concurrency (passes at ≤4 concurrent workers, fails at ≥8); it is concurrency-driven sticky RSS, not a leak, and not a node:worker_threads-conformance bug. Left as-is and surfaced for a maintainer call on whether the per-worker preload footprint is worth trimming.

Refs: worker oracle-conformance (maintainer-directed; no issue number).

### DIFF
--- a/crates/nub-cli/tests/integration.rs
+++ b/crates/nub-cli/tests/integration.rs
@@ -2654,6 +2654,30 @@ fn worker_throw_surfaces_to_parent_onerror() {
 }
 
 #[test]
+fn worker_unhandled_error_is_fatal() {
+    // ORACLE conformance (node:worker_threads): a worker 'error' with NO listener
+    // must be FATAL — surface visibly + nonzero exit — not silently swallowed.
+    // The bug: nub's browser-shape Worker is an EventTarget, whose dispatchEvent
+    // drops an unlistened event, so a failed worker load exited 0 in total silence
+    // (maintainer-found). The fixture spawns a worker whose module fails to
+    // resolve with no onerror; nub must now exit nonzero and print the error.
+    let (stdout, stderr, code) = run_nub("worker", "unhandled-error-fatal-main.ts");
+    assert_ne!(
+        code, 0,
+        "an unhandled worker error must be fatal (nonzero exit), not swallowed: \
+         code={code} stdout={stdout} stderr={stderr}"
+    );
+    assert!(
+        !stdout.contains("swallowed-and-survived"),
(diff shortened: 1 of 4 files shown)


colinhacks:

PR #245, opened 2026-06-30
Modify extension check in resolve.rs

### LINKED ISSUE #240: Files with dot in name won't load
I am trying to load files with a dot(.) in the filenames, but trying to do it as extensionless it doesn't work. This is the style of nest.js.

I have traced the potential fix to this line: https://github.com/nubjs/nub/blob/46d1269e931b66a3ebba7352f533d626b94c705b/crates/nub-native/src/resolve.rs#L113

### Steps to replicate

1. Create a file named `greeting.service.ts`.
2. Try to import it using `import './greeting.service'`.

ctorres-lynxlabs:
This should fix #240


colinhacks:

PR #248, opened 2026-06-30
fix(node-hijack): pass through args after the entry point verbatim

colinhacks:
The `node` PATH-hijack stripped `--node` from anywhere before a `--`, so a program-arg `--node` (`node app.js --node`) was eaten and compat silently flipped on. Real `node` passes every arg after the entry point to the script verbatim.

`scan_node_compat_flag` now does a leading-only scan mirroring the blessed `run_nub` file-run path: a leading standalone `--node` is the compat opt-out (consumed, vanilla); the entry point (first bare token, `-`/stdin, `-e`/`-p` eval, or the token after `--`) ends interpretation and the rest forwards untouched. With no `--node`, forwarded argv equals input byte-for-byte. Verified differentially against real `node`.

Refs #224.


colinhacks:

Issue #241, opened 2026-06-29
nub prune --prod drops production transitive deps under 0.2.9 node_modules layout (ERR_MODULE_NOT_FOUND)


tstysial:
## Summary

Under nub **0.2.9**'s new default `node_modules` layout, `nub prune --prod` removes **transitive dependencies of declared production dependencies**, breaking the runtime with `ERR_MODULE_NOT_FOUND`. This happens for packages whose transitives are peer-dependency-disambiguated in the store (e.g. `@orpc/server`, which pulls `@orpc/contract` / `@orpc/shared`).

This regressed between 0.2.8 and 0.2.9 (the node_modules-layout change, #238). 0.2.8 and 0.2.3 are unaffected.

## Minimal reproduction

```jsonc
// package.json
{
  "name": "prune-repro",
  "private": true,
  "type": "module",
  "dependencies": { "@orpc/server": "^1.13.14" },
  "devDependencies": { "typescript": "^5.9.3" }
}
```

```ini
# .npmrc
node-linker=hoisted
```

```ts
// app.ts
import { os } from "@orpc/server";
console.log("orpc/server loaded:", typeof os);
```

```bash
nub install
nub app.ts            # → "orpc/server loaded: object"   ✅
nub prune --prod      # → "Pruned 19 entries: 17 top-level, ..."
nub app.ts            # → ERR_MODULE_NOT_FOUND            ❌
```

### Actual

After `nub prune --prod`, the run fails:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@orpc/contract'
  imported from .../node_modules/@orpc/server/dist/index.mjs
```

`node_modules/@orpc/shared` and `node_modules/@orpc/contract` are gone — pruned away even though `@orpc/server` (a declared production dependency) requires them. Prune reports removing 17 top-level entries, including these production transitives.

### Expected

`nub prune --prod` should keep the full production dependency closure (transitives of declared `dependencies`), removing only `devDependencies` and their exclusive transitives. The app should still run after prune, as it does on 0.2.8.

## Also affected: default isolated layout

Without the `node-linker=hoisted` opt-out (i.e. 0.2.9's new default isolated layout), the same project after `nub prune --prod` fails differently at runtime — a **declared** dependency is reported as a phantom:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'express' imported from .../server/index.ts
"express" is installed but not reachable here — it's a phantom dependency ...
```

[...]

PR #250, opened 2026-06-30
fix(pm): keep hoisted prod transitives on prune --prod

colinhacks:
`nub prune --prod` deleted production transitive dependencies under the hoisted node-linker, breaking the install with `ERR_MODULE_NOT_FOUND`.

## Cause

prune built each importer's top-level keep-set from its direct deps only. The hoisted linker (and isolated `shamefully-hoist`) flattens the whole reachable closure into top-level `node_modules/`, so production transitives live there but appear in no importer's direct-dep list — and prune deleted them.

## Fix

- When the on-disk layout flattens transitives to the importer root, union the reachable production-closure package names into the top-level keep-set. Plain isolated stays direct-deps-only, matching pnpm — a package that is both a direct devDependency and a production transitive loses its top-level symlink on `prune --prod` rather than being retained.
- Read the layout from the recorded install state, which captures a `--node-linker=hoisted` CLI flag that was never written to `.npmrc` (re-resolving settings alone would miss it); fall back to settings for `shamefully-hoist` and for trees installed by another tool.
- Sweep dangling top-level symlinks, scoped to symlinks that point into the virtual store. A name-matched symlink can point at a virtual-store version the store sweep removed (a direct devDependency and a production transitive sharing a name at different versions); it is now cleaned instead of left dangling. The scope spares `link:`/`portal:` deps, whose target may legitimately be absent at prune time.

## Tests

Added regression tests for the npmrc-hoisted, CLI-flag-hoisted, and `link:`-preservation cases, and wired `prune.bats` into the nub-adapted bats suite with skips for the pre-existing `.aube` store-path/label and help-text divergences. Verified against pnpm 10.15.1 across hoisted, isolated, isolated + `shamefully-hoist`, and the name/version-collision case.

This is a latent correctness bug in standalone aube as well; the fix lives in `vendor/aube` and is default-preserving (no embedder-specific behavior).

Refs #241 — fixes Bug #1 (hoisted `prune --prod` drops production transitives). Bug #2 (the GVS multi-stage-Docker `node_modules` relocatability regression) is tracked separately and stays open, so this does not close the issue.

### DIFF
--- a/tests/aube-bats/run.sh
+++ b/tests/aube-bats/run.sh
@@ -31,7 +31,7 @@ shift
 
 SUITES=("$@")
 if [ ${#SUITES[@]} -eq 0 ]; then
-  SUITES=(install.bats ci.bats add.bats remove.bats update.bats lockfile_settings.bats lockfile_dir.bats)
+  SUITES=(install.bats ci.bats add.bats remove.bats update.bats prune.bats lockfile_settings.bats lockfile_dir.bats)
 fi
 
 SCRATCH=$(mktemp -d)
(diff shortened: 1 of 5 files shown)


colinhacks:

PR #253, opened 2026-06-30
feat(runtime): publish process.versions.nub self-identification marker

colinhacks:
Augmented `nub` sets `process.versions.nub` to the binary's version, so tooling can detect "running under nub" like `process.versions.bun` / `.electron`. Read-only marker, not an importable API.

The Rust spawn layer passes `env!("CARGO_PKG_VERSION")` to the shared preload via the internal `__NUB_VERSION` var (coupled to preload injection); published with `Object.defineProperty` matching Node's native entries (enumerable, configurable, non-writable), both tiers. Absent under `--node`/`NODE_COMPAT`.

Verified on Node 26/20.19/18.19; absent under `--node` and `NODE_COMPAT=1`. Test: `process_versions_nub.rs`.


colinhacks:

PR #254, opened 2026-06-30
fix(node): stop auto-injecting --experimental-shadow-realm (Electron crash #246)

### LINKED ISSUE #246: Electron crashes with EXC_BREAKPOINT during `node::CreateEnvironment` when nub injects NODE_OPTIONS
## Environment

- **nub**: 0.2.6
- **Electron**: 42.5.1
- **electron-vite**: 5.x
- **macOS**: 26.5.2 (Tahoe), ARM64 (Apple Silicon)
- **Node**: 24.18.0

## Repro

```sh
nub run dev   # dev script = electron-vite dev
```

Electron crashes immediately with `EXC_BREAKPOINT (SIGTRAP)` — no window opens.

## Crash stack (Thread 0)

```
Exception Type:  EXC_BREAKPOINT (SIGTRAP)
Termination Reason: Namespace SIGNAL, Code 5

0   v8::Object::GetAlignedPointerFromEmbedderDataInCreationContext  +47500
1   v8::String::MakeExternal
2-5 v8::Array::New
6   v8::CppHeapExternal::NewImpl
7   v8::Context::New
8   v8::Context::FromSnapshot
9   node::CreateEnvironment
...
12  node::Dotenv::GetDataFromArgs  +46052
13  node::Dotenv::GetDataFromArgs  +45024
...
22  ElectronMain  +1880
```

The crash occurs during `node::CreateEnvironment`, triggered by `node::Dotenv::GetDataFromArgs` — consistent with nub injecting `--env-file` (and/or `--import`) via `NODE_OPTIONS` into the Electron subprocess.

## Workaround

```sh
NODE_COMPAT=1 nub run dev   # disables all augmentation → Electron opens normally
```

Or in `package.json`:
```json
"dev": "NODE_COMPAT=1 electron-vite dev"
```

## Why this is a nub issue

When `nub run dev` spawns `electron-vite dev`, which in turn spawns the Electron binary, Electron's embedded Node inherits nub's `NODE_OPTIONS` (containing `--import` and/or `--env-file`). Electron's embedded Node is a patched/modified build that crashes when processing these flags during `CreateEnvironment`.

This is the same pattern as #18 (Next.js crashing under nub augmentation), where the fix was on the nub side. The expected fix here would be for nub to either:

1. Detect when the child binary is an Electron binary and skip `NODE_OPTIONS` injection
2. Strip flags unsupported by Electron's embedded Node (e.g. `--env-file`) from `NODE_OPTIONS` for non-standard Node processes
3. Document that `NODE_COMPAT=1` is required for Electron-based projects

## Additional notes

- Running `electron-vite dev` directly (without `nub run`) works fine
- `nubx electron --version` also crashes the same way — just the crashpad error, no version printed
- Same crash on Windows (different error surface — crashpad `not connected` only, no window)
- Related: #18 (Next.js augmentation crash, fixed in v0.1.3)

colinhacks:
## Problem

`nub run dev` crashes Electron immediately with `EXC_BREAKPOINT (SIGTRAP)` in `node::CreateEnvironment` (#246).

`--experimental-shadow-realm` implies V8's `--harmony-shadow-realm`, which is part of the isolate's V8-flag hash. nub injected it into the tree-wide `NODE_OPTIONS`, which is inherited by the whole process subtree — including embedded Node runtimes that boot from a V8 context snapshot, notably Electron. When the runtime flag hash differs from the snapshot's, `v8::Context::FromSnapshot` returns empty and `CreateEnvironment` aborts on a failed `CHECK`. This happens at startup, **before any `--import`/`--require` preload JS runs**, so a preload self-disable cannot prevent it.

Reproduced against the reporter's exact environment — Electron 42.5.1 (Node 24.17 / V8 14.8): `--experimental-shadow-realm` alone crashes the main process deterministically; every other flag nub injects passes. The crash is V8-snapshot-version-dependent (Electron 37 / V8 13.8 tolerated the same flag), so the fix is a categorical policy, not a per-version carve-out.

## Fix

- **Drop the ShadowRealm unflag** from `feature_matrix::FEATURES` so the flag is never injected.
- **Establish the policy:** never auto-unflag a V8 harmony / startup-snapshot-affecting flag into `NODE_OPTIONS`; only snapshot-neutral Node feature flags are eligible. Enforced by `V8_HARMONY_UNFLAG_DENYLIST` + the `no_v8_harmony_flag_in_unflag_set` guard test (a re-add trips CI), and documented at the matrix definition site.
- **Electron self-disable in the fast-tier preload** (`if (process.versions.electron) return;`): once the flag crash is removed, the preload's `module.registerHooks` load hook independently deadlocks Electron's main process. Both pieces are required for Electron to launch clean.

## Verification

- `nub run dev` → real Electron 42.5.1 now launches clean (was a deterministic SIGTRAP).
- Compat tier (host Node 20 → Electron 30) unaffected: the async `module.register` preload does not hang Electron, and no harmony flag is injected.
- `cargo clippy --all-targets --all-features -- -D warnings`, `cargo fmt --check`, and the scoped `nub-core` lib tests pass.

## Tradeoff

`globalThis.ShadowRealm` (TC39 Stage 3) is no longer auto-provided; deferred until Node ships it default-on (no flag needed, no snapshot hazard). Users who want it today can pass `--experimental-shadow-realm` themselves — nub forwards user-supplied flags verbatim.

Closes #246

### DIFF
--- a/crates/nub-core/src/node/feature_matrix.rs
+++ b/crates/nub-core/src/node/feature_matrix.rs
@@ -149,23 +149,28 @@ pub static FEATURES: &[Feature] = &[
         )],
         evidence: "flag added Node 9.6.0 (#14253); never unflagged through Node 26",
     },
-    // ── ShadowRealm global ──────────────────────────────────────────────────
-    // `--experimental-shadow-realm` gates `globalThis.ShadowRealm` (TC39 Stage 3:
-    // an isolated global with fresh intrinsics). The flag was added in Node
-    // 18.13.0 / 19.0.0 — below nub's 18.19 floor — and has NEVER been made
-    // default-on through Node 26 (still experimental, rides the V8
-    // `--harmony-shadow-realm` staging). So, like vm-modules, inject across the
-    // ENTIRE supported floor: [18.19.0, ∞). It survives as an accepted bool at
-    // every higher version, so the open-ended band never over-injects into an
-    // unsupported range.
-    Feature {
-        name: "shadow-realm",
-        mitigations: &[(
-            band((18, 19, 0), None),
-            Mitigation::Unflag("--experimental-shadow-realm"),
-        )],
-        evidence: "flag added Node 18.13.0/19.0.0; never default-on through 26 (TC39 Stage 3)",
-    },
+    // ── ShadowRealm — DELIBERATELY NOT AUTO-UNFLAGGED (the harmony-flag policy) ──
+    // POLICY (load-bearing — enforced by `no_v8_harmony_flag_in_unflag_set` and
+    // listed in V8_HARMONY_UNFLAG_DENYLIST below): nub NEVER auto-injects a V8
+    // *harmony* / startup-snapshot-affecting flag into the tree-wide NODE_OPTIONS.
+    // `--experimental-shadow-realm` implies V8's `--harmony-shadow-realm`, which is
+    // part of the isolate's V8-flag hash. NODE_OPTIONS is inherited by the WHOLE
+    // process subtree, including EMBEDDED Node runtimes that boot from a V8 context
+    // SNAPSHOT (Electron). When the runtime V8-flag hash ≠ the snapshot's, V8's
+    // `Context::FromSnapshot` returns empty and `node::CreateEnvironment` aborts on a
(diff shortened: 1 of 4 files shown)


colinhacks:

PR #255, opened 2026-06-30
feat(pm): stamp packageManager: nub@<v> on a virgin install

colinhacks:
Every package.json-modifying op (`install`/`add`/`remove`/`update`) writes `"packageManager": "nub@<version>"` into `package.json` on a **virgin** project — where nub is the first PM to touch it. The common case: `nub add <pkg>` as the first command. nub's lockfile is neutral (`lock.yaml`), so the corepack field is the only PM signal downstream tools (turbo) can read.

**Virgin predicate** (maintainer-confirmed). Stamp iff the op succeeds and ALL hold:
- no `packageManager` field already;
- no lockfile of any PM (npm/pnpm/yarn/`bun.lock`/`bun.lockb`) and no nub lockfile (⇒ writes once, never on re-ops);
- no other incumbent (`devEngines.packageManager`, pnpm-named file, lone `.yarnrc.yml`, workspace root).

Any PM signal ⇒ no write (symmetric brand boundary). `truly_fresh` is captured before the op writes the lockfile ⇒ no churn on the 2nd op. `import` is excluded (it converts a foreign lockfile ⇒ never virgin). Only `packageManager`, no caret. clippy `--all-targets` + fmt clean; 129 + 3 tests green.

### DIFF
--- a/crates/nub-cli/src/pm_engine/install_family.rs
+++ b/crates/nub-cli/src/pm_engine/install_family.rs
@@ -360,11 +360,13 @@ fn run_add(typed: &str, args: &[String]) -> Result<i32> {
             &yarn_remedy("add", &verb.packages),
         ));
     }
-    finish_quieted(
+    let code = finish_quieted(
         &globals.output,
         &session,
         aube::commands::add::run(verb, globals.effective_filter()),
-    )
+    )?;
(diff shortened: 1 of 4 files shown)


colinhacks:

PR #256, opened 2026-06-30
feat(worker): additive node-mirroring on the global Worker polyfill

colinhacks:
Additive Node `worker_threads` mirroring on the global `Worker`; the web shape (`EventTarget`/`MessageEvent`/`ErrorEvent`) stays primary, WPT gate green.

- `eval: true` inline source (gated on the option).
- `online`/`exit` as `Event` objects; `exit` carries `.code`.
- `terminate()` returns the underlying `Promise<exitCode>`.
- `.on`/`.once`/`.off`/`.emit`/`addListener`/`removeListener` delegated to the node Worker; web channel unchanged; unhandled-`error` fatality counts both channels.
- `workerData` regression test.

6 new + 1 rewritten test (replaces `worker_has_no_exit_event`). WPT green on Node 26 + 20.19. `@nubjs/types` updated. Design: web-worker plan.


colinhacks:

PR #259, opened 2026-07-01
cli: suggest `nub exec` for an unknown bareword binary

colinhacks:
`nub turbo login` / `nub eslint` errored as not-a-nub-command and pointed only at `nub run` and `nub ./` — but a dependency binary runs via `nub exec <name>` (alias `nubx <name>`). This final bail fires only after the known-script and plugin branches, so the token is most likely a binary; the hint now leads with `nub exec`.

Before:

```
  (to run a script: nub run turbo · to run a file: nub ./turbo)
```

After:

```
  (to run a dependency's binary: nub exec turbo or nubx turbo · to run a script: nub run turbo · to run a file: nub ./turbo)
```

`bareword_unknown_verb_errors` now asserts the hint includes `nub exec`/`nubx`.


colinhacks:

PR #260, opened 2026-07-01
fix(exec): set npm_config_user_agent on the bin-launch path

colinhacks:
`nubx` / `nub exec` launched a bin without `npm_config_user_agent`, so `create-*` scaffolders detected nub as npm. The run and lifecycle paths already set a role-aware UA; the bin-launch path did not.

Both `launch_bin` branches now set it, reusing `run_lifecycle_ua_product` + a shared `scripts::user_agent_string` helper: `nub/<v> npm/? …` under nub identity, incumbent-first (`pnpm/<pin> nub/<v> …`) in a compat project. A plain `nub <file>` run and compat mode (`--node`/`NODE_COMPAT`) stay unset. New test `exec_bin_reports_role_aware_user_agent`.

Out of scope: the engine dlx path (`nub x`/`dlx`/`create`, in aube's `exec_bin`) still lacks it — an aube-side follow-up.

### DIFF
--- a/crates/nub-cli/src/cli.rs
+++ b/crates/nub-cli/src/cli.rs
@@ -4409,7 +4409,11 @@ fn run_exec_with_dlx(
             if let Some(runner) = pnp_bin_runner_path() {
                 let mut cmd_args = vec![runner, bin.to_string()];
                 cmd_args.extend(args.iter().cloned());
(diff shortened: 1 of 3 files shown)


colinhacks:

PR #265, opened 2026-07-01
feat(pm): honor packageManager: nub@x.y.z by provision + delegate (self-shim)

colinhacks:
A workspace-root `packageManager` pin of a *different* exact nub than the running one now provisions that nub from nub's own release channel and execs it — corepack's model, for nub itself. Auto-delegate is on by default. Scope: the PM mutating verbs (`install`/`i`/`ci`/`add`/`remove`/`update`/`import`/`dedupe`); the file runner, `run`/`watch`, `nubx`, `upgrade`, and `node` are exempt and pay only an argv[1] string compare.

## Integrity — read this first

This is a default-on download-and-exec of a binary named by (potentially untrusted) repo content, so the integrity story is load-bearing.

**What nub's `+sha512` actually hashes:** nothing usable here. nub's own self-pin is written as a *bare* `nub@<version>` — no `+sha512` — by both write paths (`install_family.rs` #255 virgin stamp, `use_nub.rs` for `nub pm use nub`). The `+sha512.<hex>` suffix only exists for *foreign* PM pins (pnpm/yarn/npm), where it hashes the npm-registry tarball of that PM. It never applies to nub, and one hash could not cover nub's 8 per-platform release binaries anyway (corepack's `+sha512` works only because a pnpm/yarn npm tarball is one platform-independent artifact).

**So integrity rides the release channel, not the pin.** `provision_self` downloads `<base>/v<ver>/nub-<target>.tar.gz`, fetches the published `.sha256` sidecar, and SHA-256-verifies the bytes *before* extraction — the same mechanism `nub upgrade` already uses. It then probes `<bin> --version` and refuses to exec a version-mismatched artifact. A store hit is the trust cache (the version-addressed `~/.cache/nub/self/<ver>/`, created `0700`). No unverified or mismatched binary reaches `exec`.

**Named residuals** (accepted, inherent to the class — the same trust model as `nub upgrade` / corepack / nvm):
- The tarball and its `.sha256` come from the same origin, so the checksum defends corruption/truncation but is not a signature — a compromised release host or a valid-cert MITM would control both. A per-release signature layer is a follow-up.
- The shim is a downgrade primitive: repo content selects *which published nub* runs your install. It can only run a genuine, checksum-matching, officially-published nub release — never an arbitrary attacker binary — but a repo can pin an older one. Unavoidable given the reproducibility goal.

The `NUB_SELF_SHIM=0` env var disables the handoff tree-wide, and every failure mode names it so a bad committed pin can't brick a clone.

## Behavior

- Delegation fires only on an exact `nub@X.Y.Z` != the running version, with the re-entry guard unset and the opt-out off. A matching pin, a foreign pin, or a `devEngines`-only nub entry runs in place. A non-exact pin (the `packageManager` field is exact-only, as corepack enforces) prints a notice and runs in place. A platform with no published build runs in place with a notice — never a hard error.
- Loop-safe: the child carries `__NUB_SELF_DISPATCHED=<version>`, whose presence unconditionally suppresses any further delegation. Exact-version-only pins mean a spec can't resolve to a different concrete string across the exec.
- Reads the exact `packageManager` field only (`declared_package_manager_field`), never the `devEngines` range-of-intent.

## Tests

- `crates/nub-cli/src/self_shim.rs` — 11 unit tests over the pure decision (allowlist, the `+sha512`-strip regression through the real field reader, matching/mismatch/opt-out/non-exact/unsupported-platform/re-entry/devEngines-not-a-pin).
- Verified end-to-end against a `file://` release fixture (the `NUB_RELEASE_*` seams): a differing exact pin provisions + verifies + execs with the guard set; the cache hit skips re-download; a `+sha512`-suffixed matching pin is a silent no-op; the hot-path verbs never provision; the opt-out and non-exact pins notice without delegating; a version-mismatched artifact hard-errors before running.

## Self-review

Three fresh-context reviewers (correctness / impact-analysis / security). The verify-before-exec invariant holds. Fixes folded in: the platform fallback (a pin mismatch on an unprovisionable platform ran in-process instead of hard-erroring), the `packageManager`-field-only read (was reading `devEngines`), the escape hatch named in every bail, and `0700` on the self store.

## Follow-ups

- A per-release binary signature layer (closes the same-origin residual).
- GC / `nub pm cache clear` coverage for the new `~/.cache/nub/self` store.
- Real win32 provisioning (win32 ships `.zip`, not `.tar.gz`; today Windows falls through to the local nub with a notice).
- `nub pm which`/`update` still print "unsupported package manager nub" on a nub pin (non-mutating, out of this scope; a `Pm::Nub` variant would be a large blast radius).

https://claude.ai/code/session_01FRRTvRH4BLKmxtbx8GMsTH

### DIFF
--- a/crates/nub-cli/src/cli.rs
+++ b/crates/nub-cli/src/cli.rs
@@ -4839,7 +4839,7 @@ fn detect_channel(bin_path: &Path) -> UpgradeChannel {
 /// error rather than fetching a 404). musl vs glibc on Linux is a runtime
 /// distinction install.sh makes via `ldd`/`/etc/alpine-release`; we encode the
 /// glibc default here and document musl as the known gap (see note below).
-fn platform_target() -> Option<&'static str> {
+pub(crate) fn platform_target() -> Option<&'static str> {
     // NOTE: a glibc build and a musl build of the same arch are the same Rust
     // `target_env` only under `target_env = "musl"`, so this distinguishes them
     // correctly when Nub itself is built for musl. The detection matches the
@@ -5332,6 +5332,16 @@ fn provision_self(version: &str) -> Result<PathBuf> {
 
     std::fs::create_dir_all(&store)
         .with_context(|| format!("creating the nub self store at {}", store.display()))?;
+    // Defense-in-depth: the self store holds binaries this process EXECs, and a
+    // store hit skips re-verification, so restrict it to the owner (0700) — a
+    // shared/world-writable cache dir can't be used to pre-plant a binary a later
+    // run would exec unverified. Best-effort; a failure here doesn't block the
+    // provision (the verify-before-exec chain below is the real gate).
+    #[cfg(unix)]
+    {
+        use std::os::unix::fs::PermissionsExt;
+        let _ = std::fs::set_permissions(&store, std::fs::Permissions::from_mode(0o700));
+    }
     let staging = tempfile::Builder::new()
         .prefix(&format!(".self-{version}-"))
         .tempdir_in(&store)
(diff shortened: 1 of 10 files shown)


colinhacks:

PR #266, opened 2026-07-01
feat(pm): virgin install writes a devEngines caret range, not an exact packageManager pin

colinhacks:
Rolls back the exact `packageManager: nub@<v>` auto-pin #255 landed on virgin `nub install` (main-only, never released — exact pin + self-shim is lock-in).

- Virgin `nub install` writes `devEngines.packageManager = { name: "nub", version: "^<v>", onFail: "warn" }`, never `packageManager`.
- `nub pm use nub@<exact>` opts into the hard `packageManager` pin; a range/dist-tag is refused.
- Bare `nub pm use nub` writes range-only, clearing any incumbent `packageManager`.
- No `devEngines.runtime` write.

The self-shim reads `devEngines.packageManager` when `packageManager` is absent, so a devEngines-only nub project stays nub-owned.

Refs #255, #265.


colinhacks:

PR #267, opened 2026-07-01
fix(env): ignore NODE_ENV set in a .env file

### LINKED ISSUE #263: `nub run` breaks Next.js `next build` static generation — React `useContext` null / `workUnitAsyncStorage` invariant in build worker threads
## Summary

Running a Next.js 16 production build via **`nub run build`** breaks static generation: every prerender worker throws React's `TypeError: Cannot read properties of null (reading 'useContext')` (Turbopack) — i.e. React's hooks dispatcher is `null` inside Next's build worker threads. The build fails on a **different, random page each run** (including Next-internal pages like `/_global-error` and `/_not-found`), which is the signature of a **module-singleton being duplicated/not shared into the worker threads**, not any page's own code.

The exact same build **passes** when the same script is run via `pnpm run build` or by invoking the project's Next binary directly. So the differentiator is purely **the runner (`nub run`) wrapping Next's build**, not the code, the install, the bundler, or Next.

This blocks local production builds for our app. (Our Vercel builds run `pnpm run build:ci`, so production is unaffected — which is how it hid for a while.)

## Environment

- nub: reproduced on **0.2.7**, **0.2.10** (published), and **`main` built from source** (`772d732c`, `make install-dev`) — all fail identically.
- Node: `v26.2.0` (from `.node-version`)
- OS: macOS (Darwin 25.5.0), arm64
- Next.js: `16.1.1`, default bundler (**Turbopack**)
- pnpm: `10.27.0` (declared `packageManager`)
- App: large private Next 16 app in a pnpm workspace (App Router, RSC, Clerk, many client components).

## Symptom (Turbopack, via `nub run build`)

```
▲ Next.js 16.1.1 (Turbopack)
✓ Compiled successfully
  Generating static pages using 9 workers (0/61) ...
Error occurred prerendering page "/_global-error".
TypeError: Cannot read properties of null (reading 'useContext')
Export encountered an error on /_global-error/page: /_global-error, exiting the build.
⨯ Next.js build worker exited with code: 1
```

The failing page varies run-to-run (`/beta`, `/callback/loading`, `/_not-found`, `/_global-error` …). The webpack build path (`next build --webpack`) surfaces a sibling invariant from the same family — `Error [InvariantError]: Expected workUnitAsyncStorage to have a store. This is a bug in Next.js.` — consistent with Next's own module-scoped AsyncLocalStorage also not being shared into the workers.

## Discriminator matrix (all on the identical checkout + `node_modules`)

| Runner | Result |
|---|---|
| `pnpm run build` (what Vercel runs) | ✅ 61/61 pages prerender |
| `./node_modules/.bin/next build` (direct binary) | ✅ passes |
| `nub run build` (nub 0.2.7) | ❌ `useContext` null |
| `nub run build` (nub 0.2.10) | ❌ `useContext` null |
| `nub-dev run build` (`main`, from source) | ❌ `useContext` null |

Same script (`nub run -r --if-present build && next build`), same workers ("9 workers"), same everything — only the process wrapper differs.

## What does NOT reproduce it (minimal repros attempted, all via `nub run build`)

- `create-next-app@latest` (Next **16.2.9**), default template → ✅ passes under `nub run build`
- …+ a `"use client"` component calling `useState` rendered on the prerendered page → ✅ passes
- …+ made it a pnpm **workspace** with a second package and a Turbopack `resolveAlias` mapping a bare import to that package's **TS source** (mirroring our config) → ✅ passes

So it is **not** a universal `nub` + `next build` failure. It requires our app's fuller dependency graph — most likely a dependency that ships/loads its own React (a provider/context library) that ends up **duplicated under nub's resolver inside the build workers**. I wasn't able to minimize it further; happy to keep bisecting with guidance.

## Ruled out

- **Install layout** — fails under `nub run` on *both* nub's `.nub/` store and a pnpm `.pnpm/` store; passes under `pnpm run` / direct-binary on the pnpm store. It tracks the **runner**, not the store.
- **nub version** — 0.2.7, 0.2.10, and `main` all fail.
- **`withWorkflow` next-config wrapper** — removing it does not change the failure.
- **App code / a specific page** — the failing page is nondeterministic and includes Next-internal pages; direct `next build` of the same code passes.

## Hypothesis

nub's `worker_threads` / global `Worker` polyfill (the "additive node-mirroring" work — #256 `e251b3cd`, #244 `00d1ce96`) appears not to propagate module-scoped singletons — React's internal hooks **dispatcher** and Next's `workUnitAsyncStorage` — into the worker threads Next spawns for static generation. In those workers React sees a `null` dispatcher, so the first hook call during prerender throws. Real `node:worker_threads` (plain node / `pnpm run`) behaves correctly, so it's specific to how `nub run`'s runtime provides workers to a child `next build`.

## Impact / workaround

- Workaround: build via `pnpm run build` (or the direct Next binary). Vercel already does this, so deploys are unaffected.
- But `nub run build` being unusable for a real Next 16 app is a significant gap given nub is the recommended runner.

## Offer

Private repo, but I can reproduce on demand and I've already built `nub` from `main` locally — happy to run targeted diagnostics (env dumps, `--inspect` on workers, dependency bisect) or test a patch.

colinhacks:
Auto-loaded `.env*` files injected their `NODE_ENV` key into the child env. For `next build` this leaked `NODE_ENV=development` into the prerender forks — dev React against production chunks → null hooks dispatcher → `useContext` crash.

`load_env_files_raw` now drops the `NODE_ENV` key from any `.env*` file and warns once, matching dotenv/`@next/env`/Vite. An ambient `NODE_ENV` still passes through (shell-wins) and still selects `.env.<NODE_ENV>` files; the explicit `--env-file` flag is untouched.

Tests: a loader unit test plus an e2e differential through the real binary. Docs updated.

Closes #263

### DIFF
--- a/crates/nub-cli/tests/integration.rs
+++ b/crates/nub-cli/tests/integration.rs
@@ -4594,6 +4594,84 @@ fn env_file_flag_reaches_child_and_shell_wins() {
     );
 }
 
+/// #263, end-to-end through the binary — the differential that broke `next build`.
+/// A `.env` FILE must NOT leak its `NODE_ENV` into the spawned child (dotenv/
+/// @next/env/Vite parity), yet an AMBIENT `NODE_ENV` must both pass through
+/// untouched AND still select the `.env.<NODE_ENV>` file. Child ambient is set via
+/// `Command::env`, so the test never mutates its own process env (hermetic).
+#[test]
+fn dotenv_node_env_stripped_but_ambient_selects_env_file() {
+    let work = unique_test_cache();
+    let proj = work.join("proj");
+    std::fs::create_dir_all(&proj).unwrap();
+    std::fs::write(proj.join("package.json"), "{}\n").unwrap();
+    std::fs::write(proj.join(".env"), "NODE_ENV=development\nFROM_BASE=base\n").unwrap();
+    std::fs::write(proj.join(".env.production"), "FROM_PROD=prod\n").unwrap();
+    std::fs::write(
+        proj.join("print.mjs"),
+        "console.log(`NE=[${process.env.NODE_ENV || ''}] BASE=[${process.env.FROM_BASE || ''}] PROD=[${process.env.FROM_PROD || ''}]`);\n",
+    )
+    .unwrap();
+
+    let run = |ambient: Option<&str>| {
+        let mut cmd = Command::new(nub_binary());
+        cmd.arg(proj.join("print.mjs").to_str().unwrap())
+            .current_dir(&proj)
+            .env("XDG_CACHE_HOME", unique_test_cache());
+        match ambient {
+            Some(v) => cmd.env("NODE_ENV", v),
+            None => cmd.env_remove("NODE_ENV"),
+        };
+        let out = cmd.output().expect("failed to spawn nub");
+        (
+            String::from_utf8_lossy(&out.stdout).to_string(),
+            String::from_utf8_lossy(&out.stderr).to_string(),
+            out.status.code().unwrap_or(-1),
+        )
+    };
+
+    // (a) ambient NODE_ENV unset: the `.env` NODE_ENV is dropped (not "development"),
+    //     so `.env` — not `.env.production` — is selected, and nub warns once.
+    let (out_a, err_a, code_a) = run(None);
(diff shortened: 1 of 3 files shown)


colinhacks:

PR #268, opened 2026-07-01
fix(node): fall back to %LOCALAPPDATA% for the cache dir on Windows

colinhacks:
Under a Windows service/SYSTEM account (e.g. sshd), `home_dir()` can resolve to a system dir (`C:\WINDOWS`, the SYSTEM profile) or fail, so `cache_dir()` returned `None` and `nub node` / `nub pm shim` failed with "could not locate nub's cache directory".

On Windows, fall back to `%LOCALAPPDATA%\nub` when home is a system path or unresolvable. A normal user profile keeps `~/.cache/nub`; `XDG_CACHE_HOME` still wins. Unix unchanged. Found by the internal Windows papercut survey.

Tests: pure helper unit-tested off Windows; the live path runs on the windows-latest CI leg.


colinhacks:

PR #269, opened 2026-07-01
fix(pm): strip UTF-8 BOM before parsing package.json

colinhacks:
A `package.json` with a leading UTF-8 BOM (`EF BB BF`) is valid to npm/pnpm but broke nub. PowerShell 5.1 / .NET `Encoding.UTF8` and many Windows editors write one by default.

The pre-flight check rejected it (`ERR_NUB_MANIFEST_PARSE`), failing `nub run`/`add`/`install`/`ci`; other nub-side reads silently skipped it. The vendored aube engine already strips the BOM, so only nub-side reads were affected.

Adds a shared `strip_utf8_bom` in nub-core, routing every nub-side parse through it; non-BOM input is unchanged. A cross-platform fix, found by the Windows papercut survey.

Tests: strip-seam unit, pre-flight accepting a BOM'd manifest, e2e `nub run` on a BOM fixture.


colinhacks:

PR #271, opened 2026-07-01
fix(spawn): inject preload --require via NODE_OPTIONS only on the file-run path

colinhacks:
On the `nub <file>` direct-spawn path, nub injected its preload `--require <path>` (and any Yarn PnP `--require <.pnp.cjs>`) into BOTH `process.execArgv` and `NODE_OPTIONS`. A child that rebuilds its Node flags by merging the two channels — Next `next build` fork workers (`getParsedNodeOptions()` → `formatNodeOptions()`), jest-worker — collected the same preload path from both and space-joined the duplicate into a single quoted `--require "a b"`. The fork then died:

```
Error: Cannot find module '/…/runtime/preload.cjs /…/runtime/preload.cjs'
Require stack:
- internal/preload
```

This hit any fork-spawning tool run via `nub <file>`, not just Next.

## Fix

Route the value-bearing preload/PnP `--require`/`--import` flags through `NODE_OPTIONS` only. The direct child inherits `NODE_OPTIONS`, so it stays fully augmented; a child that merges both channels now sees the preload once. This matches the script-runner path (`compute_augmentation_env`), which was already `NODE_OPTIONS`-only.

`NODE_OPTIONS --require` preserves the R1 synchronous-entry semantics (`require.main.id === '.'`, `module.parent === null`, top-level `executionAsyncId === 1`, synchronous throw origin) identically to argv — R1 is the `--require` vs `--import` tier choice, not the argv-vs-`NODE_OPTIONS` channel. Boolean argv flags stay on argv (idempotent when merged twice); only the path-carrying flags, which double destructively, move single-channel.

## Verification

- A fork-reconstructing child (execArgv + `NODE_OPTIONS` merged, quote-joined, forked) collects the preload exactly once and runs, where it previously errored on the doubled path.
- Non-erasable `enum` still transpiles in the direct child and in worker threads — augmentation intact via the inherited `NODE_OPTIONS`.
- R1 synchronous-entry semantics unchanged.
- New regression test `fork_reconstructing_child_gets_a_single_preload_require` + fixtures under `tests/fixtures/nested-spawn/`.

Refs #263 (found during that investigation; distinct from its NODE_ENV cause).

### DIFF
--- a/crates/nub-cli/tests/integration.rs
+++ b/crates/nub-cli/tests/integration.rs
@@ -1296,28 +1296,30 @@ fn fork_ts_with_ipc() {
 }
 
 #[test]
-fn fork_reconstructing_child_gets_a_single_preload_require() {
-    // Dual-channel doubling regression: on the `nub <file>` path the preload
-    // `--require` must ride NODE_OPTIONS ONLY, never also argv. A tool that
-    // rebuilds a fork's flags by MERGING process.execArgv + NODE_OPTIONS (Next
-    // `next build`, jest-worker) would otherwise collect the preload path TWICE
-    // and space-join it into one quoted `--require "a b"`, killing the fork with
-    // `Cannot find module 'a b'`. The fixture reproduces that reconstruction; the
-    // count must be 1 (not doubled), and the child must still transpile its enum
-    // (proving the single-channel preload still augments).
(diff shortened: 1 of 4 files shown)


colinhacks:

PR #272, opened 2026-07-01
pm: rename nub's own lockfile to nub.lock

colinhacks:
Renames the lockfile nub writes for a nub-identity project from `package.lock` to `nub.lock`. The `package.lock` name landed on `main` via #257 but was never released — the released name through v0.2.10 is still `lock.yaml` — so this supersedes #257 before either name ships.

The change is one authoritative line (`NUB_LOCKFILE` in `crates/nub-cli/src/pm_engine/use_align.rs`); the rest is a mechanical sweep of doc-comments, test string-literals, and the install docs page. Bytes are unchanged (pnpm-lock v9).

The legacy reader is preserved: `NUB_LEGACY_LOCKFILE = "lock.yaml"` stays, so an existing `lock.yaml` still migrates byte-identically to the new name on the next mutating install. `vendor/aube` is untouched — the basename is embedder-driven.

Verified: fmt, `clippy --all-targets --all-features -D warnings`, and the pm/lockfile scoped test files pass. Two e2e fixtures confirm a fresh nub-identity project writes `nub.lock` and a legacy `lock.yaml` migrates byte-identically to `nub.lock`.

### DIFF
--- a/crates/nub-cli/src/cli.rs
+++ b/crates/nub-cli/src/cli.rs
@@ -6469,7 +6469,7 @@ fn run_pm_use(name: &str, spec: &str, cwd: &Path) -> Result<i32> {
                 );
             }
         }
-        // nub → pnpm: package.lock renamed back, byte-identical (the two-mode
+        // nub → pnpm: nub.lock renamed back, byte-identical (the two-mode
         // eject — the format was never forked, so the rename IS the eject).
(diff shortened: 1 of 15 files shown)


colinhacks:
Superseded: reverting the package.lock rename first; nub.lock will be re-opened as a held PR on the reverted base.

PR #274, opened 2026-07-01
pm: rename nub's own lockfile to nub.lock

colinhacks:
Renames nub's own lockfile (nub-identity projects) from `lock.yaml` to `nub.lock`. Bytes unchanged (pnpm-lock v9). An existing `lock.yaml` keeps working and migrates automatically the next time the lockfile actually changes.

Re-introduces the `aube-lockfile` filename decoupling so a non-`.yaml` basename can't corrupt a pnpm project's `pnpm-lock.yaml` to `pnpm-lock.lock`. Default-preserving for standalone aube. Supersedes #257's `package.lock` (reverted in #273).

## Migration model (rename-on-real-write — no-churn stays sacred)

The migration rides a REAL lockfile write, it is not a proactive rename:

- A routine install with nothing to change writes nothing and leaves `lock.yaml` exactly as-is (no `nub.lock` created).
- The first real change (dep add/remove/update, or any re-resolve that alters the graph) writes `nub.lock` and removes `lock.yaml` in the same diff — never two files.
- `nub ci` and `--frozen` never write the lockfile, so they never migrate; they read `lock.yaml` fine via the legacy candidate.

Implemented in the aube writer (`write_lockfile_as`) with a legacy-aware no-churn guard; `add --no-save` snapshots/restores the legacy file so a `--no-save` add still restores the tree exactly. Gated on the canonical (Aube) kind + the embedder's legacy-basename list, which is empty for standalone aube — so standalone aube is byte-for-byte unaffected.

**Held — do not merge; the maintainer triggers it.**

Tests: `nub-cli` `lockfile_rename` covers fresh→`nub.lock`, no-op-leaves-legacy, real-change-migrates, and the frozen/`ci`/CI-env read-only paths; `aube-lockfile` `package_lock_rename` covers the writer-level migrate/skip/prune and a pnpm project keeping `pnpm-lock.yaml`.

## Release note

> nub's own lockfile is now `nub.lock` (was `lock.yaml`). An existing `lock.yaml` keeps working — nub reads it — and it migrates automatically the next time the lockfile actually changes: on a dependency add, remove, or update, nub writes `nub.lock` and removes the old `lock.yaml` in the same diff. A routine install with nothing to change never rewrites or renames the lockfile, and `nub ci` never touches a checked-in one.

### DIFF
--- a/crates/nub-cli/src/cli.rs
+++ b/crates/nub-cli/src/cli.rs
@@ -6839,7 +6839,7 @@ fn run_pm_use(name: &str, spec: &str, cwd: &Path) -> Result<i32> {
                 );
             }
         }
-        // nub → pnpm: lock.yaml renamed back, byte-identical (the two-mode
+        // nub → pnpm: nub.lock renamed back, byte-identical (the two-mode
         // eject — the format was never forked, so the rename IS the eject).
         AlignPlan::Rename { from, remove } => {
             let to = root.join(use_align::lockfile_name(name));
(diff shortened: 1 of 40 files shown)


colinhacks:

PR #275, opened 2026-07-01
feat(nubx): interactive dlx consent prompt + exec.implicit-dlx kill-switch

colinhacks:
Replaces the single-line `[y/N]` dlx consent prompt with an interactive select on TTYs, and adds a global kill-switch for nubx's implicit registry tier.

## Prompt

On an interactive terminal, a local miss into the registry now shows a radio instead of a `[y/N]` line:

```
nubx: cowsay is not installed locally.
install and run it from the remote registry?
  ○ Yes
  ● No
  ○ Never (don't ask me again)
```

Up/Down move the highlight, Enter selects, and `y`/`n` are immediate single-key hotkeys. Esc and Ctrl-C abort as No. The default-highlighted row is No, so an accidental Enter is the safe choice. Rendered to stderr.

Non-TTY and CI still fail closed; a terminal `console` can't drive in raw mode falls back to the old `y/N` line reader. `-y`/`--yes` and explicit `nub dlx` bypass the prompt, unchanged.

## Kill-switch: `exec.implicit-dlx`

Picking `Never` writes `exec.implicit-dlx = "never"` to a new nub-owned global config file `~/.config/nub/nub.toml` (`$XDG_CONFIG_HOME/nub`, `%APPDATA%\nub` on Windows):

```toml
[exec]
implicit-dlx = "never"   # prompt | never
```

The setting lives under `[exec]` because dlx is *download and exec* — a fetch-then-exec variant of local-binary execution, the same behavior class.

In `never` mode, nubx still walks its local tiers (file, script, bin); only a full local miss — which would fall through to the registry — aborts, with no prompt and no network:

```
no matching script or executable found.
to run a package from the remote registry, try `nub dlx`
```

`gate()` reads the setting first, before any prompt, prefetch, or network. The file is read/written comment-preservingly via `toml_edit` and atomically via `aube_util`, so a hand-authored file's comments and unrelated keys survive a write.

`nub config get/set exec.implicit-dlx <prompt|never>` reads and writes it; `nub config delete exec.implicit-dlx` clears it back to the default. The key is intercepted ahead of the existing `config` verb's `.npmrc` routing; every other key stays on the `.npmrc` path.

## Notes

- `config_dir()` added to `nub-core` discovery, mirroring `cache_dir()`.
- `console` and `toml_edit` were already compiled in transitively (via `demand` and `toml`); both are promoted to direct deps with no new crate builds.

## Out of scope (fast-follows)

- Eager script-free prefetch during the prompt (needs an aube engine stage-without-exec split).
- A per-spec deny ledger.
- An `allow` (auto-consent) value — the enum reserves it, unimplemented.


colinhacks:

PR #276, opened 2026-07-01
feat(pm): add `nub pm pin` to lock a project to an exact nub version

colinhacks:
Adds `nub pm pin [<version>]` — the dedicated gesture that writes the locked identity pin `packageManager: "nub@<exact>"` (defaulting to the running nub), plus the `devEngines.packageManager` caret beside it.

## What it does

`nub pm pin` writes only the two identity fields and nothing else — no lockfile conversion, no `pnpm-workspace.yaml`/settings migration, and it leaves a project's `pnpm.*` config untouched. That heavier switch onto nub stays `nub pm use nub`'s job. It is the package-manager analog of `nub node pin`.

- Bare `nub pm pin` (or `nub pm pin nub`) locks the running nub; `nub pm pin <version>` locks that exact release. A leading `nub@` is stripped for forgiveness.
- The version must be exact — nub is the running binary, not a registry package, so a range or dist-tag (`^1`, `1.x`, `next`, `latest`) has nothing to resolve against and is refused.
- No network: the pin is a local declaration.

## Implementation notes

- Extracted the identity-field writer (`write_nub_identity_fields`) out of `apply_manifest_edits`, so `pin` reuses the tested write path without the full-migration side effects — `apply_manifest_edits` also strips the `pnpm.*` namespace, which for a lightweight pin would be data loss. The extraction is behavior-preserving for the existing `pm use nub` path.
- Split the stale `"pin" | "switch"` bail arm: `pin` now has a real handler; `switch` still names its successor.
- Docs (`pm.mdx`), help text, and the unknown-verb set updated. AGENTS.md gains one line noting the `nub pm <verb>` namespace is nub's own management surface, not a pnpm mirror.

## Verification

- `cargo clippy --all-targets --all-features -- -D warnings` and `cargo fmt --check`: clean.
- New integration test (`pm_two_mode.rs`) plus the edited inline tests pass; `apply_manifest_edits`'s existing tests are unchanged.
- e2e against a tmp fixture: bare pin, explicit version, `nub@` prefix, range/tag rejection, and idempotence confirmed.

Refs #265 — `nub pm pin` is the writer of the locked pin that #265's self-shim reads.


colinhacks:

PR #278, opened 2026-07-01
feat(run): verify dependency freshness before run/exec/file/nubx (#252)

### LINKED ISSUE #252: Auto-install on `nub run` (aube-style)
Currently running `nub run <script>` without previous `nub install` ends with all sorts of errors. 

Since "never forget to install" is aube's leading seeling point, it might be something for nub to eventually include.

```
❯ rm -rf node_modules
❯ nub run prepare
$ husky
sh: husky: command not found
```

Whereas with `aube`:
```
❯ aube run prepare
Auto-installing: install state not found
aube 1.25.1 by jdx.dev
██████████████░ 2034/2040 pkgs · 263.3 MB / ~263.7 MB · 11.9 MB/s · ETA 65ms
██████████████░ 2040/2040 pkgs · 276.3 MB · 11.4 MB/s
███████████████ 2040/2040 pkgs
✓ resolved 2040 · reused 0 · downloaded 2040 (276.3 MB) in 38.0s
...
```

I'm doing most of my work in Go and there is basically no need to run `go mod download` before running `go run ./cmd/main.go` or `go tool ...`. Go binary will automatically download required modules (albeit, precisely the ones required to build, not all of them). This becomes a muscle memory. 
Of course there are cases where pre-download is still more optimal, but that doesn't mean we can't auto-install, too 😉

colinhacks:
Closes #252.

`nub run <script>` (and `nub <file>`, `nub exec`, `nubx`) now checks the installed `node_modules` against `package.json` before running, so a missing or drifted dependency surfaces as a clear message instead of a raw `husky: command not found`:

```
nub: dependencies may be out of date (`chalk` is not installed). Run `nub install`.
```

## Behavior

A single marker-free walk of the manifest's direct dependencies against the `node_modules` chain — no lockfile is parsed, so it's immune to cross-PM lockfile churn, and it works whatever installed the tree (npm, pnpm, yarn-classic, bun, or nub). It errs toward silence: Yarn PnP, a spec that isn't a semver range, a prerelease install, or a production-only install with devDependencies omitted never warns.

- Default is `warn` (warn, then run). Set `verify-deps-before-run` in `.npmrc` to `error` (refuse to run) or `off`. `NUB_VERIFY_DEPS_BEFORE_RUN` overrides per shell; `--no-check` (or `--no-install` on `nub run`) skips per invocation.
- Policy is resolved by nub's own path, so `vendor/aube` is unchanged and standalone aube's default is untouched.
- The check fires once per command: a process latch, the `npm_lifecycle_event` re-entry guard, and an inherited internal marker keep a program that spawns `node` from repeating the warning.
- nub does not auto-install — `install` is accepted as a value but treated as `warn`, since nub won't reshape a tree another package manager set up.

## Tests

Unit tests (policy parsing, range satisfaction, dep-map) plus 12 integration tests: fresh-clone warn, warm silent, version drift, production-install safety, PnP skip, `--no-check`, compat-mode skip, `.npmrc` off/error, env override, lifecycle re-entry, inherited marker.

Docs: `site/content/docs/runner/run.mdx`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2622,6 +2622,7 @@ dependencies = [
  "libc",
  "miette",
  "mimalloc",
+ "node-semver",
  "nub-core",
  "num_cpus",
  "rayon",
--- a/crates/nub-cli/Cargo.toml
+++ b/crates/nub-cli/Cargo.toml
@@ -78,6 +78,11 @@ serde = { workspace = true }
 serde_json = { workspace = true, features = ["preserve_order"] }
 rustc-hash.workspace = true
 semver = "1"
+# npm-semantics range satisfaction for the pre-run dependency-freshness gate
+# (verify_deps.rs). The Cargo `semver` crate above uses Cargo semantics, which
+# don't match npm ranges (`x`-ranges, `||` unions, hyphen ranges); node-semver
+# is the same resolver aube already uses.
+node-semver = "2"
 # pnpm-workspace.yaml read + regeneration for `nub pm use nub` / `use pnpm`
 # (pm_engine/use_nub.rs). serde_yaml is archived/unmaintained; serde_yaml_ng
 # is the maintained, API-compatible fork, aliased as `serde_yaml` so the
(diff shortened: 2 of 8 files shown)


colinhacks:

PR #279, opened 2026-07-02
fix(pm): auto-persist minimumReleaseAgeExclude on loose-mode fallback (#262)

### LINKED ISSUE #262: Package manager does not abide by PNPM minReleaseAge
After updating some packages with `nub up -i` (interactive update), it successfully completed, but I got this:

```
[ERR_PNPM_MINIMUM_RELEASE_AGE_VIOLATION] 3 lockfile entries failed verification:
  caniuse-lite@1.0.30001800 was published at 2026-06-30T09:58:35.000Z, within the minimumReleaseAge cutoff (2026-06-30T02:52:43.280Z)
  electron-to-chromium@1.5.383 was published at 2026-06-30T22:02:45.000Z, within the minimumReleaseAge cutoff (2026-06-30T02:52:43.280Z)
  vite@8.1.2 was published at 2026-06-30T16:46:36.000Z, within the minimumReleaseAge cutoff (2026-06-30T02:52:43.280Z)
```

I tried both `pnpm dev` and `nub run dev` after the update. Both gave this error.

colinhacks:
## What

When `minimumReleaseAge` is set in loose mode (the default), the resolver installs the lowest satisfying version even when it is younger than the cutoff — pnpm's documented fallback. pnpm co-writes that package to `minimumReleaseAgeExclude` so its own install-time lockfile verifier accepts the pin. nub did the same fallback pick but never recorded the exclude, so a `nub install`/`add`/`update` produced a lockfile pnpm later rejected with `ERR_PNPM_MINIMUM_RELEASE_AGE_VIOLATION` — surfaced when a project's scripts shell out to pnpm, which is exactly the report in #262.

This mirrors pnpm's loose-mode behavior: after a successful install/add/update/dedupe, nub records each immature package in `minimumReleaseAgeExclude` and prints a factual notice. `minimumReleaseAgeStrict=true` is unchanged — it still hard-aborts.

## How

- The resolver surfaces each immature fallback pick through a small process-global sink in `aube-util` (`age_gate.rs`). It is inert unless armed (one relaxed atomic on the record path), so standalone aube is unchanged.
- nub arms the sink before an install/add/update/dedupe engine run and, on success, drains it and records the entries.

Write target — `minimumReleaseAgeExclude` is a pnpm *workspace-config-file* setting, so pnpm reads it only from `pnpm-workspace.yaml`, never from `.npmrc` (verified against pnpm 11's config reader; pnpm 10.15.1 has no `minimumReleaseAge` feature at all):

- A pnpm-compat project (an existing `pnpm-workspace.yaml`, or a `pnpm-lock.yaml` nub round-trips) → the `pnpm-workspace.yaml` sequence, created when absent. This is where pnpm reads the setting and where pnpm's own auto-persist writes it. Gated on the incumbent posture, so a nub-identity / npm / yarn / bun / fresh project never gets one created.
- Otherwise (nub identity, npm/yarn/bun compat, fresh) → the project `.npmrc` — the neutral surface nub itself reads `minimumReleaseAge*` from, so nub reads back exactly what it wrote. Nothing re-reads these projects' lockfiles as pnpm, so pnpm's `.npmrc` blind spot is moot.

Entries use pnpm's canonical per-package `name@v1 || v2` form (a port of pnpm's `mergePackageVersionSpecs`), merged with what's already recorded. A companion `yaml_patch` change re-emits a changed block sequence through the manual injector, since yamlpatch's `Op::Replace` mishandles a block sequence the same way it already mishandles a block mapping.

## Verification

- Unit tests for the sink, the yaml read/replace helpers, the npmrc merge, the routing, and the canonical merge; existing suites green (aube-resolver 257, aube-manifest 122).
- End-to-end (`nub install`/`add` with a `minimumReleaseAge` that ages out every version): a nub-identity/fresh project records to `.npmrc`; a declared-pnpm project with no lockfile has both `pnpm-lock.yaml` and a new `pnpm-workspace.yaml` written, carrying `minimumReleaseAgeExclude: [is-number@7.0.0]`, with nothing in `.npmrc`; a second immature version of the same package collapses to the canonical `name@v1 || v2` form across runs; re-running does not duplicate; strict mode still aborts with nothing written.

pnpm 11's acceptance of the created `pnpm-workspace.yaml` was validated from its config-reader source (the local pnpm is 10.15.1, which predates the feature).

Two known non-blocking edges: the write targets the invocation cwd rather than the resolved workspace root (consistent with the existing `devEngines` virgin-stamp), and the pnpm-incumbent signal keys on `pnpm-lock.yaml` presence.

Closes #262

### DIFF
--- a/crates/nub-cli/src/pm_engine/install_family.rs
+++ b/crates/nub-cli/src/pm_engine/install_family.rs
@@ -360,11 +360,13 @@ fn run_add(typed: &str, args: &[String]) -> Result<i32> {
             &yarn_remedy("add", &verb.packages),
         ));
     }
+    super::min_release_age::arm();
     let code = finish_quieted(
         &globals.output,
         &session,
         aube::commands::add::run(verb, globals.effective_filter()),
     )?;
+    super::min_release_age::persist(&session.cwd, code == 0, &globals.output);
     stamp_if_virgin(&session, code);
     Ok(code)
 }
@@ -408,11 +410,13 @@ fn run_update(typed: &str, args: &[String]) -> Result<i32> {
             &yarn_remedy("upgrade", &verb.packages),
         ));
     }
+    super::min_release_age::arm();
     let code = finish_code_quieted(
         &globals.output,
         &session,
         aube::commands::update::run(verb, globals.effective_filter()),
     )?;
+    super::min_release_age::persist(&session.cwd, code == 0, &globals.output);
     stamp_if_virgin(&session, code);
     Ok(code)
 }
@@ -1056,7 +1060,9 @@ pub fn run_install(flags: InstallFlags) -> Result<i32> {
         opts.mode = FrozenMode::Frozen;
     }
 
+    super::min_release_age::arm();
     let code = run_engine(&session, opts, yarn, &flags.output)?;
+    super::min_release_age::persist(&session.cwd, code == 0, &flags.output);
     // Virgin install only: stamp a caret RANGE into `devEngines.packageManager`
     // so the project advertises nub the standard, cross-tool way WITHOUT locking
     // itself to one exact nub version. nub's canonical lockfile is deliberately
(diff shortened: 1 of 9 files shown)


colinhacks:

PR #282, opened 2026-07-02
fix(run): keep the project node_modules/.bin on the script PATH even when it does not exist yet

### LINKED ISSUE #281: nub run: node_modules/.bin missing from script PATH when the directory does not exist at spawn
`nub run` adds `node_modules/.bin` to the script's `PATH` only when the directory already exists at spawn time. npm and pnpm add the project-level entry unconditionally. A script that creates the directory mid-run — the install-then-run pattern — fails under nub with `command not found` while working under npm and pnpm.

## Steps to reproduce

```bash
mkdir repro && cd repro
cat > package.json <<'EOF'
{
  "name": "repro",
  "version": "1.0.0",
  "scripts": {
    "gen": "npm install --no-save --loglevel=error cowsay && cowsay it-works"
  }
}
EOF
nub run gen
```

Output:

```
added 41 packages in 4s
sh: 1: cowsay: not found
```

The same script prints the cow under `npm run gen` and `pnpm run gen` (fresh `node_modules` each time).

## Cause

`bin_path()` in `crates/nub-core/src/workspace/scripts.rs` gates every `PATH` entry on `is_dir()`. pnpm prepends the project's `node_modules/.bin` whether or not it exists yet; npm prepends the working directory's and every ancestor's unconditionally. Since the `PATH` is computed before the script runs, a `.bin` directory created during the script is unreachable under nub.

Versions: nub 0.2.10, npm 11.12.1, pnpm 11.7.0, Linux.

dreygur:
Closes #281

`bin_path()` gated every `node_modules/.bin` PATH entry on `is_dir()`, computed before the script spawns. A script that creates the directory mid-run (`npm install --no-save cowsay && cowsay hi`) failed with `command not found` under nub while working under npm and pnpm, which add the project-level entry unconditionally.

The fix adds the project's own `.bin` to the PATH whether or not it exists yet; walked-up ancestor entries stay existence-gated.

Verified: new unit test (`bin_path_includes_project_bin_before_it_exists`), plus e2e on the issue's repro fixture — mid-script-created bin resolves (rc 0, was 127), the realistic install-then-run script prints its output, and the existing ancestor-`.bin` resolution is unchanged. `cargo clippy --all-targets --all-features -- -D warnings` and `cargo fmt --check` are clean.

### DIFF
--- a/crates/nub-core/src/workspace/scripts.rs
+++ b/crates/nub-core/src/workspace/scripts.rs
@@ -303,9 +303,13 @@ pub fn bin_path(project_root: &Path, workspace_root: Option<&Path>) -> String {
 
     // Walk from project root up, adding each node_modules/.bin.
     let mut dir = project_root.to_path_buf();
-    for _ in 0..16 {
+    for depth in 0..16 {
         let bin_dir = dir.join("node_modules").join(".bin");
-        if bin_dir.is_dir() {
+        // The project's own .bin goes on PATH even when it doesn't exist yet —
+        // npm and pnpm add it unconditionally, so an install-then-run script
+        // (`npm ci && tool`) can invoke a bin created mid-script (#281).
+        // Ancestor entries stay existence-gated.
+        if depth == 0 || bin_dir.is_dir() {
             dirs.push(bin_dir.to_string_lossy().to_string());
         }
         if workspace_root.is_some() && Some(dir.as_path()) == workspace_root {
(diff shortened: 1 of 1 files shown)


colinhacks:
Thanks for the quick fix and for reporting this in #281. We're landing a broader version in #283 that also covers ancestor `.bin` directories in workspaces (not just the project's own), matching npm's unconditional walk-up behavior. Closing this in favor of that one — appreciate you tracking it down.

PR #283, opened 2026-07-02
fix(run): prepend node_modules/.bin to script PATH unconditionally

### LINKED ISSUE #281: nub run: node_modules/.bin missing from script PATH when the directory does not exist at spawn

colinhacks:
`bin_path()` gated each ancestor's `node_modules/.bin` PATH segment on an `is_dir()` check computed before spawn, so a `.bin` absent at spawn was dropped — including one a script creates mid-run (install-then-invoke). npm and pnpm prepend unconditionally without stat'ing; a missing PATH entry is inert. Drop the check; keep the 16-level cap and `workspace_root` break. `find_bin()` untouched.

Tests: `bin_path_includes_missing_dir` (unit) and `run_script_finds_bin_created_mid_run` (Unix e2e).

Closes #281

### DIFF
--- a/crates/nub-cli/tests/workspace_run.rs
+++ b/crates/nub-cli/tests/workspace_run.rs
@@ -266,3 +266,31 @@ fn filtered_remove_keeps_a_workspace_dep_resolvable() {
         "the lockfile must be updated in lockstep, not left stale"
     );
 }
+
+/// Regression for #281: a script that creates `node_modules/.bin/<tool>` and
+/// then invokes it by bare name in the same run must succeed — the install-
+/// then-invoke pattern. nub used to drop `node_modules/.bin` from the child's
+/// PATH when the directory did not exist at spawn time; npm/pnpm prepend it
+/// unconditionally. Unix-only: the fixture writes a `#!/bin/sh` shim, which is
+/// runnable on Unix but not the `.cmd`-shim shape Windows resolves.
+#[cfg(unix)]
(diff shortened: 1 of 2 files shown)


colinhacks:

PR #284, opened 2026-07-02
feat(runtime): Import Text — `import x from "./f" with { type: "text" }` on any extension

colinhacks:
nub's own Import Text: `import s from "./file" with { type: "text" }` yields the raw file contents as a default-export string, on **any** extension. This is a new nub augmentation via the existing load-hook surface — it converges on the semantics Node standardized upstream (the `--experimental-import-text` `textStrategy` translator) and that Bun and Deno ship, not a port of patched Node internals. A plain-Node user registering the same `module.registerHooks` / `module.register` hook gets the identical result, so the augmenter-not-fork contract holds (verified with plain-Node prototypes across the Node matrix).

## What it fixes

`type: "text"` on a non-`.txt` extension previously errored:

```
import notes from "./README.md" with { type: "text" };
// before: TypeError [ERR_UNKNOWN_FILE_EXTENSION]: Unknown file extension ".md"
// after:  the file contents, as a string
```

The attribute is checked ahead of extension dispatch, so it wins over nub's extension-based data loaders (a `.yaml` read as text is the raw source, not the parsed object) and over Node-native JSON (a `.json` read as text is the raw string).

## Mechanism

Both tier load hooks — fast (`runtime/preload-common.cjs`, sync `module.registerHooks`) and compat (`runtime/preload-async-hooks.mjs`, async `module.register` loader worker) — check `context.importAttributes?.type === "text"` and return a synthesized `export default <string>` via the shared `loadTextImport` helper in `runtime/transform-core.mjs`. It `shortCircuit`s, so Node never runs its own attribute validation. Semantics match Node's `textStrategy`: default export only (a named import is a load-time `SyntaxError`), decoded via `TextDecoder` (UTF-8, strips a leading BOM). ESM-only — Node has no CJS `require`-with-attributes API. The extension-based `.txt`/`.yaml`/… loaders are unchanged; this is additive.

## Node version support — and one deferred decision

Works on every supported Node that can parse the `with { … }` import-attribute syntax: **Node 18.20 → latest, both the fast and compat tiers** (verified on 18.20.4 / 20.19 / 22.13 / 22.14 / 22.15 / 24.14 / 26.2).

The `with { … }` syntax landed in V8 12.3 (Node 22), backported to Node 18.20 and 20.10. **Node 18.19.0 / 18.19.1 — the two patch releases at nub's support floor — cannot parse it in V8 at all** (the import fails before any hook runs). Closing that gap is feasible via a source-level `with` → `assert` keyword rewrite (18.19 parses `assert { type: "text" }` and surfaces the attribute to the load hook identically — proven), but emitting `assert` needs a native transpile-path change (oxc emits `with` verbatim and does not lower import attributes) plus routing plain-JS-with-attributes through the transform on the floor. That is a hot-path change for exactly two patch releases where a bump to Node 18.20 (same LTS major) works natively. **Left as a maintainer decision** — native floor rewrite, a bump-to-18.20 diagnostic on 18.19.x, or accept the gap — rather than shipped here.

## Tests

- `import_text_attribute_any_extension`, `import_text_named_import_is_a_load_error` (`crates/nub-cli/tests/integration.rs`).
- `import_text_works_on_compat_tier` (`crates/nub-cli/tests/version_tiers.rs`, Node 22.13.0) so the async loader-worker path is not masked by a modern host Node.
- Fixtures under `tests/fixtures/import-text/` and `tests/fixtures/import-text-named/`.

`cargo fmt --check` and `cargo clippy --all-targets --all-features -- -D warnings` are clean.

### DIFF
--- a/crates/nub-cli/tests/integration.rs
+++ b/crates/nub-cli/tests/integration.rs
@@ -3287,6 +3287,50 @@ fn data_named_import_is_a_load_error() {
 }
 
 #[test]
+fn import_text_attribute_any_extension() {
+    // nub's Import Text: `import s from "./f" with { type: "text" }` yields the raw
+    // file contents as a default-export string, on ANY extension. The attribute is
+    // checked ahead of extension dispatch, so it wins over both nub's data loaders
+    // (a `.yaml` read as text is NOT parsed) and Node-native JSON (a `.json` read
+    // as text is the raw string). Matches Node's upstream `textStrategy` semantics.
+    let (stdout, stderr, code) = run_nub("import-text", "main.ts");
+    assert_eq!(code, 0, "stderr: {stderr}\nstdout: {stdout}");
+    assert!(
+        stdout.contains(r##"md:"# Release notes\n\n- first\n- second\n""##),
+        "text of a non-data extension (.md) returned verbatim as a string: {stdout}"
+    );
+    assert!(
+        stdout.contains("yaml-is-string:true") && stdout.contains("yaml-unparsed:true"),
+        "a .yaml imported with type:text is the RAW text, not the parsed object: {stdout}"
+    );
+    assert!(
+        stdout.contains("json-is-string:true")
+            && stdout.contains(r#"json-unparsed:"{\"name\":\"demo\",\"version\":2}\n""#),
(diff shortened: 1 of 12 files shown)


colinhacks:

PR #285, opened 2026-07-02
Import Text on the Node 18.19.x floor (with→assert keyword rewrite)

colinhacks:
Refs #284.

#284 shipped Import Text (`import x from "./f" with { type: "text" }`) for Node 18.20 through latest. It fails on Node 18.19.0 / 18.19.1 — two patch releases inside nub's documented 18.19 support floor — because their V8 cannot parse the `with` import-attribute keyword at all (it landed in V8 12.3 / Node 22, backported to 18.20 / 20.10, never 18.19.x). The parse fails before any load hook runs.

## Fix

On those two floor patches only, rewrite the `with { … }` keyword to the older `assert { … }` spelling before Node's parser sees the source. On 18.19.x, `assert` parses and surfaces the identical `importAttributes`, so the existing `loadTextImport` is unchanged; native JSON attribute imports (`with { type: "json" }`) work down to the floor too, since the syntax — not the attribute — is the wall. `assert` is removed on Node 22+, so the rewrite is gated to 18.19.x and is a no-op everywhere else.

## Mechanism

- nub-native `rewrite_import_attributes_keyword`: an oxc-parse-driven byte splice of just the keyword token (a `with` in a string / comment / regexp is untouched). One AST visitor covers `import`, `export … from`, `export * from`.
- transform-core gates on `process.versions.node` and rewrites the final source Node compiles — the oxc output for TS/JSX, and the raw bytes for a plain `.js`/`.mjs` whose only floor-incompatible construct is the attribute clause (a minimal splice, no full transpile).

The dynamic form `import(x, { with: {…} })` stays out of scope (18.19's runtime reads the older `assert:` object key; rewriting arbitrary options expressions is unsafe).

## Verification

Real 18.19.0 and 18.19.1: entry `.mjs`, `.ts`, imported module, non-`.txt` extension, string-literal safety, generic JSON all pass; no regression on 18.20.4 / 22.14 / 22.15; cross-version transpile-cache reads correct. nub-native unit tests (8) plus a `version_tiers` floor test (18.19.0, side-installed in CI). `clippy --all-targets --all-features -D warnings` and `fmt --check` clean.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -641,11 +641,18 @@ jobs:
           }
           dir_22_13=$(install_node 22.13.0)
           dir_18_18=$(install_node 18.18.0)
+          # 18.19.0: exact floor patch whose V8 cannot parse the `with {…}`
+          # import-attribute keyword — pins the with->assert rewrite (import-text on the
+          # 18.19 floor). `actions/setup-node`'s "18.19" leg resolves to the latest
+          # patch, so the exact .0 is side-installed like the other tier reps.
+          dir_18_19=$(install_node 18.19.0)
           "$dir_22_13/node" --version
           "$dir_18_18/node" --version
+          "$dir_18_19/node" --version
           {
             echo "TEST_NODE_BIN_22_13_0=$dir_22_13"
             echo "TEST_NODE_BIN_18_18_0=$dir_18_18"
+            echo "TEST_NODE_BIN_18_19_0=$dir_18_19"
           } >> "$GITHUB_ENV"
       - run: pnpm install --frozen-lockfile
       # The data-format loader tests need the nub-native addon (the npm yaml/toml/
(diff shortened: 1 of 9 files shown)


colinhacks:
Not pursuing: the 18.19.0/18.19.1 floor-only rewrite adds real complexity (a dedicated oxc-parse-driven source rewrite pass) for two narrow patch releases. Import Text's floor stays 18.20+, as shipped in #284.

colinhacks:
Review findings addressed in `2f2bcde4` + `efbda7be`:

- Floor stderr: the text load hook shortCircuits before Node reads the attribute, so no `ExperimentalWarning` fires (empty stderr on real 18.19.0). Added a clean-stderr assertion.
- `maybeTranspilePlainJs` format: a `.js` under `type: commonjs` wrongly loaded as module on the floor; now gated on the real module format, erroring consistently across 18.19/18.20/22.15. Added a discriminating ESM-parent test.
- Comment between `with` and `{`: not a regression; documented floor-only limitation.

PR #291, opened 2026-07-03
config: move nub's global settings file from nub.toml to nub.jsonc

colinhacks:
Switches nub's sole global-config field from `~/.config/nub/nub.toml` `[exec] implicit-dlx` (TOML) to `~/.config/nub/nub.jsonc` `exec.implicitDlx` (JSONC). The prior `nub.toml` was never released, so this is a clean switch — no migration, no compat read.

- Reads via `jsonc-parser` `parse_to_serde_value` (best-effort: absent/malformed/unknown-key yield the `prompt` default, never failing the nubx consent path); writes via the comment- and trailing-comma-preserving `cst` module. A pre-existing non-object `exec` is replaced rather than panicking. Only `nub.jsonc` is read — `nub.json` is never accepted.
- The `config` verb accepts `exec.implicitDlx` (canonical) and the kebab `exec.implicit-dlx` (read/route alias); it always writes camelCase.
- `toml_edit` is dropped (`config.rs` was its only user); nub-cli gains the `jsonc-parser` `cst` feature. The `toml` crate is unchanged.

Scope is GLOBAL config only; project-level `nub.jsonc` precedence and fail-loud-on-unknown-key remain deferred to the full config-file effort.

Verified: `cargo fmt --check`, `cargo clippy --all-targets --all-features -D warnings`, config unit tests (55), and an end-to-end run of `config set`/`get`/`delete` against a real `~/.config/nub/nub.jsonc` (comment/`$schema`/trailing-comma preservation, kebab alias, malformed→default, non-object-`exec` recovery).

Refs #275

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2230,6 +2230,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "jsonc-parser"
+version = "0.24.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2ebcfcb2f2b016f7372835b99cd3be24cc13ce9823107beb0f98b09c6092ba9f"
+dependencies = [
+ "serde_json",
+]
+
+[[package]]
 name = "junction"
 version = "2.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


colinhacks:

Issue #286, opened 2026-07-02
pm: nub doesn't reproduce pnpm's hidden hoist tree (hoistPattern '*'), so ambient @types/* aren't resolvable → spurious TS errors


AlexGodard:
## Summary

`nub install` / `nub ci` don't reproduce pnpm's **hidden hoist tree** (`node_modules/.pnpm/node_modules/`, populated by the default `hoistPattern: ['*']`). As a result, `@types/*` packages that a consumer relies on ambiently — without declaring them directly — aren't reachable by TypeScript's module-resolution walk, producing typecheck errors that don't exist under pnpm.

> **Correction (thanks @pullfrog):** an earlier version of this issue claimed nub should seed pnpm's default `public-hoist-pattern` of `*types*`/`*eslint*`. That default was **removed in pnpm 7.0.0** ([pnpm/pnpm#4459](https://github.com/pnpm/pnpm/pull/4459)) and has been `[]` ever since — so nub's `publicHoistPattern: []` is correct pnpm-11 parity. The real divergence is the missing hidden hoist tree, described below.

## Impact

On a pnpm 11 monorepo (`packageManager: pnpm@11.5.2`, lockfileVersion 9.0, default isolated linker), packages that declare `react`/`react-dom` but rely on the ambient `@types/react` (e.g. `next-themes`, `@react-pdf/renderer`) resolve React **without its type definitions**. tsgo/tsc then report cascading errors:

```
error TS2786: 'Document' cannot be used as a JSX component.   (@react-pdf/renderer)
error TS2339: Property 'children' does not exist on type 'ThemeProviderProps'.  (next-themes)
error TS7006: Parameter 'e' implicitly has an 'any' type.      (DOM event handlers)
```

Same lockfile + same `tsgo` version is green on CI (installed with pnpm) and red locally (installed with nub) — 86 errors, none of them real.

## Root cause (verified)

- pnpm's default `hoistPattern: ['*']` populates a **hidden hoist tree** at `node_modules/.pnpm/node_modules/`, where `@types/react` is surfaced. TS's resolution walk from `next-themes`/`@react-pdf/renderer` finds it there — no root public-hoist needed.
[...]

PR #293, opened 2026-07-03
fix(pm): make hoist default GVS-aware — hidden hoist tree wherever GVS does not engage

colinhacks:
## Problem

nub pushed an embedder-tier `hoist=false` unconditionally to engage the shared global virtual store (GVS), since aube couples the axes (`effective_gvs = planned && !hoist && Isolated`). Wherever GVS does not actually engage — CI, `nub ci`, a next/nuxt/parcel trigger, an explicit `enableGlobalVirtualStore=false`, dlx — the project still got `hoist=false`, so the pnpm-parity hidden hoist tree (`node_modules/.nub/node_modules/`) was missing for zero GVS benefit. That broke ambient `@types/*` resolution for store-resident packages and made `enableGlobalVirtualStore=false` an incomplete escape hatch.

## Change

New embedder profile flag `gvs_over_default_hoist` (default `false`). Under it, only an EXPLICITLY-set `hoist=true` vetoes GVS; a default hoist lets GVS engage, and the hidden hoist tree is built only where GVS is off. nub sets the flag and stops pushing `hoist=false`; injected deps push an explicit `hoist=true`. The command layer pre-folds the (effective GVS, hidden-tree) decision once and threads it to the link phase, so the linker's own coupling fallback is never reached under the flag (issue #71 lockstep preserved). A codegen'd `hoist_explicit(ctx) -> Option<bool>` accessor distinguishes an explicit hoist from the built-in default.

Isolated linker, no user-set hoist:

| case | GVS | hidden tree |
|---|---|---|
| off-CI, no trigger | on | no (byte-identical to today) |
| CI · `nub ci` · trigger · `enableGlobalVirtualStore=false` · dlx | off | yes |
| explicit `hoist=false` | — | never |
| explicit `hoist=true` | off | yes |

## Fork-discipline

Flag `false` ⇒ standalone aube is byte-for-byte unchanged on its default path: the effective-GVS formula's flag-off branch is the old formula, the codegen companion accessor is strictly additive, the linker is untouched, and the link phase's non-flag path passes the raw override + resolved hoist exactly as before.

## Verification

21/21 e2e truth-table cells pass (real installs, scratch store); aube + nub-cli tests green; clippy/fmt clean. Three fresh-context review passes (correctness, impact-analysis, fork-discipline) found no bugs. Previously-affected projects take one relink on first post-upgrade install (shape-hash flip, self-healing). The plain non-CI→CI install with a stale `node_modules` short-circuiting one hidden-tree build is a pre-existing, uncommon edge that self-heals; `nub ci` is unaffected.

Refs #286

### DIFF
--- a/crates/nub-cli/src/pm_engine/identity.rs
+++ b/crates/nub-cli/src/pm_engine/identity.rs
@@ -182,6 +182,16 @@ pub(crate) const NUB: aube_util::Embedder = aube_util::Embedder {
     // `--loglevel debug` / `RUST_LOG`. The per-project fallback behavior itself
     // is unchanged; only the notice is silenced.
     gvs_incompatible_warning: false,
+    // GVS-precedence hoisting (#286): a DEFAULT hoist no longer vetoes the
+    // shared virtual store, so GVS engages wherever it's active (off-CI, no
+    // trigger, no explicit opt-out) with no hidden tree, and the pnpm-parity
+    // hidden hoist tree is built wherever GVS is OFF (CI, `nub ci`, a
+    // next/nuxt/parcel trigger, an explicit `enableGlobalVirtualStore=false`,
+    // dlx) — restoring ambient `@types/*` resolution for store-resident
+    // packages. Only an EXPLICIT `hoist=true` (nub's injected-deps push, or a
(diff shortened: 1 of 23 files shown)


colinhacks:

Issue #294, opened 2026-07-03
`node` not on PATH with `nub node install 24` on new install without other node installs


fardarter:
Greets!

I'm setting up a machine for the first time and installed nub with `brew`, thinking I might use it as my default node runner. However, it can't seem to find the binary.

```
Error: no Node binary found on PATH
  Install one with: nub node install (or `nub node install <version>` to pick a version)
  (nub augments your Node — it doesn't bundle one; a pin in .node-version / .nvmrc / engines.node is provisioned automatically.)
```

I assume this is a bug, but what is the correct addition to `PATH`? Might be worth adding to the docs.

Looking forward to exploring more.

colinhacks:
Investigating.

fardarter:
Thanks!  

Re the pullfrog comment, delightful behaviour here for me would be having my first install version that is not linked to a nearby version file offered as default by a prompt. If I say no, I can be told how to set one, and I’m still free to operate without a global pinned version, but now I know that pinning is what sets that version the first time.

colinhacks:
Fixed in [#296](https://github.com/nubjs/nub/pull/296): with no pin and no Node on PATH, `nub` now auto-provisions — the highest cached version if the store has one, otherwise the latest release from nodejs.org (announced, no pin written). The error message also no longer suggests the dead-ending bare `nub node install`.

fardarter:
Amazing!

colinhacks:
Shipped in v0.3.0: https://github.com/nubjs/nub/releases/tag/v0.3.0

PR #298, opened 2026-07-03
fix(pm): clear rejection for `nub pm update` on a bun-pinned project

colinhacks:
`nub pm update` on a bun-pinned project printed `no package manager is pinned to update — declare one with nub pm use <pm>` — a dead-end (bun IS declared; bun isn't provisionable, so `nub pm use bun` can't fix it either).

The `update`/`up` arm now detects a declared-but-unprovisionable identity (mirroring the `which` arm) and names bun + its own path (`e.g. bun upgrade`). Excludes nub and yarn Berry, which resolve elsewhere; provisionable pins unaffected. Adds a regression test.

Surfaced by the #294 misleading-message audit.


colinhacks:

PR #299, opened 2026-07-03
cli: rename --shell-emulator to --posix-shell

colinhacks:
Renames the `nub run` flag `--shell-emulator` to `--posix-shell`.

The flag locates a system POSIX `sh` (`PATH` / Git-for-Windows) and errors if none exists — the inverse of the ecosystem's "shell emulator" (pnpm, yarn berry, bun), a bundled JS/native interpreter needing no system `sh`. The shared name was a misnomer; `--posix-shell` names what the flag does. A true bundled emulator is a separate, future question.

The old spelling is not aliased: `--shell-emulator` hard-errors with guidance toward `--posix-shell`, since the semantics were a false friend. Routing is otherwise unchanged. Tests in `crates/nub-cli/tests/posix_shell.rs` cover both paths.

### DIFF
--- a/crates/nub-cli/src/cli.rs
+++ b/crates/nub-cli/src/cli.rs
@@ -100,18 +100,18 @@ fn npm_config_reporter_is_silent() -> bool {
         .unwrap_or(false)
 }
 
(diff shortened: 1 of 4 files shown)


colinhacks:

Issue #288, opened 2026-07-03
nub i stuck on linking step


toplac:
nub i
nub 0.2.10 — fetching   [██████████████████░░]   367/379  pkgs · ETA 22ms
WARN path-expression-matcher@1.5.0: registry response has no `dist.integrity`, importing without content verification. Set `snub 0.2.10 — fetching   [██████████████████░░]   368/379  pkgs · 36 kB · 42 kB/s · ETA 28ms
WARN fast-xml-builder@1.2.0: registry response has no `dist.integrity`, importing without content verification. Set `strict-store-integrity=true` to refuse instead. code=WARN_NUB_MISSING_INTEGRITY
WARN xml-naming@0.1.0: registry response has no `dist.integrity`, importing without content verification. Set `strict-store-integrity=true` to refuse instead. code=WARN_NUB_MISSING_INTEGRITY
WARN strnum@2.4.0: registry response has no `dist.integrity`, importing without content verification. Set `strict-store-integrity=true` to refuse instead. code=WARN_NUB_MISSING_INTEGRITY
WARN semver@7.8.1: registry response has no `dist.integrity`, importing without content verification. Set `strict-store-integnub 0.2.10 — fetching   [███████████████████░]   372/379  pkgs · 130 kB · 145 kB/s · ETA 19ms
WARN fast-xml-parser@5.7.3: registry response has no `dist.integrity`, importing without content verification. Set `strict-store-integrity=true` to refuse instead. code=WARN_NUB_MISSING_INTEGRITY
WARN @types/aws-lambda@8.10.162: registry response has no `dist.integrity`, importing without content verification. Set `strict-store-integrity=true` to refuse instead. code=WARN_NUB_MISSING_INTEGRITY
WARN anynum@1.0.0: registry response has no `dist.integrity`, importing without content verification. Set `strict-store-integnub 0.2.10 — linking    [███████████████████░]   379/379  pkgs

After roughl 3 minutes i canceld the step

colinhacks:
Shipped in v0.3.0: https://github.com/nubjs/nub/releases/tag/v0.3.0

#300 replaced the linking display: the `N/N pkgs` counter you saw is now a live, growing file count that ticks up as the linker materializes files, so a slow link re-paints instead of freezing at the total. The append-only/CI renderer heartbeats the same count, so a slow link keeps producing output instead of going quiet.

[...]

PR #300, opened 2026-07-03
feat(install): bun/uv-style linking progress — spinner + live file count

colinhacks:
## What

[#288](https://github.com/nubjs/nub/issues/288) reported `nub i` looking stuck on the linking step. The linking phase surfaced no live progress: the TTY bar froze at its end-of-fetch fill for the whole phase, and the append-only heartbeat dedup'd on a constant `N/N pkgs · linking` string, so on a slow link (slow/network FS, no reflink, AV scanning) both renderers went silent and read as wedged.

This PR fixes the **observability** half. It redesigns the install-progress UI to the bun/uv shape:

- **TTY:** an animated spinner, the phase verb, and a live count — **no progress bar**. During linking the count is the materializer's file count, ticking up; the spinner is proof-of-life independent of it (a stalled filesystem still shows a moving spinner). The verb is right-padded to the widest verb so the count column never shifts when the phase word changes. A ~300ms debounce keeps the line hidden until the install has run long enough, so a fast install flashes nothing and only its summary prints.
- **Append-only / CI:** the linking line now carries the advancing file count, so a slow link re-emits on the ~2s heartbeat instead of going silent — while staying periodic, not per-file spam.

Mechanism: the linker bumps a shared counter once per materialized file (the isolated, GVS, and hoisted paths all funnel through it); a background ticker reads it and repaints, decoupled from the per-file rate so the repaint cost is negligible.

## Fork discipline

Gated on the embedder's `tty_progress` seam. Standalone aube keeps its append-only `N/N pkgs` output, never spawns the ticker, and never touches the counter — byte-identical default behavior.

## Verification

pty + piped captures on a 42k-file slow-copy re-link and an instant no-op:
- slow link: spinner animates while the count ticks `69 → 31924 → 37289 files`; no bar glyphs.
- instant install: zero spinner/progress frames, only the summary (debounce holds).
- phase transitions: the count field starts at one fixed column across resolving/fetching/linking (no layout shift).
- append-only: one `37231 files · linking` heartbeat line at the ~2s mark.

Gates: `clippy --all-targets` clean (aube, aube-linker, nub-cli), `fmt` clean, 721 aube + 109 aube-linker unit tests pass.

Environmental root cause (why *this* reporter's link was slow — FS/AV/registry) is separate and still needs their OS/FS + `RUST_LOG=debug` info, so this is `Refs`, not `Closes`.

Refs #288

### DIFF
--- a/vendor/aube/crates/aube-linker/src/builder.rs
+++ b/vendor/aube/crates/aube-linker/src/builder.rs
@@ -42,6 +42,29 @@ impl Linker {
             modules_dir_name: "node_modules".to_string(),
             aube_dir_override: None,
             no_integrity_read_keys: std::collections::BTreeMap::new(),
+            link_progress: None,
+        }
+    }
+
+    /// Supply a shared counter the materialize pass bumps once per linked
+    /// file, so the progress UI can render a live file count during the
+    /// linking phase. The install driver hands in the progress UI's
+    /// `files_linked` atomic. Omitted by standalone callers and tests →
+    /// the pokes are no-ops and link behavior is unchanged.
+    pub fn with_link_progress(
+        mut self,
+        counter: std::sync::Arc<std::sync::atomic::AtomicUsize>,
+    ) -> Self {
+        self.link_progress = Some(counter);
+        self
+    }
+
+    /// Bump the live file-linking counter by `n`, if one was installed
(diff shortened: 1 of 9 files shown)


colinhacks:

Issue #280, opened 2026-07-02
Unresolved dependency (`@crawlee/cheerio` → `@apify/datastructures`)


gullmar:
When using `@crawlee/cheerio`, the required `@apify/datastructure` is not resolved correctly, despite both being installed in `~/.cache/nub/pm/virtual-store` and `node_modules`. It works correctly when installing the dependencies with `npm`.

## Steps to reproduce

1. Create an empty project:

```bash
mkdir nub-demo && cd nub-demo && npm init -y
```

2. Add the dependency:

```bash
nub add @crawlee/cheerio
```

3. Create an `index.js` file with the following content:

```js
const { CheerioCrawler } = require("@crawlee/cheerio");

async function main() {
  const crawler = new CheerioCrawler({
    requestHandler: () => {
      console.log("Request handled.");
    },
  });

  await crawler.run([{ url: "https://example.com" }]);
}

main();
```

4. Run it:

```bash
nub index.js
```

Output:

```log
❯ nub index.js
~/.cache/nub/runtime-0.2.9-47bba2b1/preload-common.cjs:449
      throw err;
      ^

Error: Cannot find module '@apify/datastructures'
Require stack:
- ~/.cache/nub/pm/virtual-store/@crawlee+basic@3.17.0-b936eb22a3419960/node_modules/@crawlee/basic/internals/basic-crawler.js
- ~/.cache/nub/pm/virtual-store/@crawlee+basic@3.17.0-b936eb22a3419960/node_modules/@crawlee/basic/index.js
- ~/.cache/nub/pm/virtual-store/@crawlee+http@3.17.0-a63c7961b339a34f/node_modules/@crawlee/http/index.js
- ~/.cache/nub/pm/virtual-store/@crawlee+cheerio@3.17.0-7ca17fcfb301305a/node_modules/@crawlee/cheerio/index.js
- ~/Projects/work/nub-demo/index.js
    at Module.<anonymous> (node:internal/modules/cjs/loader:1568:15)
    at module_._resolveFilename (~/.cache/nub/runtime-0.2.9-47bba2b1/preload-common.cjs:752:34)
    at wrapResolveFilename (node:internal/modules/cjs/loader:1122:27)
    at defaultResolveImplForCJSLoading (node:internal/modules/cjs/loader:1146:10)
    at defaultResolve (node:internal/modules/cjs/loader:1203:20)
    at nextStep (node:internal/modules/customization_hooks:189:26)
[...]

PR #304, opened 2026-07-04
feat(pm): per-package force-materialization for subpath-adapter phantoms under GVS

colinhacks:
## What

Under nub's default global virtual store (GVS, off-CI), a package's realpath lives in the machine-global store, outside the project. Node resolves an undeclared import by walking realpath parents, so a subpath adapter that statically imports a consumer-installed backend it doesn't declare — `@hookform/resolvers/zod` importing the `zod` the app installed — never reaches the project root from inside the shared store, and the import fails. pnpm's default (per-project isolated) resolves this class; nub's default did not.

This adds a per-package escape hatch: a new `forceMaterializePackages` setting makes each listed package materialize as a real project-local directory even under GVS, while every other package stays a shared-store symlink. The listed adapter's realpath is then project-local, so Node's upward walk reaches the backend at the project root. nub seeds the list as an embedder default with the detector-validated subpath adapters, minus the build-time/helper subpath imports (`@babel/*`, `babel-plugin-macros`, `tslib`, `typescript`) that resolve anyway.

## How

- **aube-linker** — `Linker.force_materialize` set (empty by default); a branch in the GVS link pass materializes listed packages locally via the existing per-project `materialize_into` instead of symlinking to the shared store. Empty set ⇒ the GVS pass is byte-for-byte unchanged (standalone aube default path preserved).
- **install** — the install-state hash folds the list, so introducing or changing it forces a relink; `.aube` mode-detection is now symlink-priority, so the now-legitimately mixed tree (forced real dirs + shared-store symlinks) is classified as GVS rather than misread as per-project and wiped.
- **nub** — seeds the curated 16-adapter list as an embedder default; the whole-install `disableGlobalVirtualStoreForPackages` trigger stays at the validated-clean `next,nuxt,parcel`.

## Verification

Empirically confirmed under genuine default GVS (isolated HOME/XDG, off-CI):

- `@hookform/resolvers` is a real project-local directory; `react-hook-form` and `zod` stay machine-global symlinks.
- ESM and CJS `import '@hookform/resolvers/zod'` resolve (previously threw `Cannot find package 'zod'`).
- pnpm resolves the same fixture identically.
- A second install is idempotent — the forced entry stays a real dir, no wipe.

Tests: force-materialize layout (aube-linker), mixed-tree mode detection (aube), embedder-list override + accessor audit (aube-settings), seeded-list + List-1 trigger (nub-cli). `cargo clippy --all-targets --all-features -- -D warnings` and `cargo fmt --check` clean.

Refs #286
Refs #280

### DIFF
--- a/crates/nub-cli/src/pm_engine/mod.rs
+++ b/crates/nub-cli/src/pm_engine/mod.rs
@@ -2053,6 +2053,22 @@ fn strip_yarnrc_value(rest: &str) -> &str {
     rest.split('#').next().map(str::trim).unwrap_or(rest)
 }
 
+/// Curated `forceMaterializePackages` list: subpath adapters that statically
+/// import a consumer-installed backend they don't declare. Under nub's default
+/// GVS each of these must be a real project-local directory so its realpath
+/// stays inside the project and Node's upward walk reaches the backend at the
+/// project root (`@hookform/resolvers/zod` → the app's `zod`). Derived from the
+/// phantom-dep detector's top-5000 subpath-adapter offenders, minus the entries
+/// whose imported target is a ubiquitous BUILD tool (`@babel/*`,
+/// `babel-plugin-macros`, `tslib`, `typescript`) rather than a runtime
+/// consumer-backend — those resolve anyway and aren't the pick-your-backend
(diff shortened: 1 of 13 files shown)


colinhacks:

PR #306, opened 2026-07-04
cli: remove --posix-shell run flag

colinhacks:
Removes the `--posix-shell` run flag (renamed from `--shell-emulator` in #299, never released, never documented). It was a no-op on Unix (default script shell is already `sh`) and on Windows was sugar over `--script-shell`, which stays as the explicit escape hatch. Both old spellings now error as ordinary clap unknown arguments.

Drops the clap arg, `POSIX_SHELL` static, `find_posix_sh()`, and the spawn-path branch from `cli.rs`; deletes `tests/posix_shell.rs`; converts the Windows papercut-survey check to a `--script-shell` probe that locates `sh.exe` itself, preserving coverage.


colinhacks:

PR #309, opened 2026-07-04
pm: drop 4 storybook-family entries from the force-materialize seed

colinhacks:
Removes `@storybook/addon-interactions`, `@storybook/core`, `storybook`, `@storybook/builder-webpack5`: differential-verified false inclusions (real `storybook build` succeeds under default GVS with force-materialize on or off — flagged imports are esbuild/webpack-internal, never Node-resolved). `builder-webpack5` was actively harmful: materializing it alone broke its dependent `react-webpack5`; fixed post-removal, build now matches pnpm.

Keeps `drizzle-orm`: verified via `require.resolve()` on the real `expo-sqlite` subpath.

Reconciles `.fray/storybook-false-inclusion-reconcile.md`. Gates green (clippy/fmt/tests).

Refs #286


colinhacks:

PR #310, opened 2026-07-04
pm: force-materialize @prisma/client to stop shared-store codegen corruption

colinhacks:
Under default GVS, `@prisma/client`'s postinstall `prisma generate` writes the generated client into the package realpath — the machine-global store, keyed by version, not project/schema. Two projects with different schemas share one mutable client and silently overwrite each other's models.

Adding `@prisma/client` to the `forceMaterializePackages` default makes the entry project-local, so `generate` stays in-project (matches pnpm).

Validated before/after (shared realpath + bidirectional corruption → distinct project-local, isolated). Orphan-safe: `@auth/prisma-adapter` (store-resident peer) still resolves the intact store copy.

Refs #286


colinhacks:

PR #311, opened 2026-07-04
pm: add react-native to the default GVS-off list

colinhacks:
Bare React Native's Metro bundler crawls modules by realpath from the project root, so nub's machine-global virtual store is out of scope and even declared deps (`@babel/runtime`) report unresolved. Adds `react-native` to the default `disableGlobalVirtualStoreForPackages` (with `next`/`nuxt`/`parcel`) so RN projects get a project-local store Metro can see. A store-locality break, not a phantom-dep class.

Verified: under default GVS an RN project now links project-local vs. the machine-global store before. The trigger scans only direct deps, so a transitive-only `react-native` doesn't flip unrelated projects. Expo is store-aware and unaffected.

Refs #286


colinhacks:

PR #312, opened 2026-07-04
fix(linker): keep force-materialized package's store copy so store-resident dependents don't orphan

colinhacks:
The GVS force-materialize branch (#304) skipped the link-phase shared-store write, leaving the offender's store copy to the fetch prewarm — whose subtree hash can differ from the link-phase `subdir` a store-resident dependent's sibling symlink targets, dangling it. Force-materializing `@storybook/builder-webpack5` regressed `@storybook/react-webpack5` to `Cannot find module` (why #309 dropped it).

Fix: also ensure the store copy at `virtual_store_subdir(dep_path)` alongside the project-local dir — the write a normal install performs. Default-preserving (empty force-materialize set → branch never runs). Adds a regression test that fails without the fix.

Refs #304, #309


colinhacks:

PR #302, opened 2026-07-03
feat(nub-phantom): phantom-dependency detector + top-N ecosystem scanner

colinhacks:
Internal/eval tool that finds npm packages with undeclared (phantom) deps, to build the empirical `packageExtensions` list. Refs #286.

`crates/nub-phantom`: own Cargo workspace, excluded from the root workspace (like `nub-native`), keeping its oxc dep out of the shipped binary and CI.

Walks a package's published module graph (oxc), classifying each bare import: hard-phantom / soft / declared-optional-peer / declared / builtin / self. Excludes optional peers, guarded loads, type-only imports, unreached files, non-packages. Subpath-adapter provenance isolates the consumer-backend class.

Scan (top-5000): naive flags 607, real keeps 120 hard; 21 show the adapter class.

PR #314, opened 2026-07-04
feat(run): warn on undeclared (phantom) dependencies before a run

colinhacks:
Adds a static, warn-only phantom-dependency check to the pre-run `verify_deps` family. It warns when the project's own source imports a package that isn't declared in `package.json` but is only present transitively — the classic phantom that resolves under a flat npm/Yarn install and breaks under an isolated tree. The warning names the importing file and the fix; it never changes the exit code.

```
nub: src/index.ts imports `ansi-styles`, which isn't in package.json. Run `nub add ansi-styles` (WARN_PHANTOM_DEP).
```

## What it warns on

A package is flagged only when **every** one of these holds:

- imported **unguarded** (a `try/catch` / conditional optional load is left alone — reuses the detector's guard modeling),
- absent from `dependencies` / `devDependencies` / `optionalDependencies` / `peerDependencies`,
- not a Node builtin, the project's own name (self-import), or a `@types/*` package,
- and it actually **resolves in `node_modules` to a genuine hoisted dependency** — its real (symlink-resolved) location is inside a `node_modules` directory. This confirms the "present transitively" half of the definition and excludes workspace members (symlinked into `node_modules` but resolving to first-party source outside it).

It reads first-party source only (`node_modules` and build output skipped) and is bounded (file-count + file-size caps) so the once-per-command pre-run path can't blow up on a large repo.

## Precision (tested on real projects)

Verified end-to-end against real npm-installed (flat-hoisted) trees — zero false positives on every clean project:

| Project | Result |
| --- | --- |
| Plain TS lib (`chalk`) | clean → silent; inject `import 'ansi-styles'` → warns |
| Express app (67 hoisted transitive deps) | silent; inject `require('body-parser')` → warns |
| Monorepo member importing a workspace sibling via symlink | silent (not flagged) |
| Guarded `try { require('ansi-styles') } catch {}` | silent; unguarded control → warns |

## nub-phantom wiring

The precision-critical primitives (specifier extraction + guard modeling + builtin/name classification) move into a new lean crate, `nub-phantom-core`, shared by this CLI scan and the excluded `nub-phantom` eval tool — one source of truth. It depends on the individual oxc parser/AST crates (`oxc_parser`/`oxc_ast`/`oxc_ast_visit`/`oxc_span`/`oxc_allocator`, pinned `=0.132.0`) rather than the `oxc` umbrella with `features=["full"]`, so the shipped CLI gains the parser but not the transformer/codegen/semantic/minifier tail the main workspace excludes `nub-phantom` to avoid. `nub-phantom` now depends on the shared crate and drops its own `oxc` dep.

## Scope

Warn-only, on by default. Off via the neutral `phantom-check` `.npmrc` key or `NUB_PHANTOM_CHECK`; `--no-check` and `--node`/compat mode skip it, matching the freshness check. The runtime module-resolution-hook enforcement half is deferred.

Refs #302


colinhacks:

PR #316, opened 2026-07-05
perf(linker): fire whole-dir clonefile for scoped packages on macOS

colinhacks:
The macOS whole-dir `clonefile(2)` fast path created only `node_modules` before the syscall, not the scoped destination's `@scope` parent — so `clonefile` ENOENT'd for every scoped package and it fell back to the per-file copy loop. The fast path fired only for unscoped packages.

Fix: create `pkg_nm_dir.parent()` (the `@scope` dir for scoped, `node_modules` for unscoped). On a real firebase tree the macOS link phase drops ~3x (clonefile fires 145/145 vs 57).

Output is byte-identical (clone source built from the same CAS via `link_file_fresh`); macOS-only. 111 linker tests pass; clippy+fmt clean.

Refs #314

### DIFF
--- a/vendor/aube/crates/aube-linker/src/materialize.rs
+++ b/vendor/aube/crates/aube-linker/src/materialize.rs
@@ -985,13 +985,23 @@ impl Linker {
             if index.is_empty() {
                 return Ok(false);
             }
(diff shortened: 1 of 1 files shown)


colinhacks:

Issue #315, opened 2026-07-05
Shared (global) virtual store


adamhl8:
Related to #280 

I'm wondering if nub should rethink/adjust its architecture around `node_modules`. I've been migrating a lot of my projects to use nub, and I've already run into multiple issues where builds fail or errors occur because certain dependencies/tools try to resolve things using the realpath (and therefore are outside of the local `node_modules`).

A couple of examples:
- When running the dev server in an Astro project, Vite will throw these errors: `23:03:23 [ERROR] [vite] The request id "/Users/adam/.cache/nub/pm/virtual-store/@astrojs+react@6.0.1_@types+node@26.1.0_@types+react@19.2.17_@types+react-dom@19.2.3_@t_47d53c0c9b957dff170be869efb166ce/node_modules/@astrojs/react/dist/client.js" is outside of Vite serving allow list.`
- oxlint's type-aware rules can fail in certain cases. I opened an issue on the upstream tsgolint project for this: https://github.com/oxc-project/tsgolint/issues/1055

It seems like there are two fixes for this issue.
1. Set `node-linker=hoisted`, or
2. `enableGlobalVirtualStore=false`

I know there's also `forceMaterializePackages[]` and `forceMaterializePackages[]`, but it's not always clear which package needs to be added to these lists, and I'm also left wondering if I missed one and it'll bite me randomly later on.

Regardless, I actually really like that by default the `node_modules` install is isolated/non-hoisted. I think that the isolated structure is much more clear and I think it's a good thing that phantom dependencies will fail to resolve.

I know that you're force materializing packages when they have issues like this (e.g. @prisma/client), but I'm wondering if that's gonna be a never-ending game of whackamole.

For now, I've resorted to setting `enableGlobalVirtualStore=false` on all my projects.

Anyway, just want to open an issue for discussion on this topic.

Blinks44:
I’m seeing the same kind of issue in my project.

With nub and the global virtual store enabled, my docs app fails to start because a package from `~/.cache/nub/pm/virtual-store/...` tries to import vite, and Node can’t resolve it from there.

In my case this happens with `@tanstack/start-plugin-core` while loading a Vite config:
`Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'vite' imported from ~/.cache/nub/pm/virtual-store/.../@tanstack/start-plugin-core/...`
vite is installed in the app itself, so this also looks like a realpath / virtual-store resolution problem rather than a missing dependency.

[...]

PR #317, opened 2026-07-05
fix(pm): restore vite/vitepress to the disable-GVS default trigger

colinhacks:
Adds `vite`/`vitepress` back to nub's `disableGlobalVirtualStoreForPackages` default: `next,nuxt,vite,vitepress,parcel,react-native`.

Vite/Astro dev servers reject the global-store realpath via `server.fs.allow`, so symlink-to-global installs fail at dev-server startup (#315). nub-cli embedder default only; vendor/aube untouched.

Fallback is project-local isolated (`node_modules/.nub`, pnpm-`.pnpm` analog): isolation preserved, not hoisted. Tradeoff: Vite projects lose cross-project store sharing, pending upstream work.

Verified: clippy/fmt/test green; a vite-only fixture logs the fallback and symlinks `node_modules/vite` into a project-local `.nub` dir.

Refs #315

### DIFF
--- a/crates/nub-cli/src/pm_engine/mod.rs
+++ b/crates/nub-cli/src/pm_engine/mod.rs
@@ -2167,16 +2167,21 @@ fn nub_setting_defaults(
         ("stateDir".to_string(), "node_modules/.nub".to_string()),
         (
             "disableGlobalVirtualStoreForPackages".to_string(),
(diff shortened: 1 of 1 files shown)


colinhacks:
Superseded by #318, which preserves symlink-GVS for Vite via the `.modules.yaml` + backported-sniff approach rather than disabling GVS for vite/vitepress.

PR #318, opened 2026-07-05
Vite symlink-GVS serving compat (#315)

### LINKED ISSUE #315: Shared (global) virtual store

colinhacks:
Vite dev servers 403 store-resident modules under nub's machine-global virtual store (`~/.cache/nub/pm/virtual-store`, outside the project root): `403 … is outside of Vite serving allow list`. Vite realpath-checks every `/@fs`-served module against `server.fs.allow` (default `[workspaceRoot]`), and a store-resident dep served raw (framework client entries, dev-toolbar modules, SSR/`optimizeDeps`-excluded deps) falls outside it. pnpm never hits this — its virtual store is project-local, under the workspace root.

## The fix

One mechanism, PM-agnostic, works-without-nub, CAS-pristine — default-on, `NUB_VITE_COMPAT=0` opts out. Gated on Vite being in the graph.

- **Unit A — `node_modules/.modules.yaml`.** On install, whenever Vite is anywhere in the graph, write JSON `{"virtualStoreDir":"<abs store>"}` at the workspace root. Vite 8.1+ reads it natively (vitejs/vite#22415) and allows the store. Covers direct and library-embedded Vite.
- **Unit B — dist backport for Vite < 8.1.** For a directly-declared Vite, the linker force-materializes it project-local (the shared CAS store is untouched — only the ejected copy is modified), and nub codegen-inserts Vite's own 8.1 `.modules.yaml` sniff at its `fs.allow`-default site (`let allowDirs = server.fs.allow;` for v6/v7, `[searchForWorkspaceRoot(root)]` for v5). The inserted sniff is YAML-tolerant and PM-agnostic — it reads whatever `virtualStoreDir` any tool wrote; nub's store path lives only in the data file, never in the patch. Idempotent, fail-open (a missing anchor is a no-op, never a broken Vite).

The fix lives on disk in `node_modules` and injects nothing at runtime, so it **works whether or not nub is in the process** — a teammate on plain Vite, or CI, gets the same result.

## Library-embedded frameworks

A framework like Astro is store-resident and loads its Vite through a store-to-store sibling symlink — the shared store copy, not the project-local ejected copy — so the dist backport can't reach it CAS-safely. Unit A covers those for Vite ≥ 8.1 (the store Vite reads `.modules.yaml` natively), which is the reported #315 shape: Astro 7 + `@astrojs/react` pins Vite 8.1.3. A framework pinning Vite < 8.1 that does not set its own `fs.allow` (VitePress already does) is the remaining gap — it shrinks as the ecosystem moves to Vite 8.1+; closing it would mean patching the shared store copy, a separate call.

## Validation

Matrix harness under `tests/vite-compat/` (README + driver). Verified against real installs:

| case | Vite | result |
|---|---|---|
| Astro 7 + @astrojs/react (the #315 repro) | 8.1.3 | `@astrojs/react/dist/client.js` via `/@fs` 403→200, works-without-nub (`npx astro dev`) |
| create-vite ×7 (react/vue/svelte/solid/preact/lit/vanilla) | 8.1.3 | pass (native sniff) |
| SvelteKit, Storybook (Vite builder) | 8.1.3 | pass (native) |
| VitePress | 5.4.21 | pass (self-mitigates) |
| direct-dep `vite dev`, bare CLI, no nub | 5.4.21 / 6.4.3 / 7.3.6 | 403→200, backport applied |
| `NUB_VITE_COMPAT=0` | 7.3.6 | no `.modules.yaml`, no eject, byte-identical default |
| install ×2 | 7.3.6 | second is "Already up to date" — no re-link, patch not doubled |

Closes #315

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2668,6 +2668,7 @@ dependencies = [
  "aube-registry",
  "aube-scripts",
  "aube-settings",
+ "aube-store",
  "aube-util",
  "aube-workspace",
  "base64",
--- a/crates/nub-cli/Cargo.toml
+++ b/crates/nub-cli/Cargo.toml
@@ -61,6 +61,10 @@ aube-scripts = { path = "../../vendor/aube/crates/aube-scripts" }
 # Diagnostics / critical-path recorder: nub's engine_session calls diag::init()
 # so AUBE_DIAG_* env vars work when nub install runs the embedded aube engine.
 aube-util = { path = "../../vendor/aube/crates/aube-util" }
+# Store dir resolution: the Vite compat post-install step (pm_engine/vite_compat.rs)
+# reads the global virtual-store path via `aube_store::dirs::cache_dir()` (embedder
+# cache_namespace-aware) to write it into `.modules.yaml`. Already a transitive dep.
+aube-store = { path = "../../vendor/aube/crates/aube-store" }
 # Same major + `fancy` as vendor/aube — present.rs renders the engine's
 # miette::Report exactly as aube's own cli_main would, then rewrites it.
 miette = { version = "7", features = ["fancy"] }
--- a/crates/nub-cli/src/pm_engine/install_family.rs
+++ b/crates/nub-cli/src/pm_engine/install_family.rs
@@ -368,6 +368,12 @@ fn run_add(typed: &str, args: &[String]) -> Result<i32> {
     )?;
     super::min_release_age::persist(&session.cwd, code == 0, &globals.output);
     stamp_if_virgin(&session, code);
+    // `nub add vite` (or adding any dep to a vite project) changes the graph;
+    // refresh `.modules.yaml` + the < 8.1 patch. Note: a FIRST `nub add vite`
+    // can't have force-materialized vite yet (the manifest didn't declare it at
+    // settings time), so Unit A applies immediately (≥ 8.1) and the < 8.1 dist
+    // patch lands on the next `nub install` once vite is a declared direct dep.
+    apply_vite_compat(&session, code);
     Ok(code)
 }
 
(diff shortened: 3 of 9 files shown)


colinhacks:

PR #319, opened 2026-07-05
feat(pm): selective-subtree ancestor-closure force-materialization (behind NUB_DYNAMIC_PHANTOM_EJECT)

colinhacks:
Selective-subtree ancestor-closure force-materialization, behind the default-off `NUB_DYNAMIC_PHANTOM_EJECT` flag. Flag-off is byte-for-byte the shipped behavior.

## Why

Force-materializing a package project-local is only sound for a *transitively-consumed* package if its whole ancestor-closure materializes with it. Materializing the package alone leaves every store-resident importer resolving the un-materialized shared-store copy — a silent singleton split (two realpaths, two module instances). And a package's *undeclared* import (a transitive phantom target) still 404s even after the closure materializes, because the target isn't a sibling on its walk.

## What

Two rungs, both gated on the flag:

- **Rung 1 — ancestor-closure.** Each force-materialize seed grows to `LockfileGraph::importer_closure` — the seed plus every package that transitively imports it (reverse-BFS over importer edges). Bounded to the affected subtree by construction: unrelated top-level subtrees are never importers, so a framework's closure is a small slice of a large tree (measured 0.3–2.1%). Auto-detects an embedded `vite<8.1` and ejects its `[framework…vite]` closure, so `#318`'s dist patch reaches a now-project-local vite — the library-embedded-vite residual that direct-dep-only ejection left.
- **Rung 2 — hoist-within.** A known transitively-phantom importer (`@firebase/database`→`@firebase/app`, `vue-router`→`@vue/compiler-sfc`, `@nuxt/devtools`→`unstorage`) gets its already-resolved target injected as an extra sibling inside its own materialized `node_modules`.

## Fork-discipline

aube gains only a neutral graph primitive (`importer_closure`) and an embedder-pluggable expansion hook (`set_force_materialize_expand_hook` / `expand_force_materialize`) plus a `phantom_hoist` field. With no hook installed — standalone aube, every test — the hook is the identity and the materialize pass is byte-for-byte unchanged. All policy (the flag, vite detection, the phantom map) lives in nub-cli.

## Evidence

- **Firebase, flag-on:** `@firebase/database` doesn't declare `@firebase/app` (a real phantom), yet `require.resolve('@firebase/app')` from its realpath resolves. Flag-off: `MODULE_NOT_FOUND` (the shipped break).
- **Astro 5, flag-on:** embedded `vite@6.4.3` closure = 5 packages (1.5% of 327); a bare `astro dev` (no nub in the process) serves a store `/@fs` module `200` (was `403`).
- **Bounded-subtree**, measured on realistic large trees (framework + ~100 unrelated deps): Astro 0.28%, Nuxt 2.10% of 1284, Expo 1.67%, nx 0.98% — symlink-eligible 97.9–99.7%.

`cargo clippy --all-targets --all-features -- -D warnings` and `cargo fmt` clean; aube-linker's 111 tests plus the new `importer_closure` / `phantom_closure` unit tests pass.

## Scope

The default flip stays a separate decision. Rung-2's target map is provisional — the dynamic per-version scanner is its intended auto-source. Two `tests/vite-compat/` acceptance cases (Astro 5, Nuxt) are documented behind the flag; the driver's dev-server leak is fixed (kill the whole process tree, parse the bound port from the log).

Known limitations (documented in-code, for the experimental flag): the flag isn't folded into the install-state fingerprint, so enabling it on an already-installed tree needs a clean install to take effect; the provisional rung-2 map resolves a multi-version target to the first copy by name. Both are fixed at productionization.

Refs #315

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2663,6 +2663,7 @@ dependencies = [
  "anyhow",
  "aube",
  "aube-codes",
+ "aube-linker",
  "aube-lockfile",
  "aube-manifest",
  "aube-registry",
--- a/crates/nub-cli/Cargo.toml
+++ b/crates/nub-cli/Cargo.toml
@@ -31,6 +31,10 @@ nub-phantom-core = { path = "../nub-phantom-core" }
 # `nub ci` route through `aube::commands::install::run` in-process instead of
 # shelling out. See vendor/aube/AGENTS.md and src/pm_engine/.
 aube-lockfile = { path = "../../vendor/aube/crates/aube-lockfile" }
+# aube-linker: nub installs a force-materialize expansion hook (selective-subtree
+# materialization) via `aube_linker::set_force_materialize_expand_hook`; a pure
+# library crate (no CLI/tokio surface), same lean-graph rationale as aube-lockfile.
+aube-linker = { path = "../../vendor/aube/crates/aube-linker" }
 aube-manifest = { path = "../../vendor/aube/crates/aube-manifest" }
 # default-features = false drops aube's default set (`mimalloc`, `config-tui`),
 # neither of which nub uses: aube's `config-tui` (ratatui/crossterm) only feeds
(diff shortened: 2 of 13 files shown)


colinhacks:

PR #320, opened 2026-07-05
feat(phantom): wire dynamic per-version scanner into #319's force-materialize closure

colinhacks:
Follow-on to #319. Wires nub's extract-time phantom detector into #319's selective-subtree force-materialization as the auto-source for phantom importers and their undeclared targets, replacing the provisional hardcoded list. Everything stays behind the pre-existing default-off `NUB_DYNAMIC_PHANTOM_EJECT` flag.

## What changed

- **Producer** — an aube-store extract hook scans each fetched package for unguarded undeclared imports and writes a per-content-integrity verdict sidecar; a pre-install backfill fills sidecars for warm-cached packages on `install`/`ci`. Gated additive seam — standalone aube registers nothing, so the default path is a no-op.
- **Consumer** — `phantom_closure::expand` reads those sidecars to drive #319's closure: rung-1 seeds gain each flagged importer, rung-2 hoists its scanned targets within the materialized package.
- **Retires** the provisional `KNOWN_PHANTOM_TARGETS` map — the per-version sidecars are now the source, so there is no hand-maintained list.
- **Precision seed-selection** — a flagged importer is skipped only when every undeclared target is both reachable inside its own subtree and absent from the project top level; it seeds on any uncertainty, since a wrong skip re-introduces a real phantom break.
- Drops the superseded link-time eject consumer and its aube-store sidecar-dir seam.

## Scope

Flag off installs no hook and touches no sidecar, so the default install path is byte-identical to #319. Does not flip the default and does not retire the static `NUB_FORCE_MATERIALIZE_PACKAGES` seed.

## Verification

Gates green (`fmt`, `clippy --all-targets`, scoped tests for nub-cli / aube-linker / aube-store / nub-phantom-scan). End-to-end on a genuinely cold store: `@firebase/database` ejects project-local through the dynamic source (flag off leaves it symlinked) with `@firebase/app` hoisted as a sibling (rung-2); a scanner-flagged but transitively-satisfied `es-abstract` stays symlinked (the precision filter); neutralizing a package's sidecar removes its eject and restoring it brings it back. The `vendor/aube` diff is gated additions only (39 lines across three aube-store files), and aube stays oxc-free.

Refs #319

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2685,6 +2685,7 @@ dependencies = [
  "node-semver",
  "nub-core",
  "nub-phantom-core",
+ "nub-phantom-scan",
  "num_cpus",
  "rayon",
  "rustc-hash",
@@ -2741,6 +2742,15 @@ dependencies = [
  "oxc_ast_visit",
  "oxc_parser",
  "oxc_span",
+]
+
+[[package]]
+name = "nub-phantom-scan"
+version = "0.2.10"
+dependencies = [
+ "nub-phantom-core",
+ "serde",
+ "serde_json",
 ]
 
 [[package]]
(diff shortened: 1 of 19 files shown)


colinhacks:

PR #321, opened 2026-07-05
feat(phantom): default-on dynamic phantom eject; purge static materialize/GVS lists

colinhacks:
Makes dynamic per-version phantom detection + ancestor-closure eject the default install behavior. Previously this was behind the default-off `NUB_DYNAMIC_PHANTOM_EJECT` flag (the detector→closure integration landed in the #319/#320 stack). Now the detector is the source of truth for which packages force-materialize project-local, and `NUB_DYNAMIC_PHANTOM_EJECT=0` is the opt-out to a pure-symlink tree.

## What changed

- **Flag default → on.** The on/off arm is centralized in one function (`dynamic_phantom::enabled`) shared by the extract-time scanner, the link-time closure, and the fingerprint, so they cannot drift. Unset/empty/any non-falsey value is on; `0`/`false`/`no`/`off` is the opt-out.
- **Retired the 14-entry static `NUB_FORCE_MATERIALIZE_PACKAGES` const.** The dynamic detector subsumes it, including the singleton-hazard adapters (preact, @angular/common, @angular/router, @apollo/client) that #319's ancestor-closure now makes sound. The user `forceMaterializePackages` override is unchanged; the embedder default now carries only the vite eject (#315).
- **Dropped `nuxt` from `disableGlobalVirtualStoreForPackages`** (closure-ejectable at rung 1). `next` (Turbopack single-root chroot), `parcel`, and `react-native` (Metro realpath crawl) remain.
- **Folded the effective flag into aube's install-state `settings_hash`** via a new default-preserving embedder hook (`Embedder::extra_settings_fingerprint`, `None` for standalone aube), so flipping the flag — an upgrade moving the default, or a user opt-out — invalidates a warm tree and re-links instead of trusting a stale `node_modules`.
- **Split `phantom_closure::expand`** into a store-IO entry + a pure `plan_from_flags` planner so the closure unit tests stay hermetic under the new default.

Standalone aube is byte-preserved on its default path (the fingerprint hook is `None`, the `disableGlobalVirtualStoreForPackages` schema default is untouched).

## Verification

`clippy --all-targets --all-features -D warnings` clean, `fmt` clean, nub-cli unit tests + integration suite (install_engine, pm_two_mode, pm_verbs, integration) green, aube identity-neutrality + `settings_hash` tests green. e2e on real fixtures: `@hookform/resolvers` + `preact` eject by default; `NUB_DYNAMIC_PHANTOM_EJECT=0` returns to pure symlink; a clean vite+react app stays ~97% symlink (only the vite compat eject); `settings_hash` differs by flag state.

Refs #319, #320.

### DIFF
--- a/crates/nub-cli/Cargo.toml
+++ b/crates/nub-cli/Cargo.toml
@@ -26,7 +26,8 @@ nub-core = { path = "../nub-core" }
 # crate's Cargo.toml for why that keeps the shipped CLI's build graph lean.
 nub-phantom-core = { path = "../nub-phantom-core" }
 # The reachable-graph phantom SCANNER (graph walk + classify) for the dynamic
-# per-version scan → sidecar path (behind NUB_DYNAMIC_PHANTOM_EJECT). Lean by
+# per-version scan → sidecar path (default-on; NUB_DYNAMIC_PHANTOM_EJECT=0 opts
+# out). Lean by
 # construction — same oxc-parser weight as nub-phantom-core, no fetch stack. The
 # `ScanResult` it defines is the sidecar payload phantom_closure deserializes.
 nub-phantom-scan = { path = "../nub-phantom-scan" }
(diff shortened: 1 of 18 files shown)


colinhacks:

PR #322, opened 2026-07-06
Phantom detector: remove dead extraction ladder, dedup eval tool, close two precision gaps

colinhacks:
Cleanup and precision work on the phantom-dependency detector, in four scoped commits.

## Remove the rejected `extract_optimized` ladder

A static-imports-only extraction fast-path was built, measured to regress the parse-dominated scan, then retained solely as an A/B benchmark seam in a crate linked into the shipped binary. This removes it end to end — the `extract_optimized` fn and its helpers, its parity test, the `NUB_PHANTOM_EXTRACT_MODE` dispatch, and the scan-bench binary. The baseline full-AST `extract` is the sole path.

## Dedup the eval tool onto `nub-phantom-scan`

The excluded `nub-phantom` eval tool carried its own byte-identical `manifest.rs` and near-identical `classify.rs`/`graph.rs`. The shipped scanner was lifted into `nub-phantom-scan`, but the eval tool was never rewired onto it, leaving duplicate source that could silently diverge. It now depends on `nub-phantom-scan` and re-exports its `classify`/`graph`/`manifest`; the local copies are deleted. The scan crate's walk is output-identical, so the tool's behavior is unchanged (verified against `@hookform/resolvers`).

## Classify `~/…` src-root aliases as non-packages

A leading `~` cannot begin an npm package name, but a bundler `~/components` src-root alias previously parsed to `Bare("~")` and could false-flag as a phantom, forcing an unnecessary eject. Added `~` to the package-name leading-char reject.

## Detect the immediately-invoked `createRequire` and no-substitution template forms

Two dynamic module forms carrying a static string target were dropped, missing real phantoms: an immediately-invoked `createRequire(import.meta.url)('pkg')` and a no-substitution template passed to `import(...)`. Both are now recorded via the same guard-aware path as a literal `require`, strictly limited to literal-arg forms — a substituted template, a computed argument, and a `createRequire` result bound to a name and called later all stay skipped, so no false positives are added.

## Verification

`cargo clippy --all-targets --all-features -- -D warnings` and `cargo fmt --check` are clean; the nub-phantom-core, nub-phantom-scan, and nub-cli phantom tests pass. An end-to-end scan confirms the two new forms flag and the `~/…` alias does not.


colinhacks:

PR #323, opened 2026-07-06
test(framework-matrix): fix driver.sh dev-server leak; add fresh-build acceptance runner

colinhacks:
Test-harness only — no runtime or PM behavior change.

**driver.sh dev-server leak.** The Vite-compat driver tore down the dev server with a `pgrep -P` walk, which misses a worker once it reparents to init before the walk snapshots it. A leaked worker held the port, so the next run bound a different port and curled the stale (unpatched) server, producing false 403s. The dev server is now launched as its own process-group leader (`set -m`) and killed via `kill -SIGNAL -PGID`, which signals the whole tree atomically; the `pgrep` walk remains only as a fallback.

**New framework-matrix runner.** `tests/framework-matrix/run.sh` drives one scaffolded app through the three distinct code paths — dev server, production build, serve-the-build — asserting each on HTTP status / exit code, plus a realpath-escape probe reporting whether a dep symlinks into the global virtual store or is materialized project-local. It reuses the process-group teardown above.


colinhacks:

PR #325, opened 2026-07-06
docs(install): drop Nuxt from the whole-install project-local-store list

colinhacks:
The whole-install project-local-store list on the install page named Nuxt, but Nuxt no longer takes a whole-install store. Under the default shared store it ejects only `nuxt` and its phantom closure project-local; ordinary deps stay symlinked into the global virtual store. Verified out of the box — install, dev, build, and preview all pass on a fresh Nuxt 4 scaffold, with `vue`/`vue-router` resolving into the global store and only `nuxt` ejected. The list is now Metro (bare React Native), Next.js, and Parcel.


colinhacks:

PR #324, opened 2026-07-06
refactor(pm): rename forceMaterializePackages to diskMaterializePackages

colinhacks:
Clean rename aligning the per-package materialization knob with the locked `install.materialization: symlink|disk` axis. Nothing shipped in a nub release, so no alias — every occurrence renamed:

- Setting `forceMaterializePackages` → `diskMaterializePackages` (settings.toml + npmrc/env/yaml keys; `resolved::disk_materialize_packages` accessor follows).
- Linker internals `force_materialize*`/`ForceMaterializePlan`/`FmExpandHook` → `disk_materialize`/`DiskMaterialize`; module renamed.

Default is the empty list, so the resolved layout is byte-for-byte unchanged (aube + nub). First of two PRs; the GVS/hoist structural collapse follows stacked.

PR #326, opened 2026-07-06
refactor(install): make illegal GVS/hoist layout states unrepresentable

colinhacks:
Stacked on #324. Collapses the `gvs_over_default_hoist`+`hoist`+`enableGlobalVirtualStore` decision tangle into a sound type + loud error.

- `Materialization { Symlink | Disk { hidden_tree } }` replaces the `(effective_gvs, build_hidden_tree)` bools whose `(true,true)` was a representable contradiction (hidden tree in the shared store, aube #6) — now unrepresentable.
- Explicit `enableGlobalVirtualStore=true` + explicit `hoist=true`/`node-linker=hoisted` errors loudly instead of silently dropping the request.

Behavior-preserving: unit tests assert `resolve` reproduces the old values; only the contradiction changes. Standalone aube unchanged. E2E-verified.


colinhacks:

PR #327, opened 2026-07-06
fix(resolver): auto-installed peers bind a resolved major, not the registry highest

colinhacks:
Fixes an out-of-box install failure on Expo / React Native: a fresh scaffold with `react-native-reanimated` installs cleanly, but `expo export` (and Metro dev bundling) fails with:

```
[Worklets] Requires Babel "^7.0.0-0", but was loaded with "8.0.1".
```

No open issue tracks this; it was found by a framework install-matrix sweep.

## Root cause

`react-native-worklets` declares `@babel/core: "*"` as a peer with no co-declared dependency, and `react-native` hard-deps `@babel/core: "^7.25.2"` (→ 7.29.7). Nothing in the graph hard-deps a babel 8 — `@babel/core@8.0.1` is only an uninstalled transitive devDependency.

With `auto-install-peers=true` (pnpm's default), a missing peer like this gets auto-installed. pnpm hoists missing peers as a second phase *after* the dependency tree is built, binding each to `maxSatisfying(preferredVersions, range)` over versions already resolved in the graph — so the `*` peer resolves to 7.29.7 and babel 8 never enters the tree.

nub resolved the auto-installed peer inline in its single resolution pass. When a `*` peer's version pick won the race against the hard dependency that constrains it, `pick_version` took the registry-highest satisfying version (8.0.1) — a major nothing declared a dependency on. That poisoned the resolved-version pool, so the whole tree bound babel 8 and Metro's Worklets plugin rejected it.

## Fix

Auto-installed transitive peers are parked in a new `deferred_auto_peers` queue instead of the main resolution queue, and drained only once the whole tree has resolved — mirroring pnpm's post-tree peer hoist. Each parked peer binds to the highest already-resolved version satisfying its range; only a genuinely missing peer (nothing resolved satisfies) falls through to a registry fetch. This makes the aube resolver match pnpm and never introduce a major that no dependency requested.

## Verification

Minimal fixture (`react-native@0.86.0` + `react-native-worklets@0.10.1`), before → after:

- Before: `@babel/core` resolves to both 7.29.7 and 8.0.1; worklets binds 8.0.1.
- After: only 7.29.7; worklets binds 7.29.7. Matches pnpm 10.

Full Expo scaffold with reanimated:

- Before: `expo export` fails with the Worklets babel error above.
- After: `expo export` succeeds (iOS bundle, 938 modules).

Gates: `clippy --all-targets --all-features -D warnings` clean, `fmt` clean, the aube-resolver suite passes (259 tests), and a new `#[ignore]` network regression test (`wildcard_peer_binds_resolved_major_not_registry_highest`) passes with the fix and fails without it.

The change is scoped to transitive auto-installed peers, where the failure occurs.

### DIFF
--- a/crates/nub-cli/tests/install_engine.rs
+++ b/crates/nub-cli/tests/install_engine.rs
@@ -787,3 +787,53 @@ fn frozen_install_with_trust_downgrade_still_aborts() {
         "no package may be linked when the trust gate aborts resolution"
     );
 }
+
+/// Regression (Expo/RN out-of-box failure): an auto-installed WILDCARD peer
+/// must bind to a major already present in the resolved graph, never a
+/// registry-highest major nothing declared a dependency on.
+///
+/// `react-native-worklets@0.10.1` declares `@babel/core: "*"` as a peer with
+/// no co-declared dependency, while `react-native@0.86.0` hard-deps
+/// `@babel/core: "^7.25.2"` (→ a 7.x). pnpm binds the `*` peer to that 7.x.
+/// nub used to resolve the auto-installed peer inline mid-BFS, racing the hard
+/// dep; when the peer won it fetched the registry-highest `@babel/core` (8.x)
+/// that no range asked for, so Metro's Worklets babel plugin rejected the tree
+/// (`Requires Babel "^7.0.0-0", but was loaded with "8.0.1"`) and `expo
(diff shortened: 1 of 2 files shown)


colinhacks:

PR #328, opened 2026-07-06
fix(phantom): key sidecars by scanner version so a detection improvement re-scans cached content

colinhacks:
Phantom-dependency sidecars are keyed by content fingerprint alone (`<store>/v1/phantom/<fingerprint>.json`). A version's bytes hash the same forever, so a scanner-logic improvement — e.g. the R3 createRequire/template detection — never re-analyzes already-cached content: the producer's warm-hit early return and the backfill both stop at the existing sidecar, and the stale verdict is served indefinitely. Harmless while nothing is released, but once the default-on eject flip ships in a release, a user who scanned under scanner-vN and then upgrades keeps the old verdicts and can miss a real phantom the newer scanner would catch (a latent resolution break).

This folds a scanner-version segment into the sidecar path so a version bump invalidates the prior generation:

- Adds `PHANTOM_SCANNER_VERSION` (bump on any change to the scanner's detection logic) and routes both halves through one `sidecar_path(base, fingerprint)` helper returning `<store>/v1/phantom/s<VERSION>/<fingerprint>.json`. The extract-time producer (`scan_and_cache`) and the link-time consumer (`dynamic_phantom_flags`) now derive the path identically and cannot drift — a producer/consumer disagreement would silently read every verdict as "no eject."
- A version bump makes each prior sidecar's path unreachable, so the extract hook + `backfill_from_lockfile` re-scan and write fresh verdicts under the new version; the old `s<N>` subdir is ignored and GC-able as a unit.
- The version is also folded into the install-state fingerprint (`settings_fingerprint`), so a bump invalidates the WARM tree too: without it a bump re-scans content but aube skips the link phase on an unchanged lockfile + flag, leaving the improved verdict written but never applied. The coupling is now structural — bump the const and the warm tree relinks and picks up the new verdicts. Folded in the enabled branch only, so the opt-out token stays byte-identical (`dynamic_phantom_eject=false`) and the no-op install path is preserved.
- Initial version is 1 (the post-R3 scanner); the per-version subdir keeps each generation grouped for wholesale cleanup.

The `NUB_DYNAMIC_PHANTOM_EJECT=0` opt-out remains a strict no-op — no phantom dir is touched, and its install-state token is unchanged.

## Verification

Dev build, isolated data dir, cold + warm installs:

- Fresh scan writes under `store/v1/phantom/s1/<fingerprint>.json`; nothing at the flat path.
- A planted old flat-path sidecar is ignored — the backfill re-scans to `s1/` and never reads or overwrites the flat file.
- Bumping the const (throwaway) relocates fresh verdicts to `s2/`, leaves `s1/` untouched, and changes the persisted install-state `settings_hash` (H1 → H2) — so a warm tree with an unchanged lockfile + flag relinks and picks up the new-version verdicts; a same-version reinstall keeps the hash stable (the relink is one-time). Reverted after.
- Flag-off cold install writes no `phantom/` dir at all, and its token stays `dynamic_phantom_eject=false`.

Gates green: `cargo clippy --all-targets --all-features -- -D warnings`, `cargo fmt --check`, scoped `cargo test -p nub-cli` (phantom + identity suites + a new `sidecar_path` contract test).

Refs the dynamic-phantom-detector scanner-version forward-compat gap.

### DIFF
--- a/crates/nub-cli/src/dynamic_phantom.rs
+++ b/crates/nub-cli/src/dynamic_phantom.rs
@@ -75,7 +75,13 @@ pub(crate) fn enabled() -> bool {
 /// scanner (`dynamic_phantom_eject=false`), preserving the strict no-op install
 /// path (a scanner version can't matter when nothing scans).
 pub(crate) fn settings_fingerprint() -> String {
-    if enabled() {
+    settings_token(enabled())
+}
+
+/// Pure token builder, split from [`settings_fingerprint`] so the byte-identity
+/// contract is testable without mutating the process-global `enabled()` env.
+fn settings_token(enabled: bool) -> String {
+    if enabled {
         format!("dynamic_phantom_eject=true;phantom_scanner={PHANTOM_SCANNER_VERSION}")
     } else {
         "dynamic_phantom_eject=false".to_string()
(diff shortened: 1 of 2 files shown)


colinhacks:

PR #329, opened 2026-07-06
fix(install): stop force-materializing vite ≥ 8.1 and its closure

colinhacks:
The disk-materialize embedder default seeds a version-blind `"vite"` name for any direct-dep vite, but vite ≥ 8.1 reads `node_modules/.modules.yaml` from the shared store natively (#318 "Unit A", written post-install regardless of eject) and needs no eject. The version-blind seed dragged vite plus its whole ancestor-closure project-local for zero benefit, defeating global-store dedup for the vite frameworks.

## The fix

`plan_from_flags` (the version-aware closure hook, which sees the resolved lockfile) now drops the `"vite"` name-seed and lets the existing `vite_lt_8_1` dep_path auto-seed decide vite. vite < 8.1 still ejects unchanged (the auto-seed is a strict superset of what the name-seed caught for < 8.1, direct and embedded); vite ≥ 8.1 stays symlinked to the global store and is served via `.modules.yaml`. The version-blind seed can't be gated at the mod.rs settings-defaults site — that runs pre-resolve and can't see the concrete version — so the closure hook is the only version-aware chokepoint.

## Before / after (real `create-*` scaffolds, isolated store each)

vite@8.1.3 resolved across all three; `node_modules/vite` realpath:

| Scaffold | before (ejected → project `.nub`) | after (symlinked → global store) |
|---|---|---|
| Vite (react), direct `vite@8.1.1` | `vite` + closure ejected | `vite` symlinked |
| SvelteKit, direct `vite@8.0.16` | `vite`, `@sveltejs/kit`, `vite-plugin-svelte`, `adapter-auto` all ejected | all symlinked (`svelte-check` still ejects — independent dynamic-phantom seed, not vite) |
| Astro, embedded `vite@7.0.6` | already symlinked (never declared vite directly) | unchanged (control) |

## Merge gate — vite ≥ 8.1 works symlinked

On the after (symlinked) installs, with `.modules.yaml` written:

| Scaffold | `run build` | `run dev` `/` | `run dev` `/@vite/client` | `run preview` `/` |
|---|---|---|---|---|
| Vite (react) | rc 0 | 200 | 200 | 200 |
| SvelteKit | rc 0 | 200 | 200 | 200 |
| Astro | rc 0 | 200 | 200 | 200 |

No `fs.allow` / 403 errors in any dev-server log. `/@vite/client` is served from the store via `/@fs`, so its 200 confirms the store path is on `fs.allow` (the exact #315 surface). The eject was not load-bearing for vite ≥ 8.1.

## Caveat

Under the `NUB_DYNAMIC_PHANTOM_EJECT=0` opt-out no closure hook is installed, so the raw name-seed still over-ejects vite ≥ 8.1 there — strictly-more-materialization on a rare opt-out, not a break (`.modules.yaml` is written regardless).

Separately, `settings_fingerprint` folds only the eject on/off flag, not the eject policy, so a project installed pre-fix (vite ≥ 8.1 ejected) and re-installed post-fix on the warm-tree fast path keeps the stale ejected shape until a genuine re-link (lockfile change / cache miss / `--force`). Deferred dedup, not a break — fresh installs get the new shape. Whether a nub-upgrade should force-relink existing vite ≥ 8.1 trees is a separate product call.

Refs #318

### DIFF
--- a/crates/nub-cli/src/pm_engine/phantom_closure.rs
+++ b/crates/nub-cli/src/pm_engine/phantom_closure.rs
@@ -97,6 +97,10 @@ fn plan_from_flags(
     // pre-resolve and can't see the concrete version. (The
     // `NUB_DYNAMIC_PHANTOM_EJECT=0` opt-out installs no hook, so its raw name-seed
     // still over-ejects vite ≥ 8.1 — an accepted cost of that rare opt-out path.)
+    // Provenance-blind: this also strips a user's explicit `vite` in
+    // `diskMaterializePackages`, which is fine — vite ≥ 8.1 works symlinked and
+    // vite < 8.1 is re-seeded below regardless of source, so a working vite is
+    // served either way.
     let seed_names: Vec<&str> = seed_names
         .iter()
         .map(String::as_str)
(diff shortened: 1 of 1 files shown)


colinhacks:

PR #330, opened 2026-07-06
fix(install): restrict phantom-eject safe-skip reachability to depth-1 deps

colinhacks:
## The bug

The phantom-eject precision filter (`should_seed`) skips the eject for a flagged importer when every undeclared target is already reachable without a project-local copy. It computed "reachable" with a multi-hop BFS over the whole `dependencies` closure — but under nub's isolated (GVS) layout a store-resident package's `node_modules` holds only its DIRECT (depth-1) deps as symlinked siblings, and Node's ancestor walk from its files reaches nothing deeper. A target declared by a transitive (depth-≥2) dep was counted reachable yet is not actually resolvable, so the eject was wrongly skipped and the phantom import failed.

Found by the #280 acceptance test: `@crawlee/basic` phantom-imports `@apify/datastructures`, declared by `@crawlee/core` (a direct dep of `basic`) — depth 2 → skipped → `Cannot find module '@apify/datastructures'`.

## The fix

Replace the BFS with a depth-1 `direct_dep_names`: only the importer's own direct declared deps that resolve in the graph count as reachable. A depth-≥2 or absent target falls through to SEED (eject + rung-2 hoist), matching the module's stated safety invariant. Monotonic in the eject direction — it only removes over-confident skips, never adds one. The top-level gate is unchanged.

## #280 before / after (isolated cold store, default eject on)

- `@crawlee/basic` → `@apify/datastructures`: **before** `MODULE_NOT_FOUND`; **after** `@crawlee/basic` disk-ejects (real dir, not a store symlink), `@apify/datastructures` is hoisted in as a sibling, resolves. `NUB_DYNAMIC_PHANTOM_EJECT=0` still reproduces the break — the eject is what fixes it.
- `@inkjs/ui@2` → `react` (the other #280 pair, driven by the untouched top-level gate): unaffected, still fixed.

## Impact — more ejects, only where a depth-≥2 phantom is real

A/B, same install on this branch vs pre-fix `main`, ejected-package delta:

| tree | pkgs | ejected before | ejected after | result |
|---|---|---|---|---|
| es-shims (`es-abstract` + friends) | 91 | 1 | 8 | before: `Cannot find module 'available-typed-arrays'`; after: all resolve |
| firebase | 86 | 3 | 3 | resolves both |
| Vite-react | 65 | 3 | 3 | builds both |
| SvelteKit | 50 | 7 | 7 | builds both |

The delta is 0 on trees with no depth-≥2 phantom-flagged package; it adds ejects only for genuinely-broken importers. es-shims was silently broken pre-fix and is fixed here too. No tree lost an eject.

## Tests

`cargo test -p nub-cli` green; `clippy --all-targets --all-features` and `fmt --check` clean. New: `depth2_phantom_target_seeds_not_skipped` (the #280 shape — fails before, passes after) and `depth1_phantom_target_still_skips` (guards the precision win); the `direct_dep_names` unit tests are reframed to depth-1 semantics.

Refs #280

### DIFF
--- a/crates/nub-cli/src/pm_engine/phantom_closure.rs
+++ b/crates/nub-cli/src/pm_engine/phantom_closure.rs
@@ -28,15 +28,15 @@
 //! reads those sidecars — so there is no hand-maintained list of phantom classes;
 //! the detection is per-version and auto-current. A precision SEED-SELECTION
 //! filter (see [`should_seed`]) drops a flagged importer whose undeclared targets
-//! are all already reachable inside its own subtree and absent from the project
-//! top level, so a transitively-satisfied over-flag never ejects.
+//! are all already resolvable as its own DIRECT (depth-1) siblings and absent from
+//! the project top level, so a directly-satisfied over-flag never ejects.
 //!
 //! Opt-out (`NUB_DYNAMIC_PHANTOM_EJECT=0`) ⇒ no hook installed ⇒ aube's
 //! `expand_disk_materialize` returns the seed verbatim ⇒ the disk-materialize
 //! pass is byte-for-byte the pre-productionization pure-symlink behavior. All
 //! policy lives here; aube owns only the neutral seam + the graph primitive.
 
-use std::collections::{BTreeMap, HashSet, VecDeque};
+use std::collections::{BTreeMap, HashSet};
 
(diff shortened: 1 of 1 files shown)


colinhacks:

PR #333, opened 2026-07-06
fix(phantom): scan extensionless main/exports entries instead of dropping them

colinhacks:
The phantom scanner discovers a package's entry points from `main`/`module`/`bin`/`exports`, then walks the reachable module graph to find undeclared (phantom) imports the installer must disk-eject. `collect_entry_points` gated each candidate entry path on a literal JS extension, so an extensionless entry was dropped before the walk ever ran. Node resolves `main: "./dist/index"` to `./dist/index.js` and a directory `main: "./dist"` to `./dist/index.js`, but the scanner saw no extension and discarded the entry — the package scanned zero files, no phantom was flagged, nothing was ejected.

`@vercel/static-build@2.11.4` declares `main: "./dist/index"` and imports `@vercel/build-utils` without declaring it. Before this change a default install scanned zero files for it, so the phantom was never ejected and the import failed at runtime:

```
Error: Cannot find module '@vercel/build-utils'
```

## Fix

Admit extensionless and directory-style candidates and defer their resolution to the graph walk's existing `resolve_entry`, which already runs Node's own ladder (exact file → `<path>.{js,mjs,cjs,…}` → `<path>/index.{…}` → `package.json` `main`). A candidate that resolves to nothing contributes nothing, so no false entry is introduced. Non-JS asset/type targets (`.json`, `.node`, `.wasm`, `.css`, `.d.ts`) are still dropped and never probed. The change is monotonic toward recall — every path the old gate admitted is still admitted, so no currently-detected phantom is lost.

`PHANTOM_SCANNER_VERSION` bumps 1 → 2 so content already scanned under the old logic (its sidecar recorded no phantom) is re-scanned instead of served stale.

## Verification

- `@vercel/static-build@2.11.4`, fresh isolated install: the scanner now flags `@vercel/build-utils`, the package ejects, and the import resolves. With `NUB_DYNAMIC_PHANTOM_EJECT=0` the `MODULE_NOT_FOUND` reproduces, confirming the eject is what closes the gap.
- A verdict differential across 148 installed packages (including `firebase`, `@hookform/resolvers`, and their trees) changes exactly one verdict — `@vercel/static-build` — and leaves every other package byte-identical.
- Odd `main` shapes each behave correctly: a missing file, a directory with no index, a `.json` main, a `.node` main, no `main`, and an exports-only map with extensionless conditional targets all resolve to the right file or to nothing — no crash, no false entry.

New regression tests cover the extensionless and directory `main` shapes end-to-end (`files_analyzed > 0` and the phantom flagged); both fail on the old gate and pass after the fix.

This is the producer-side (scanner recall) complement to #330's consumer-side reachability fix.

Refs #330

### DIFF
--- a/crates/nub-cli/src/dynamic_phantom.rs
+++ b/crates/nub-cli/src/dynamic_phantom.rs
@@ -221,7 +221,7 @@ pub(crate) fn phantom_cache_dir() -> Option<PathBuf> {
 /// after upgrade; harmless and expected (the whole point is to pick up the better
 /// verdict). Just bump the number when the scanner logic changes — the coupling
 /// is structural, nothing else to remember.
-pub(crate) const PHANTOM_SCANNER_VERSION: u32 = 1;
+pub(crate) const PHANTOM_SCANNER_VERSION: u32 = 2;
 
 /// THE single source of truth for a phantom sidecar's location: the versioned
 /// subdir `<phantom_cache_dir>/s<PHANTOM_SCANNER_VERSION>/<fingerprint>.json`.
--- a/crates/nub-phantom-scan/src/lib.rs
+++ b/crates/nub-phantom-scan/src/lib.rs
@@ -266,6 +266,56 @@ mod tests {
     }
 
     #[test]
+    fn extensionless_and_directory_main_are_scanned() {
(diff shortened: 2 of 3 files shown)


colinhacks:

PR #335, opened 2026-07-06
Make phantom-eject unconditional for users; keep an internal test seam

colinhacks:
Maintainer decision (2026-07-06): users cannot disable phantom-eject. The escape hatch for a suspected eject bug is disabling the global virtual store (`node-linker`/`disableGlobalVirtualStore`), not a phantom toggle.

- `enabled()` no longer reads `NUB_DYNAMIC_PHANTOM_EJECT`; a stale `=0` now has no effect.
- Sole off-switch is `__NUB_PHANTOM_EJECT_DISABLE`, an undocumented internal test seam for the A/B control.
- #328 fold simplified: the user token folds only `PHANTOM_SCANNER_VERSION`, so a scanner bump still relinks a warm tree.

No `vendor/aube` change. Clippy/fmt/phantom tests (299) green; e2e confirms the old var is ignored, the seam disables, the default ejects, and a scanner bump relinks.


colinhacks:

PR #337, opened 2026-07-06
fix(progress): animate the install spinner fast + smooth

colinhacks:
Spinner crawled at ~5 Hz: clx's `{{spinner()}}` advances the glyph at a hardcoded 200 ms/frame (no override API) and floors repaints at `interval()/2` = 100 ms.

Fix, entirely in the nub-only TTY path (standalone aube uses the append-only renderer, unchanged):
- Self-compute the braille glyph as a `spin` prop (`elapsed / 60 ms`) instead of `{{spinner()}}`.
- Ticker 90->16 ms; `set_interval(2x tick)` lowers clx's repaint floor.

Verified on a real TTY (PTY-timestamped): glyph ~16.7/s (was ~5), repaints ~53/s (was ~10).

`N files` count shows `0` on APFS is pre-existing + orthogonal (clonefile fast path) — separate follow-up.

### DIFF
--- a/vendor/aube/crates/aube/src/progress/mod.rs
+++ b/vendor/aube/crates/aube/src/progress/mod.rs
@@ -89,14 +89,34 @@ const TTY_PHASE_W: usize = {
 /// the spinner + live count reveal.
 const TTY_REVEAL_DELAY: Duration = Duration::from_millis(300);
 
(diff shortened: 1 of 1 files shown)


colinhacks:

PR #338, opened 2026-07-06
fix(gvs): host-filter the prewarm graph so the shared store doesn't over-split @parcel/core

colinhacks:
## Root cause

Parcel failed under the global virtual store with `DataCloneError` at worker-farm startup: `@parcel/core` was materialized at two byte-identical store dirs, so `parcel` and `@parcel/workers` loaded different `@parcel/core` instances → two serializer registries → structured-clone failure across the worker boundary.

The resolver widens its graph with every platform's optional native deps for a portable lockfile; the link phase runs `filter_graph` to host-only before hashing. The GVS prewarm materializer hashed the **widened** graph instead. Any package whose subtree carries a platform-specific optional native dep (Parcel's whole tree: `@parcel/watcher-*`, `@swc/core-*`, `lmdb`, `msgpackr`) then hashed differently in the two phases, so the same `dep_path` landed at two store dirs; the existence-gated link step never rewired over the prewarm cohort. Reproduced on a pristine store: 206 dirs vs 158 distinct (48 dups = Parcel's native subtree).

## Fix

Host-filter the prewarm graph to match the link phase. `filter_graph` is the only transform between prewarm-spawn and link that changes a node's graph hash (the intervening mutations touch only lockfile metadata the hash ignores), so both phases now hash the identical host-only graph and every `dep_path` resolves to one store dir. `prewarm_host_filter_inputs` shares the filter inputs across both prewarm sites and the pre-link filter so they can't drift. The filter sits after the non-GVS early return, so aube's default non-GVS path is unchanged.

`parcel` is dropped from `disableGlobalVirtualStoreForPackages`.

## Verification

- One `@parcel/core` dir + `parcel build` exit 0 across 2.9.3, 2.10.3, 2.11.0, 2.12.0, 2.13.3, 2.16.4.
- Hermetic test `gvs_prewarm_and_link_agree_only_after_host_filtering_the_widened_graph` pins the graph-hash agreement.
- E2E harness under `tests/parcel-gvs/`.
- `next` hit the same over-split and is also fixed, but its Turbopack chroot break is independent, so it stays blocklisted pending separate verification.

## Fork discipline

Default-preserving on aube's non-GVS default; corrects the GVS path (a latent aube bug). Changes shared-store dir names for affected installs — a one-time re-materialization; old dirs are orphaned, not corrupted.

### DIFF
--- a/crates/nub-cli/src/pm_engine/mod.rs
+++ b/crates/nub-cli/src/pm_engine/mod.rs
@@ -2162,17 +2162,25 @@ fn nub_setting_defaults(
             // NOT a phantom-dep class (those the dynamic detector + closure now
             // eject per-package). `nuxt` was DROPPED: its walk-up is only 2 phantom
             // edges (~2.3% closure), so it's closure-ejectable at rung 1 —
-            // symlink-GVS works (see [[nuxt-gvs-unlock]]). What remains:
+            // symlink-GVS works (see [[nuxt-gvs-unlock]]). `parcel` was DROPPED: its
+            // real break was never `.parcelrc` walk-up but a GVS store-dir over-split
+            // — the prewarm hashed the widened (all-platform) graph while the link
+            // phase hashed the host-filtered one, so parcel's native subtree
+            // materialized `@parcel/core` at two byte-identical store dirs; two
+            // consumers loaded different copies → two serializer registries →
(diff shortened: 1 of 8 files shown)


colinhacks:

PR #339, opened 2026-07-06
fix(install): drop the fetch-phase ETA from the nub TTY progress line

colinhacks:
The nub-only animated progress line's package-count ETA is dishonest on real trees: installs end on a tail of a few large native binaries (esbuild, @next/swc, sharp), so it prints absurd sub-second values (measured `ETA 14ms` while seconds remained). A bytes/rate estimate is no better — the download-size estimate runs ~19x low mid-fetch. Drop it; count, bytes, rate, and spinner already convey progress.

Removes the `eta` prop, `refresh_eta`, and `completed_at_fetch_start` from the TTY path. Standalone aube's CI renderer (`tty_progress=false`) is untouched.

Gates green: clippy, fmt, `test -p aube progress` (24/24); e2e cold zod install shows zero ETA, rest intact.


colinhacks:

PR #340, opened 2026-07-06
fix(runtime): async-tier registration when composing with a foreign loader on Node 22.15–24.11.0

colinhacks:
Fixes a crash: `nub run <tsx-script>` on Node 22.15.0–24.11.0 throws `ERR_METHOD_NOT_IMPLEMENTED: resolveSync`. nub's sync `module.registerHooks` forces synchronous resolution; a foreign async loader (tsx/ts-node) is then reached via its `resolveSync` stub, unimplemented on that Node band (fixed in Node 24.11.1; nodejs/node#59666).

Fix: on that band, when the child hosts a foreign async loader (`tsx`/`ts-node` basename, or `--import`/`--loader`), the launcher signals the preload to register hooks via the async loader-worker tier instead of sync `registerHooks`. Per-process — other runs keep the fast tier.

Verified: real tsx on 22.16 runs; unit + e2e tests; clippy/fmt/version_tiers pass. Full detail in the commit body.

### DIFF
--- a/crates/nub-cli/src/cli.rs
+++ b/crates/nub-cli/src/cli.rs
@@ -3518,6 +3518,14 @@ fn build_script_command(
     // `node` was resolved above (its path fed npm_node_execpath).
     let nub_binary = nub_core::node::spawn::current_nub_binary()?;
     let pnp_ctx = nub_core::pnp::detect(&project.root);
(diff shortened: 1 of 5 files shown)


colinhacks:

PR #346, opened 2026-07-07
fix(pm): append store dir to Vite fs.allow instead of skip-if-set (VitePress)

colinhacks:
## The bug

nub's Vite symlink-GVS compat (`crates/nub-cli/src/pm_engine/vite_compat.rs`) codegen-patches an ejected Vite `< 8.1` copy to append the machine-global virtual-store dir onto Vite's `server.fs.allow` serving allow-list, so a store-resident `/@fs` module is not rejected with `403 … outside of Vite serving allow list`.

The Vite 5 patch site anchored at `allowDirs = [searchForWorkspaceRoot(root)];`, which in Vite 5's dist lives **inside** `if (!allowDirs) {…}`. A framework that sets its own `server.fs.allow` — VitePress hardcodes `[DIST_CLIENT_PATH, srcDir, searchForWorkspaceRoot(cwd)]` — makes that block never execute, so the appended sniff never ran and the store `/@fs` module stayed `403`. Confirmed failing on VitePress@1.6.4 (pins Vite 5).

## The fix

Append the store dir to whatever `fs.allow` resolved to — the framework's array if it set one, else Vite's default — matching Vite 8.1's native `.modules.yaml` handler, which pushes the `virtualStoreDir` unconditionally (`allowDirs.push(virtualStoreDir)`, no `!allowDirs` gate).

- Anchor at the unconditional declaration: `let allowDirs = server.fs?.allow;` (Vite 5) / `let allowDirs = server.fs.allow;` (Vite 6/7). The two forms are mutually exclusive across versions.
- The injected sniff prepends `if(!allowDirs)allowDirs=[__wr];` so it defaults `allowDirs` when unset (Vite 5's own default line then becomes a harmless no-op), then appends the store dir unconditionally.
- The `.map(resolvedAllowDir)` normalization line is not a usable anchor — the bundler mangles its loop var (`i` → `i$1`) and indentation across builds.
- Vite `>= 8.1` (native) already appends unconditionally, so Unit A (`.modules.yaml`) needs no change.

## Verification

Driver harness (`tests/vite-compat/driver.sh`) against real, freshly-installed projects (dev server run bare, no nub in the process):

| case | vite | tier | `/@fs` | 403 in log | build |
|---|---|---|---|---|---|
| VitePress@1.6.4 (framework sets `fs.allow`) | 5.4.21 | backport | **200** | no-403 | ok |
| create-vite + custom `server.fs.allow` | 8.1.3 | native | 200 | no-403 | ok |
| bare vite@5 (unset `fs.allow`, regression) | 5.4.21 | backport | 200 | no-403 | ok |

CAS store verified pristine (the shared-store vite dist is never patched — only the project-local ejected copy). `cargo clippy --all-targets --all-features -- -D warnings` and `cargo fmt --check` clean; `vite_compat` unit tests (9, incl. a new v5-anchor test) pass.

Refs #315

### DIFF
--- a/crates/nub-cli/src/pm_engine/vite_compat.rs
+++ b/crates/nub-cli/src/pm_engine/vite_compat.rs
@@ -21,8 +21,10 @@
 //!   the majority of installed Vite, so for < 8.1 nub disk-materializes just the
 //!   `vite` package project-local (the linker's `diskMaterializePackages` path —
 //!   the shared CAS store stays pristine, only the local ejected copy is touched)
-//!   and codegen-inserts the sniff at Vite's own `fs.allow`-default computation
-//!   site in the bundled (non-minified) dist. That site is upstream of
+//!   and codegen-inserts the sniff at Vite's own `allowDirs` declaration in the
+//!   bundled (non-minified) dist, APPENDING the store dir to whatever `fs.allow`
+//!   resolved to — the framework's own array if it set one, else Vite's default —
+//!   matching Vite 8.1's native unconditional append. That site is upstream of
 //!   `createServer`, so it covers a bare `vite dev` CLI as well as
 //!   library-embedded Vite (Astro/SvelteKit/Nuxt). The inserted sniff is
 //!   YAML-tolerant (JSON first, block-YAML regex fallback) and reads whatever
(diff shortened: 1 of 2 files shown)


colinhacks:

PR #347, opened 2026-07-07
fix(install): install Nuxt project-local to resolve its optional-peer gap

colinhacks:
Nuxt 4 crashed at `nub install`: `nuxt prepare` → `Cannot find package '@vue/compiler-sfc'`. Nuxt's `vue-router@5` declares it an optional peer and unguardedly imports it; nub's resolver doesn't thread the peer, the scanner treats an optional peer as not a hard phantom (eject skips it), and symlink-GVS has no store-root public-hoist to catch it. Full cause in the commit.

Fix: add `nuxt` to `disableGlobalVirtualStoreForPackages` so it installs project-local, whose `.nub/node_modules/` hidden hoist tree resolves the peer. Same as `next`/`react-native`.

Verified: install exit 0, types generated, `nuxt dev` serves 200 with SSR markup. Gates green; self-review: ship.

### DIFF
--- a/crates/nub-cli/src/pm_engine/mod.rs
+++ b/crates/nub-cli/src/pm_engine/mod.rs
@@ -2160,9 +2160,26 @@ fn nub_setting_defaults(
             // The whole-install GVS-off last resort — reserved for a store-LOCALITY
             // break (a tool whose resolver can't reach the machine-global store),
             // NOT a phantom-dep class (those the dynamic detector + closure now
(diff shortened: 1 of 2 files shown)


colinhacks:
Superseded by #352 (shipped in v0.4.1): the collective project-local hidden tree became the sole phantom mechanism, and the framework matrix confirms Nuxt 4 resolves under the global virtual store without excluding it. This fallback would re-add nuxt to the GVS-exclusion list, disabling GVS for Nuxt unnecessarily. Closing as obsolete.

PR #349, opened 2026-07-07
fix(install): exempt --lockfile-only from the CI auto-frozen default

colinhacks:
Under `CI`, an install auto-selects frozen mode, which errors on lockfile drift — making `--lockfile-only` (whose job is to regenerate the lock) fail with `ERR_NUB_OUTDATED_LOCKFILE` on the drift it exists to resolve.

The CI auto-default now gates on `is_ci() && !lockfile_only`, mirroring pnpm's `opts.ci && !opts.lockfileOnly`. Plain CI installs stay auto-frozen; explicit `--frozen-lockfile` still forces frozen.

Verified against pnpm 10.15.1 on an identical drifted fixture: `--lockfile-only` rc=0 (re-resolves), plain install frozen-errors — both match. Adds a network integration test + a unit test. Embedder-agnostic, so not profile-gated.

### DIFF
--- a/crates/nub-cli/tests/install_engine.rs
+++ b/crates/nub-cli/tests/install_engine.rs
@@ -53,6 +53,24 @@ fn run_install(dir: &Path, args: &[&str]) -> (String, String, i32) {
     )
 }
 
+/// Like [`run_install`], but with `CI=true` in the environment so the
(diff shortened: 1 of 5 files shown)


colinhacks:

PR #351, opened 2026-07-07
feat(env): select .env mode from --env > APP_ENV, drop NODE_ENV (breaking)

colinhacks:
Env-file mode (which `.env.{mode}` files load) is now resolved `--env <mode>` (new flag) > `APP_ENV` (non-empty). Neither set → no mode (only `.env.local` + `.env`).

**Breaking (v0.x):** `NODE_ENV` is no longer a mode source. nub treats it as the application's own dev/prod variable — never sets it, never reads it to select env files, won't let a `.env` change it. Migrate an ambient `NODE_ENV=<mode>` to `APP_ENV=<mode>` (or `--env <mode>`). A non-dev/prod/test `NODE_ENV` used as a file selector silently flips frameworks into development — the footgun Vite/varlock/dotenvx/Node/Deno all decoupled.

One source — `env_file_names`/`resolve_mode` in `crates/nub-core/src/workspace/env.rs` — shared by the file-run and watch paths. `--env` (captured in a `OnceLock`) coexists with `--env-file <path>`. Mode is restricted to `[A-Za-z0-9_.-]`, closing a traversal vector.

Tests cover the `--env > APP_ENV` precedence, `.env.{mode}` ordering + `test`-skip, and the guard. `env.mdx` updated.

### DIFF
--- a/crates/nub-cli/src/cli.rs
+++ b/crates/nub-cli/src/cli.rs
@@ -168,6 +168,27 @@ fn env_file_flag_present() -> bool {
     ENV_FILE_PRESENT.get().copied().unwrap_or(false)
 }
 
+/// The explicit env-file *mode* from `--env <mode>`, captured once at startup —
+/// the highest-precedence mode source (over `APP_ENV`/`NODE_ENV`) that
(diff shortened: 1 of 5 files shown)


colinhacks:

PR #350, opened 2026-07-07
fix(pm): hoist satisfied optional peers into ejected closure members (GVS)

colinhacks:
Nuxt 4 crashes under the global virtual store: `nuxt prepare` fails `Cannot find package '@vue/compiler-sfc'`. `vue-router@5`'s `/vite` statically imports `@vue/compiler-sfc` but declares it an optional peer, so nub does not thread or hoist it, and under GVS the ejected `vue-router` copy cannot reach it though it is present in the tree.

Fix: rung-2b in the disk-materialize planner hoists each ejected closure member's optional peers that are present, range-satisfying, and not already linked into the member's own `node_modules` — matching pnpm. Iterates only the existing closure; no change to the store, dep_paths, or lockfile.

Verified: Nuxt 4 under GVS installs, `nuxt prepare` and `nuxt dev` (200 + SSR) pass. Corpus results below.

Refs #315
Refs #347

PR #352, opened 2026-07-07
feat(linker): collective hidden tree as the sole phantom mechanism under GVS (replaces per-importer hoist) + flatten Materialization

colinhacks:
## Collective project-local hidden hoist tree — the sole phantom mechanism under GVS

Under the global virtual store, most packages are symlinks into the machine-global shared store, so their realpath escapes the project. The disk-materialized (ejected) subset is the exception: those are real project-local directories, so Node's upward `node_modules` walk from inside an ejected package passes through `node_modules/.nub/node_modules/`. Before this change that tree was always swept under GVS, so an ejected package could only resolve an undeclared phantom that a per-importer mechanism specifically detected and placed as an extra sibling.

This builds one collective project-local hidden hoist tree at `node_modules/.nub/node_modules/` over the whole graph (blanket, first-write-wins by name — pnpm's undeclared-import version choice) whenever GVS is active and the ejected set is non-empty. Any ejected package then resolves any undeclared phantom via the walk-up — detection-free.

Because the collective tree covers every undeclared phantom for the ejected set, the per-importer hoist is now redundant, and this PR **removes it entirely** — it is the sole phantom-resolution mechanism under GVS.

### What this removes

- The per-importer phantom-hoist (`hoist_within`) — both the scanner-flagged undeclared-target rung and the optional-peer rung added in #350. The linker's `phantom_hoist` field, `with_phantom_hoist`, `phantom_hoist_for`, and the sibling-injection in the disk-materialize branch all go; `DiskMaterializePlan` carries `names` only.
- In nub's disk-materialize hook (`phantom_closure.rs`), the planner now only grows the eject set (the ancestor-closure). The rung-2 target map, the rung-2b optional-peer loop, and the `resolve_target_dep_path` / `resolve_satisfying_dep_path` helpers are gone.

### Materialization enum flattened

The materialization decision was `Symlink | Disk { hidden_tree: bool }` — a nested bool. It is now three flat sibling variants: `Symlink | DiskWithHiddenTree | DiskNoHiddenTree`. The illegal shared-store-with-hidden-tree state stays unrepresentable (there is no `Symlink`-with-tree variant), without the nesting. `resolve()`, `uses_shared_store()`, and `build_hidden_tree()` keep their exact semantics.

### Why it is safe

- **Not aube issue #6.** #6 forbids the hidden tree at the shared machine-global store root, where a bare-name alias is cross-project-mutable. This tree is project-local — its aliases are per-project, so that collision cannot occur. The shared-store tree is still purged.
- **Leak-free.** Only ejected (project-local-realpath) packages reach the tree on their walk-up. A symlinked package's realpath is in the shared store, so its walk ascends the store and never the project — it can never consume the tree. Verified empirically (vue-ts: all 5 ejected realpaths project-local; all 44 symlinked realpaths escape the project).
- **No worse than pnpm.** pnpm's GVS-off hidden tree is this same blanket tree, on the walk-up of every package. Here only the ejected subset sees it, so the behavior is a strict subset of pnpm's.
- **Standalone aube unchanged.** `disk_materialize` is empty for standalone aube and every existing test, so the new branch is unreachable there — the pass is byte-for-byte identical.

### Verification

Nuxt 4 under GVS, with the per-importer mechanism removed: `nub install` rc=0, `nuxt prepare` rc=0, `nuxt dev` HTTP 200 (SSR). vue-router's own `@vue/compiler-sfc` sibling is absent (mechanism gone); `.nub/node_modules/@vue/compiler-sfc` is present and used — the exact case #350 patched per-importer, now covered by the collective tree alone.

Framework corpus, all green with the per-importer mechanism removed, zero regressions:

| project | install | dev | build | preview | collective tree |
|---|---|---|---|---|---|
| nuxt4 | ok | 200 | ok | — | built |
| vue-ts | ok | 200 | ok | 200 | 33 entries |
| vitepress | ok | 200 | ok | — | 77 entries |
| sveltekit | ok | 200 | ok | — | 47 entries |
| react-ts | ok | 200 | ok | 200 | none (ejected set empty) |
| astro-min | ok | 200 | ok | 200 | none (ejected set empty) |

The tree is a strict no-op where nothing ejects (react-ts, astro-min).

Refs #350.

### DIFF
--- a/crates/nub-cli/src/pm_engine/phantom_closure.rs
+++ b/crates/nub-cli/src/pm_engine/phantom_closure.rs
@@ -7,22 +7,26 @@
 //! it — otherwise a store-resident importer keeps resolving the un-materialized
 //! shared-store copy, a silent singleton split (two realpaths, two module
 //! instances). This hook expands aube's flat disk-materialize seed into a
-//! graph-aware plan against the resolved lockfile:
+//! graph-aware plan against the resolved lockfile — the ancestor-closure
+//! (rung 1). Each seed grows to [`LockfileGraph::importer_closure`] — the seed
+//! UNION every package that transitively imports it. Bounded to the affected
+//! subtree by construction (unrelated top-level subtrees are not importers),
+//! measured 0.3–2.1% of real large trees. Also SUBSUMES the #315
+//! library-embedded-vite<8.1 residual: an embedded vite<8.1 (a framework's
+//! transitive engine, no direct-dep symlink) is auto-detected and its
+//! `[framework…vite]` closure ejected, so #318's dist sniff patch reaches a
+//! now-project-local vite.
 //!
-//! - **Rung 1 — ancestor-closure.** Each seed grows to
-//!   [`LockfileGraph::importer_closure`] — the seed UNION every package that
-//!   transitively imports it. Bounded to the affected subtree by construction
-//!   (unrelated top-level subtrees are not importers), measured 0.3–2.1% of real
-//!   large trees. Also SUBSUMES the #315 library-embedded-vite<8.1 residual: an
-//!   embedded vite<8.1 (a framework's transitive engine, no direct-dep symlink)
-//!   is auto-detected and its `[framework…vite]` closure ejected, so #318's dist
-//!   sniff patch reaches a now-project-local vite.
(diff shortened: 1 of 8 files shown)


colinhacks:

PR #354, opened 2026-07-07
feat(env): remove --env flag; mode is APP_ENV only; --env-file repeatable (last-wins)

colinhacks:
The `--env <mode>` flag (shipped in 0.4.0) is removed. Env-file mode is now selected by `APP_ENV` alone; to load a specific file, pass `--env-file <path>` (Node's own flag), now repeatable.

Removed the `Watch --env` arg, the `ENV_MODE_OVERRIDE` OnceLock, the manual `--env` arms, and the `mode_override` threading through the env loaders. The charset guard and `.env` cascade are unchanged.

Multiple `--env-file` now matches Node — subsequent files override earlier ones (last-wins; was first-wins); shell env still wins.

Tests: `resolve_mode` is APP_ENV-only; new test asserts last-wins. Fast-follow on #351.


colinhacks:

PR #355, opened 2026-07-07
feat(pm): accept `update <pkg>@<version>` and `add --lockfile-only` (pnpm-compat)

colinhacks:
## Problem

nub's PM CLI rejected two valid pnpm grammars, both surfaced by the Dependabot ecosystem work (which pins a dependency to an exact version and refreshes only the lockfile):

- `nub update <pkg>@<version>` — the aube engine hard-errored any non-`latest` spec (`package spec 'is-odd@3.0.1' is not supported by 'update'`).
- `nub add <pkg>@<version> --lockfile-only` — `--lockfile-only` was absent from aube's `AddArgs` (`unexpected argument '--lockfile-only'`).

Both are accepted by real pnpm. This closes the gap in nub's own PM CLI frontend.

## Behavior, matched to pnpm empirically

`update <pkg>@<version>` glues the manifest's **existing** range operator onto the resolved version — the arg's own operator is discarded, and out-of-range downgrades keep the operator. Verified against pnpm 10.15.1 on identical fixtures:

| from | command | pnpm | nub |
|------|---------|------|-----|
| `^3.0.0` | `up is-positive@3.1.0` | `^3.1.0` | `^3.1.0` |
| `3.1.0` | `up is-positive@3.0.0` | `3.0.0` | `3.0.0` |
| `^3.1.0` | `up is-positive@3.0.0` (downgrade, out of range) | `^3.0.0` | `^3.0.0` |
| `^3.0.0` | `up is-positive@~3.1.0` (arg operator discarded) | `^3.1.0` | `^3.1.0` |
| `^3.0.0` | `up is-positive@3.1.0 --lockfile-only` | `^3.1.0` | `^3.1.0` |

## Implementation

- **update** — generalizes the existing `@latest`/`--latest` machinery to any concrete version or range. The arg spec is injected as the resolver range for that entry, then the manifest rewrite preserves the existing operator (the same `rewrite_specifier` path `@latest` already used). The rejection now only blocks specs with no registry semver target (empty, `workspace:`, `catalog:`, git). `update --lockfile-only` was already wired in aube, so `update <pkg>@<version> --lockfile-only` now works end-to-end.
- **add** — threads a `--lockfile-only` flag through `AddArgs` into the install pipeline so the linker is skipped while the lockfile + manifest are written.

Fork-discipline: this is a pnpm-compat fix to the vendored aube engine; the `--latest`/`@latest`/bare-`update` paths keep their exact prior behavior (byte-identical output, including the "already latest" wording).

## Tests

- aube unit tests: operator-preservation across caret/tilde/exact/downgrade/`-E`/alias; the reject filter accepts versions/ranges/`latest`, rejects empty/`workspace:`/`catalog:`/git.
- nub e2e (`pm_verbs.rs`, network `#[ignore]` per convention): `up <pkg>@<version>` preserves the caret and resolves the lockfile; `add <pkg>@<version> --lockfile-only` writes the lockfile without linking `node_modules`. Both pass against the live registry.
- Gates green locally: `cargo fmt --check`, `cargo clippy --all-targets --all-features -- -D warnings` (aube + nub-cli), the new tests.

## Docs

`install/index.mdx` gains the two grammar forms under `nub add` and `nub update`.

Refs the nub-update-pin-version-grammar thread.

### DIFF
--- a/crates/nub-cli/tests/pm_verbs.rs
+++ b/crates/nub-cli/tests/pm_verbs.rs
@@ -278,10 +278,120 @@ fn update_latest_moves_the_manifest_and_lockfile_forward() {
         manifest.contains("3.1.0"),
         "--latest must rewrite the manifest past the pin: {manifest}"
     );
-    let lock = std::fs::read_to_string(dir.join("pnpm-lock.yaml")).unwrap();
+    // Fresh project with no incumbent lockfile → nub writes its neutral
+    // nub.lock; a project already on pnpm-lock.yaml keeps that.
+    let lock = std::fs::read_to_string(dir.join("nub.lock"))
+        .or_else(|_| std::fs::read_to_string(dir.join("pnpm-lock.yaml")))
+        .expect("update must write a lockfile");
     assert!(
         lock.contains("3.1.0") && !lock.contains("is-positive@3.0.0"),
         "the lockfile must resolve the updated version: {lock}"
+    );
+}
+
+/// `nub up <pkg>@<version>` pins the named dep to that version, preserving
+/// the manifest's existing range operator (`^3.0.0` + `is-positive@3.1.0` ->
+/// `^3.1.0`) — matching `pnpm update <pkg>@<version>`. The pre-fix engine
+/// rejected any non-`latest` spec here.
+#[test]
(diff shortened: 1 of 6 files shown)


colinhacks:

PR #356, opened 2026-07-07
fix(env): ignore process-control vars (NODE_OPTIONS, ...) from .env files, matching Deno

colinhacks:
A `.env` configures the user's program, not the runtime nub launches. This ignores four Node control vars from a `.env` / `--env-file` source, mirroring Deno's [`ENV_FILE_DENYLIST`](https://github.com/denoland/deno/blob/main/cli/util/env.rs): `NODE_OPTIONS`, `NODE_TLS_REJECT_UNAUTHORIZED`, `NODE_EXTRA_CA_CERTS`, `NODE_REPL_EXTERNAL_MODULE`.

Filtered at load in `env.rs`, covering every consumer plus the `--env-file` path. Only env-file-sourced values are dropped; ambient values and nub's own `NODE_OPTIONS` are unaffected. Tradeoff: `NODE_OPTIONS=--max-old-space-size` in a `.env` moves to the shell/CI env.

Verified: `workspace::env` tests (2 new), e2e, fmt + clippy.

### DIFF
--- a/crates/nub-cli/src/cli.rs
+++ b/crates/nub-cli/src/cli.rs
@@ -1372,6 +1372,14 @@ fn run_nub() -> Result<i32> {
     // the expansion load_env_files applies on the non-watch run path.
     nub_core::workspace::env::expand_env_map(&mut env_file_vars);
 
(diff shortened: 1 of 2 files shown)


colinhacks:

PR #358, opened 2026-07-07
pm: rename project-local virtual store to node_modules/.store

colinhacks:
Renames the project-local isolated virtual-store directory from `node_modules/.nub` to `node_modules/.store`, and stops writing an empty applied-patches sidecar.

## Why `.store`

`.store` is the vendor-neutral isolated-store convention — npm's own isolated-mode RFC-0042, Yarn-berry's pnpm-linker (`pnpmStoreFolder`), and cnpm all default to `node_modules/.store`. Tools that walk `node_modules` to find a project root recognize it as an install marker; `.nub` was invisible to them. It is also brand-cleaner.

The concrete fix is the simple-git-hooks class. simple-git-hooks looks for a `.store` (or `.pnpm`/`.deno`) path segment to locate the project root in its postinstall — it does not know `.nub`. Under a project-local store the package lands at `node_modules/.nub/simple-git-hooks@2.13.1/node_modules/simple-git-hooks`, so the old resolver fell through to its "ends in `node_modules/<pkg>`" branch and returned the store slot dir as the root, then crashed with `ENOENT` reading a `package.json` that isn't there. With `.store` the marker resolves to the real project root:

```
OLD .nub  -> resolved root: <proj>/node_modules/.nub/simple-git-hooks@2.13.1   (no package.json -> ENOENT)
NEW .store -> resolved root: <proj>                                            (correct)
```

## Empty applied-patches sidecar

A patch-free install wrote `node_modules/.nub-applied-patches.json` containing `{}` — information-free clutter, since a missing file already reads back as the empty map. `write_applied_patches` now skips the write when the patch set is empty and removes a stale sidecar when the last patch is dropped. This is correct for standalone aube too (an empty manifest carries nothing).

## Scope

- Project-local store only. The machine-global store (`~/.cache/nub/pm/virtual-store/`) is unchanged — a separate, larger blast radius, and it isn't under the project so it doesn't affect the store-marker case.
- The nub-branded state filenames (`.nub-state`, `.nub-applied-patches.json`) are embedder-name-derived, not store-marker directories, so they keep the nub brand; only the containing store dir renamed. The install state now lives at `node_modules/.store/.nub-state/`.
- Standalone aube's default (`.aube`) is unchanged — this is a nub-embedder value. The store-dir name has a single source of truth (`PROJECT_VIRTUAL_STORE_LEAF`), consumed by the setting defaults, the vite-compat scan, and the runtime phantom-dep detector.

## Verification

- `nub install` on a fresh fixture lands the store at `node_modules/.store/`, with top-level symlinks pointing into it and no applied-patches sidecar.
- simple-git-hooks fixture (project-local store): `nub install` exits 0 with no ENOENT; run from the store package dir the postinstall resolves the correct project root and installs the hook.
- Under the default global virtual store the store entry's realpath is the machine-global store (outside the project), so the marker is out of reach there regardless of its name — the fix lands for the project-local store (CI / `nub ci` / GVS-off), which is where postinstall root-resolution runs.
- Gates green: `cargo fmt --check`, `clippy --all-targets --all-features -D warnings`, updated unit + e2e tests (store-defaults, rewrite passthrough, phantom-dep hint, a new empty-sidecar test, and an install assertion that no sidecar is written).

### DIFF
--- a/crates/nub-cli/src/pm_engine/info_family.rs
+++ b/crates/nub-cli/src/pm_engine/info_family.rs
@@ -79,7 +79,7 @@
 //!   spellings rebrand ("`aube outdated -w`" reads "`nub outdated -w`") and
 //!   config-location spellings map to nub's configured contract
 //!   (`aube-workspace.yaml` → `pnpm-workspace.yaml`, `why`'s
-//!   `.aube/<dep_path>` example → `.nub/<dep_path>`).
+//!   `.aube/<dep_path>` example → `.store/<dep_path>`).
 //! - `outdated` / `audit` / `peers check` signal "findings exist" via
 //!   `std::process::exit(1)` *inside* the engine (pnpm-compat), after the
 //!   report is fully printed — they bypass this file's `Result<i32>` return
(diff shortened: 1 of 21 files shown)


colinhacks:

PR #359, opened 2026-07-07
fix(pm): scan warm-cached phantom packages on-demand at link time (GVS)

colinhacks:
Nuxt 3 (and any phantom-having package) could break under the global virtual store on a first `nub install` / `nub add` over a warm cache: `nuxt prepare` crashed with `Cannot find package 'scule'`, from `@nuxt/schema`'s unguarded top-level `import 'scule'`.

## Root cause

The undeclared-phantom eject pipeline is correct end-to-end — the scanner flags the phantom, the closure ejects the importer, the collective hidden tree resolves the phantom — but it only fires when a per-content **phantom sidecar** exists at link time, and the two sidecar producers both miss one case:

- The extract hook writes a sidecar only on a genuine tarball **fetch**.
- `backfill_from_lockfile` reads only the **pre-existing lockfile** — empty on the first install/add of a project, and it runs on `install`/`ci` only.

So a package **warm in the CAS with no sidecar** (GC'd, or cached by a pre-eject-default nub — common right after upgrading, since the accumulated cache predates the sidecars) on a first install/add reaches the link-time decision with no verdict. The consumer treated a missing sidecar as "no eject", leaving the package symlinked to the shared store, so its undeclared import failed to resolve.

Both `install` and `add` reproduce it deterministically: delete the `@nuxt/schema` sidecar, then in a fresh project run `nub install` (declaring `nuxt@3.21.8`) or `nub add nuxt@3.21.8` — `@nuxt/schema` stays symlinked, `scule` is `MODULE_NOT_FOUND`, `nuxt prepare` fails.

## Fix

Make the link-time consumer (`dynamic_phantom_flags`) scan the already-loaded CAS index **on-demand** when the sidecar is missing, and cache the verdict — so the eject decision is correct regardless of extract-hook/backfill timing, at the one point where the resolved graph and the CAS index are both in hand. Covers install/add/update, first or Nth, GC'd or pre-eject-default caches.

The scan and the atomic sidecar write are factored out of `scan_and_cache` into shared helpers; `cached_or_scan_verdict` is the new read-or-scan entry.

## Verification

- Deterministic repro (delete sidecar + fresh project) now ejects `@nuxt/schema` and resolves `scule` on **both** `install` and `add`; `nuxt prepare` succeeds.
- nuxt 3.21.8 and nuxt 4.4.8 both green under GVS (cold + warm; `nuxt prepare` succeeds).
- Regression: cold isolated-store install (extract-hook path) and warm install with the sidecar present are unchanged.
- New unit test exercises the store-IO scan-on-miss path (scan + cache, then a cache-served second read).
- `cargo clippy --all-targets --all-features -D warnings` and `cargo fmt --check` clean; phantom + install-engine unit tests pass.

### DIFF
--- a/crates/nub-cli/src/dynamic_phantom.rs
+++ b/crates/nub-cli/src/dynamic_phantom.rs
@@ -41,7 +41,7 @@
 use std::path::{Path, PathBuf};
 
 use aube_store::{PackageIndex, index_content_fingerprint};
-use nub_phantom_scan::scan_index;
+use nub_phantom_scan::{ScanResult, scan_index};
 use rayon::prelude::*;
 
 /// Whether dynamic phantom detection + ancestor-closure eject is armed.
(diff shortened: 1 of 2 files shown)


colinhacks:

PR #360, opened 2026-07-08
pm: rename machine-global virtual store to store

colinhacks:
Renames nub's machine-global virtual store from `~/.cache/nub/pm/virtual-store/` to `~/.cache/nub/pm/store/`. It is the real, shared store packages materialize into — only "virtual" in that a project's `node_modules` symlink out of it — so pnpm's "virtual-store" term (which names the flattened *per-project* tree) never fit. `store` reads clean.

This is a naming change only — the shared-store behavior is unchanged.

## Fork discipline

The leaf name was a hardcoded constant in the vendored aube engine (`aube_store::VIRTUAL_STORE_SUBDIR`), so it moves onto the embedder profile as `Embedder::virtual_store_subdir`:

- Standalone aube (the `AUBE` profile) keeps `virtual-store` byte-for-byte — no behavior change on the default path.
- nub (the `NUB` profile) sets `store`.

`Store::virtual_store_dir()` and nub's vite-compat path builder both read the profile field, so the two places that compute the global-store path can never drift apart.

## Old cache is orphaned, not migrated

An existing `~/.cache/nub/pm/virtual-store/` is left in place. It is a regenerable cache: new installs write to `store`, and existing projects' absolute `node_modules` symlinks keep resolving to the still-present `virtual-store` until they next reinstall (which repopulates `store`). No migration, no re-download, no breakage; the old directory is reclaimed on the next store GC.

## Why `store`, not `.store`

`store` deliberately avoids a leading dot. simple-git-hooks (and similar postinstall tools) derive the project root by slicing `process.cwd()` at a path segment exactly equal to `.store`/`.pnpm`/`.deno`. Under the global virtual store, a dependency's lifecycle script runs with `process.cwd()` resolved into the store, so a `.store` leaf would have made those tools resolve the project root to the cache directory. A plain `store` segment is not matched, so they fall through to their generic branch exactly as they did under `virtual-store` — no behavior change for that class of tool.

## Verification

Fresh `nub install` on a throwaway fixture populates `~/.cache/nub/pm/store/`, the project `node_modules` symlinks resolve into it, with no `.store` or `virtual-store` directory created. No sibling collides with `store` under `~/.cache/nub/pm/` (the content-addressed store lives at `~/.local/share/nub/store`, a separate root). A `simple-git-hooks` differential fixture confirms its postinstall no longer resolves to the cache directory. `cargo fmt`, `cargo clippy --all-targets --all-features -D warnings`, and the touched aube + nub test suites pass.

Refs #358


colinhacks:

PR #362, opened 2026-07-08
feat(env): add --no-env-file to disable all env-file loading

colinhacks:
Adds `--no-env-file`: load zero env files. It suppresses the eager `.env*` auto-discovery **and** any explicit `--env-file` / `--env-file-if-exists`, while every other augmentation (TypeScript, JSX, the module hooks) stays on. For when the environment is already managed elsewhere — CI, a secret manager, `direnv` — and you want nub to keep its hands off.

```bash
# no .env* auto-discovery, and the --env-file is ignored — the child sees neither
nub --no-env-file --env-file=.env.ci server.ts
```

## Behavior

- **Wins over `--env-file` and `--env-file-if-exists`**, regardless of arg order — it means "load nothing, period."
- **All surfaces**: the file run (`nub <file>`), `nub run`, `nubx` (both the standalone binary and `nub x`/`dlx`), and `nub watch` — where no `.env*` file is handed to the watched Node.
- **Flag-only.** No env-var companion and no tree-wide inheritance: the same per-process boundary as `--env-file`. For a persistent, whole-tree opt-out that also disables the rest of nub's augmentation, `--node` / `NODE_COMPAT=1` remain the tools.

Like `--env-file`, the flag is a pre-subcommand option: `nub --no-env-file run build`, not `nub run --no-env-file build`.

## Implementation

`--no-env-file` is captured in the hand-parsed arg loop into a `NO_ENV_FILE` process flag. `merge_child_env` short-circuits to an empty map, `overlay_env_file_vars` / `apply_env_file_vars` early-return, and the auto-discovery load sites (file run + watch) gate on it. The standalone `nubx` binary forwards `--env-file` to Node directly (it is a Node value-flag, not captured by nub's loop), so when `--no-env-file` is present `run_nubx` strips the whole `--env-file*` family from its leading argv before Node sees it — arity-aware for the space form — so the flag wins on that surface too.

## Tests

Six committed tests: the `merge_child_env` unit contract plus five integration tests — auto-`.env` suppressed while TS still runs, precedence over `--env-file` and `--env-file-if-exists`, the standalone-`nubx` forwarding case, and the watch path.

## Docs

`site/content/docs/runtime/env.mdx` gains a "Loading nothing" section.

### DIFF
--- a/crates/nub-cli/src/cli.rs
+++ b/crates/nub-cli/src/cli.rs
@@ -2220,12 +2220,11 @@ fn run_nubx() -> Result<i32> {
     // two mechanisms are needed (clap fails-closed on a future Node flag).
     let mut args: Vec<String> = env::args().skip(1).collect();
 
-    // `--help`/`--version`/`--no-env-file` are nubx's own flags only when they
-    // appear BEFORE the subject (the three-position rule: a flag after the subject
-    // reaches the resolved runner verbatim — `nubx eslint --help` is eslint's
-    // help). When no subject has been seen yet and one of these leading flags
-    // appears, honor it like `nub --help`/`nub --version`/`nub --no-env-file`.
-    let mut no_env_file_seen = false;
+    // `--help`/`--version` are nubx's own flags only when they appear BEFORE the
+    // subject (the three-position rule: a flag after the subject reaches the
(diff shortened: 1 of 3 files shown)


colinhacks:

PR #363, opened 2026-07-08
pm: abort install on unresolvable bun.lock sources

colinhacks:
Extends #217's abort-eagerly policy (yarn) to the bun lockfile reader. Refs #217.

Under nub's `strict_unsupported_source` toggle, a `bun.lock` entry with an unrecognized source protocol fails at plan time (`ERR_NUB_LOCKFILE_UNSUPPORTED_SOURCE`) instead of silently reclassifying to a registry pin that 404s mid-install. Optional deps warn + skip, recorded so frozen/`ci` drift tolerates them. bun resolves transitive deps by name, so the edge check walks the union of parsed and withheld key sets. Standalone aube keeps its lenient default (pinned by a unit test). No real bun protocol triggers this today — defense-in-depth.

https://claude.ai/code/session_01EZAM6j11iDztC76fx61J17

### DIFF
--- a/crates/nub-cli/tests/abort_eagerly.rs
+++ b/crates/nub-cli/tests/abort_eagerly.rs
@@ -146,3 +146,78 @@ fn install_proceeds_on_optional_unresolvable_source() {
         "the optional skip should warn; got:\n{out}"
     );
 }
+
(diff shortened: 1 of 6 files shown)


colinhacks:

PR #364, opened 2026-07-08
feat(install): retire the user-facing diskMaterializePackages knob under nub

colinhacks:
## What

Retires the user-facing `diskMaterializePackages` npmrc/env/`pnpm-workspace.yaml` knob under nub. The dynamic phantom detector (#352, on by default) auto-ejects undeclared-import adapters, so a hand-listed force-eject is redundant for its documented purpose.

## Why

`diskMaterializePackages` let a user force a package project-local under the global virtual store so a phantom subpath import (`@hookform/resolvers/zod` reaching the app's `zod`) could resolve. Since #352, the detector scans each package's real published code and ejects those adapters automatically — the manual list no longer adds anything for its documented case. Verified empirically under default GVS:

- `@hookform/resolvers@3.9.1 + zod`, no knob set → `@hookform/resolvers` auto-ejected, `import('@hookform/resolvers/zod')` resolves.
- With `diskMaterializePackages=lodash` set, lodash is no longer ejected after this change (it was before).

Any residual detector miss stays covered by the coarser, retained `disableGlobalVirtualStoreForPackages`.

## How

nub's disk-materialize expansion hook (nub-side, registered only by nub) now filters the resolved seed through `nub_internal_seed()`, keeping only nub's own embedder-default names (the internal `vite` #315 seed) and dropping every user-source entry before planning.

Fork-disciplined: `vendor/aube/**` is untouched. Standalone aube installs no hook, so it keeps honoring the setting byte-for-byte. The internal `vite` embedder default is kept for the phantom-eject opt-out (no-hook) path. Docs drop the `diskMaterializePackages` instruction; `disableGlobalVirtualStoreForPackages` remains the documented escape hatch.

## Test

- New unit test asserts the hook drops user-source seed names, keeping only the internal `vite`.
- Existing vite / phantom-closure unit tests unchanged and green (14 total).
- fmt + clippy (`-D warnings`) clean; e2e verified on a dev binary.

### DIFF
--- a/crates/nub-cli/src/pm_engine/mod.rs
+++ b/crates/nub-cli/src/pm_engine/mod.rs
@@ -2129,10 +2129,11 @@ fn nub_setting_defaults(
     // ejects them through #319's ancestor-closure, subsuming the old 14-entry
     // const (incl. the singleton-hazard adapters the closure now makes sound). So
     // this embedder default carries ONLY the #315 vite eject; empty otherwise
-    // (aube's `parse_string_list` drops the empty entry). A user's own
-    // `diskMaterializePackages` still wins — the
-    // embedder default is the lowest precedence tier — so the additive escape
-    // hatch is intact.
+    // (aube's `parse_string_list` drops the empty entry). nub exposes NO
+    // user-facing `diskMaterializePackages` knob (maintainer 2026-07-07): the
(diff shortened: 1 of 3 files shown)


colinhacks:

PR #366, opened 2026-07-08
bench: prune non-owned lockfiles per tool in run-4way.sh

colinhacks:
tests/bench/install/run-4way.sh copies the full fixture (which ships all four lockfiles) into each tool's directory without pruning, so nub install fails with ERR_NUB_LOCKFILE_AMBIGUOUS. This strips the non-owned lockfiles per tool, matching the split already done in run.sh's setup_workdir.


colinhacks:

PR #367, opened 2026-07-08
bench(install): nub leg installs from native nub.lock; 5-bar GVS table + staleness guard

colinhacks:
The install-bench nub leg was compat-reading `pnpm-lock.yaml` instead of installing from nub's own lockfile. Every PM bench fixture now ships a committed `nub.lock` (native pnpm-v9 format), and each harness strips the foreign lockfiles from the nub workdir so the nub leg exercises nub's native path.

## Changes

- `nub.lock` committed for all 8 fixtures (`simple`, `monorepo`, `t3`, `tanstack-start`, `large`, `gvs-eligible`, `gvs-ineligible`, `fat`), generated by a current release nub.
- `run.sh`, `run-4way.sh`, `run-warm-gvs.sh`, `cold-cas.sh`: the nub workdir keeps `nub.lock` only; the pnpm/bun/npm legs keep their own lockfile.
- `run-4way.sh` is now a 5-bar table — nub v0.3 (GVS on, symlink-into-store relink) vs nub v0.2 (GVS off, per-project materialize; same binary, `CI` env toggles GVS) vs bun/pnpm/npm — so the GVS before/after shows in one run.
- Staleness guard in `run.sh` and `run-4way.sh`: the nub binary version must match `Cargo.toml`. A stale/missing default `target/release/nub` auto-builds the release; an explicit `NUB=` override that mismatches hard-errors. This is what let the bench silently measure a v0.2.8 binary.
- `make bench` (build-if-needed + run the table); `gen-fixtures.sh` regenerates `nub.lock` for every fixture.

## Perf verification (tanstack-start, warm, load ~14)

| bar | mean |
|---|---|
| nub v0.3 (GVS on) | 337 ms |
| bun install | 503 ms |
| nub v0.2 (GVS off) | 971 ms |
| pnpm install | 2.42 s |
| npm ci | 6.70 s |

nub v0.3 is fastest — 2.88x faster than nub v0.2. The earlier ~1.1s nub number was the stale v0.2.8 binary doing a 15,144-file per-file materialize. On the current binary the GVS-on link phase is 253 store-symlinks + 6 clonedirs (45 ms); the GVS-off path is 253 whole-dir clonefiles (395 ms) — neither is the old per-file copy.

Bench-infra only; no Rust changes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ else
   CARGO_FLAGS =
 endif
 
-.PHONY: build addon addon-fast install-dev uninstall-dev test test-node-matrix clean npm-build npm-publish npm-publish-dry
+.PHONY: build addon addon-fast install-dev uninstall-dev test test-node-matrix bench clean npm-build npm-publish npm-publish-dry
 
 build: addon
 	$(CARGO) build $(CARGO_FLAGS)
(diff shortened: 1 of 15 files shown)


colinhacks:

PR #372, opened 2026-07-08
bench(install): fix SIGPIPE killing run-4way.sh on Linux

colinhacks:
`report_nub_layout`'s `find .store … | head -1` trips SIGPIPE → `pipefail` → `set -e` when `.store` has many entries (252 in the tanstack-start fixture): `find` keeps enumerating after `head` closes the pipe, so the pipeline returns 141 and the script dies before the timed block. Surfaces on Linux; macOS timing survived.

It's untimed diagnostic output, so the numbers are unaffected — but `make bench` fails outright on Linux.

Fix: `find … -print -quit` (first match, no pipe, no `head`; BSD + GNU). Verified `bash -n` clean and the layout signal still prints with 300+ entries under `set -euo pipefail`.

Refs #367


colinhacks:

PR #374, opened 2026-07-08
fix(lockfile): enumerate yarn berry workspace-member importers

### LINKED ISSUE #373: nub install silently skips yarn berry workspace members (exit 0)
`nub install` on a yarn-incumbent monorepo with a berry (v2+) `yarn.lock` installs only the root's dependencies and skips every workspace member, exiting 0. Found on cal.com (yarn@4.12.0, 114 workspaces): `react`/`next`/`@calcom/*` absent from members' `node_modules`.

Root cause: the berry reader synthesizes only the root (`.`) importer, so a frozen install never links members. The classic (v1) reader already reconstructs members from disk; berry did not.

Repro: a two-package `workspaces` project with a committed berry `yarn.lock` — `nub install` exits 0 but the member's `node_modules` is empty. `npm`/`pnpm` install the member dep on the same fixture.

colinhacks:
Closes #373

A yarn-incumbent berry monorepo installed only the root's deps and silently skipped every workspace member, exiting 0 (cal.com: react/next absent) — the berry reader synthesized only the root importer.

It now reconstructs member importers from the root `package.json` `workspaces` globs + each member's own `package.json`, mirroring the classic reader (#154). A `yarn.lock` records only resolutions, no importer structure (unlike pnpm/npm), so both yarn readers glob disk; reading manifests keeps the dev/prod split exact. Shared helpers move to `yarn/mod.rs`.

Verified on a fixture and real cal.com (114 members enumerate). Two new berry tests.


colinhacks:

PR #384, opened 2026-07-08
docs: correct GVS auto-disable trigger list (next + react-native)

colinhacks:
The code auto-disables the global virtual store only for `next` and `react-native` (pm_engine/mod.rs:2192). Docs still listed the dropped `nuxt`/`parcel`. Fixes the install-docs bundler list and the bench README trigger line. Dated release blogs are correct point-in-time and untouched.


colinhacks:

PR #387, opened 2026-07-08
feat(env): clamped NODE_ENV fallback for .env-file mode selection

colinhacks:
Adds a clamped `NODE_ENV` fallback for `.env.{mode}` file selection. Maintainer-greenlit behavior change; partially reverses #351/#354.

## Contract

Mode resolution for `.env.{mode}` file selection:

- `APP_ENV` non-empty → mode = `APP_ENV` (primary; wins even when `NODE_ENV` is also set).
- else `NODE_ENV` exactly `development` / `production` / `test` → mode = that value (clamped — exact Next.js/Bun parity; no short forms).
- else (both unset, or a non-canonical `NODE_ENV` like `staging`) → no mode; only `.env` and `.env.local` load.

The charset guard and the `test`-mode `.env.local` skip apply to the resolved mode, so `NODE_ENV=test` skips `.env.local`, consistent with `APP_ENV=test`. A non-canonical `NODE_ENV` is silently ignored for file selection — use `APP_ENV` for arbitrary modes.

## Why

The earlier decision dropped `NODE_ENV` as a mode selector entirely. This restores it as a fallback for migration parity with Next.js and Bun (a `NODE_ENV=production` invocation loads `.env.production`), while the clamp neutralizes the arbitrary-value footgun that motivated the original drop: a non-canonical value can no longer silently flip the loaded file set.

## Changes

- `resolve_mode(app_env, node_env)` — pure, hermetically tested resolver; `resolve_env_mode()` reads both from process env. `CLAMPED_NODE_ENV_MODES = [development, production, test]`.
- Regression tests: hermetic unit tests for every contract branch, a composition test tying resolution to the file set, and the updated `dotenv_node_env_stripped_and_app_env_selects_env_file` integration test (NODE_ENV=production selects `.env.production`; NODE_ENV=staging does not).
- Docs: `site/content/docs/runtime/env.mdx` rewritten (the `## NODE_ENV` section no longer claims NODE_ENV never selects files). Stale comments in `env.rs` and `cli.rs` corrected to the APP_ENV -> clamped-NODE_ENV model.

## Verification

`cargo fmt --check`, `cargo clippy -p nub-core -p nub-cli --all-targets --all-features -- -D warnings`, and `cargo test -p nub-core` all green.

Refs #351, #354

### DIFF
--- a/crates/nub-cli/src/cli.rs
+++ b/crates/nub-cli/src/cli.rs
@@ -205,12 +205,13 @@ fn overlay_env_file_vars(env_map: &mut HashMap<String, String>) {
 /// Build the env map for a child, merging eager `.env*` auto-discovery with the
 /// explicit `--env-file` vars. The gate (the maintainer, 2026-06-15): when the user passed
 /// any `--env-file` flag, auto-discovery is **suppressed entirely** — none of the
-/// four auto files (`.env.<NODE_ENV>.local`, `.env.local`, `.env.<NODE_ENV>`,
-/// `.env`) load, and only the explicit file(s) reach the child. With no
-/// `--env-file`, the autos load as before.
+/// four auto files (`.env.<mode>.local`, `.env.local`, `.env.<mode>`, `.env`)
+/// load, and only the explicit file(s) reach the child. With no `--env-file`, the
+/// autos load as before.
 ///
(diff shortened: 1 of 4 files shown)


colinhacks:
Shipped in v0.4.4: https://github.com/nubjs/nub/releases/tag/v0.4.4

PR #390, opened 2026-07-08
perf(linker): memoize the per-package clonedir volume probe

colinhacks:
The hoisted macOS link pass called `ensure_trees_dir_then_probe` once per package, each re-running `create_dir_all` on the store `trees/` root plus `can_clonedir`'s two `metadata` + one `statfs` — all run-invariant. Memoize both (mirrors the existing `detect_strategy_cross` cache). macOS-linker-only; no-op elsewhere.

Verified on a 252-package warm frozen-offline install: byte-identical output (972 entries — path/type/mode/size/symlink-target unchanged), rc=0. 116 `aube-linker` tests pass; clippy/fmt clean.

Removes redundant per-package syscalls; measured wall delta is within host-contention noise (the ~46ms link phase is `clonefile(2)`-bound, not probe-bound).


colinhacks:

PR #392, opened 2026-07-09
fix(pm): write --save-catalog to package.json under nub identity

### LINKED ISSUE #369: `nub add --save-catalog` or `nub add --save-catalog-name` writes default catalog to ignored pnpm-workspace.yaml under nub identity
### Summary

In a project using nub identity with workspace/catalog configuration in root `package.json`, `nub add --save-catalog` can partially mutate manifests and then fail with `ERR_NUB_UNKNOWN_CATALOG`.

It appears to write the default catalog entry to `pnpm-workspace.yaml`, but under nub identity that file is not read. The dependency manifest is still updated to use `catalog:`, so the next resolve fails because the default catalog is not defined in the effective nub config.

### Environment

- nub: `0.4.2`
- project shape: root `package.json` has `devEngines.packageManager.name = "nub"`

Run from a workspace package:

### Reproduction

Given a root `package.json` like:

```json
{
  "workspaces": {
    "packages": [
      "apps/*",
      "packages/*"
    ]
  },
  "devEngines": {
    "packageManager": {
      "name": "nub",
      "version": "^0.4.2",
      "onFail": "warn"
    }
  },
  "engines": {
    "node": ">=24"
  }
}
```

```sh
nub add --save-catalog @types/node -D --filter=server
```

The command first updates the package manifest of the workspace package:

```text
Resolving @types/node@latest...
  + @types/node@26.1.0 (specifier: catalog:)
Updated package.json
```

Then it fails:
```text
ERR_NUB_UNKNOWN_CATALOG

  ╰─▶ @types/node: catalog reference `catalog:` does not resolve — catalog `default` is not defined (add it to `catalog:` / `catalogs.default:` in pnpm-workspace.yaml, or under
      `workspaces.catalog` / `pnpm.catalog` in package.json)
  help: no catalogs are defined in this workspace; add a `catalog:` block to `pnpm-workspace.yaml` or a `workspaces.catalog` entry in root `package.json`
```

It also creates/writes:

```yaml
catalog:
  '@types/node': ^26.1.0
```

to `pnpm-workspace.yaml`, but nub identity does not read that file.

### Expected behavior

Under nub identity, `--save-catalog` should write the default catalog to an effective nub-readable location, likely root `package.json#workspaces.catalog`.

Alternatively, if nub cannot safely create the catalog in the active config surface, it should fail before mutating the package manifest.

colinhacks:
Closes #369

Under nub identity, `nub add --save-catalog` wrote the catalog to a pnpm-named workspace YAML that nub identity does not read, so the resolve failed `ERR_NUB_UNKNOWN_CATALOG` leaving `package.json` on an unresolvable `catalog:`.

The write target is now identity-aware (gated on `read_branded_pnpm_config`): nub identity writes `workspaces.catalog(s)` in the root `package.json`; pnpm incumbent / standalone aube write the YAML unchanged. It is pre-flighted before the manifest is mutated, so a failed `--save-catalog` leaves no dangling `catalog:`.

Verified: 4 unit tests + e2e.

### DIFF
--- a/vendor/aube/crates/aube/src/commands/add/manifest.rs
+++ b/vendor/aube/crates/aube/src/commands/add/manifest.rs
@@ -556,6 +556,19 @@ pub(super) async fn update_manifest_for_add(
         place_dep_in_manifest(&mut manifest, pkg_name_for_manifest, specifier, &opts);
     }
 
+    // Pre-flight the catalog write target BEFORE mutating the dependent
+    // manifest. Under nub identity the catalog cannot live in the
+    // pnpm-named workspace YAML (the next resolve refuses to read it),
+    // so it must land in `workspaces.catalog` in package.json; resolving
+    // + validating the target here means a `--save-catalog` that can't
+    // land its entry fails before rewriting the manifest to an
+    // unresolvable `catalog:` reference (issue #369).
+    let catalog_target = if catalog_upserts.is_empty() {
+        None
+    } else {
+        Some(crate::commands::catalogs::resolve_catalog_write_target(cwd)?)
+    };
+
     // Write the updated package.json. Under `--no-save` callers still
     // write the mutated manifest to disk for the duration of the
     // resolver + install pipeline (both re-read from disk), then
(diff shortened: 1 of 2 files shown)


colinhacks:

PR #393, opened 2026-07-09
fix(patches): support name-only and range pnpm patch keys

### LINKED ISSUE #376: `nub` does not support pnpm patched dependencies
## Expected behavior

`nub install`  install dependencies and apply patch.

## Current behavior

```console
  × invalid patch key "sonda": missing version
```

Minimal reproduction case https://github.com/ghoullier/nub-patched-dependencies-issue

See this workflow run https://github.com/ghoullier/nub-patched-dependencies-issue/actions/runs/28929499834

## Dev Env Version

```console
> nub -v
v0.4.2
» node v24.18.0 (resolved from .nvmrc)
```

```console
> uname -mprs 
Darwin 24.5.0 arm64 arm
```

colinhacks:
`nub install` rejected any pnpm `patchedDependencies` key that wasn't exact `name@version`, hard-erroring at parse time:

```
× invalid patch key "sonda": missing version
```

A name-only key — "patch every resolved version of `sonda`" — is valid pnpm. pnpm accepts four forms, grouped by package name and resolved per package (`getPatchInfo`): exact `name@1.2.3`, range `name@>=1`, wildcard `name@*`, and bare `name`.

## Change

- New `aube_lockfile::patch_groups` module ports pnpm's `groupPatchedDependencies` + `getPatchInfo`: classify each declared key, then resolve each concrete resolved `name@version` to the patch that applies (priority: exact > single satisfying range > all).
- The verbatim source key stays the lockfile identity — pnpm records it unresolved in the `patchedDependencies` block. The concrete `name@version` is used only where a version is required: the linker apply, the graph-hash fold, the `(patch_hash=…)` dep-path suffix, the no-churn write guard, and the yarn/berry writer.
- Two ranges matching one resolved version raise `ERR_NUB_PATCH_KEY_CONFLICT`; an invalid non-`*` range raises `ERR_NUB_PATCH_NON_SEMVER_RANGE`, matching pnpm's messages.
- Exact-only keys resolve to their own `name@version`, so an all-exact project's install, lockfile bytes, and virtual-store paths are unchanged.

## Verification

Against real pnpm 10.15 on a name-only-key fixture:

- `nub install` succeeds and applies the patch; the `patchedDependencies` block and the `name@version(patch_hash=…)` dep-path suffix are byte-identical to pnpm's.
- Real pnpm accepts nub's generated lockfile under `--frozen-lockfile`.
- Range (`name@>=3`), wildcard (`name@*`), conflict, and invalid-range cases match pnpm's behavior and error messages.

Known limitation: converting a project that declares a non-exact patch key to `bun.lock` records the key unresolved (bun supports only exact keys); bun-incumbent projects are unaffected.

Closes #376

### DIFF
--- a/vendor/aube/crates/aube-codes/src/errors.rs
+++ b/vendor/aube/crates/aube-codes/src/errors.rs
@@ -62,6 +62,8 @@ pub const ERR_AUBE_GIT_ERROR: &str = "ERR_AUBE_GIT_ERROR";
 // ── linker ──────────────────────────────────────────────────────────
 pub const ERR_AUBE_LINK_FAILED: &str = "ERR_AUBE_LINK_FAILED";
 pub const ERR_AUBE_PATCH_FAILED: &str = "ERR_AUBE_PATCH_FAILED";
+#[rustfmt::skip] pub const ERR_AUBE_PATCH_KEY_CONFLICT: &str = "ERR_AUBE_PATCH_KEY_CONFLICT";
+#[rustfmt::skip] pub const ERR_AUBE_PATCH_NON_SEMVER_RANGE: &str = "ERR_AUBE_PATCH_NON_SEMVER_RANGE";
 pub const ERR_AUBE_MISSING_PACKAGE_INDEX: &str = "ERR_AUBE_MISSING_PACKAGE_INDEX";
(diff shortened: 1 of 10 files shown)


colinhacks:

PR #395, opened 2026-07-09
Defer to native import-text on Node 26.5+

colinhacks:
Node 26.5.0 added `--experimental-import-text` ([nodejs/node#62300](https://github.com/nodejs/node/pull/62300)) for `import x from './f' with { type: "text" }`. nub already polyfills text imports; this defers to native `textStrategy` where it exists.

- feature_matrix: `import-text` Unflag `[26.5.0, ∞)` → inject the flag; stripped below the floor.
- preload-common.cjs: feature-detect via `allowedNodeEnvironmentFlags`; 26.5+ `nextLoad`s to native, else polyfills.
- docs: "polyfilled below 26.5, native above", floor 18.20.

Byte-parity vs `node --experimental-import-text` on 26.5.0; polyfill on 26.4.0/18.20.4.

PR #396, opened 2026-07-09
runtime: defer text imports to Node native on 26.5+

colinhacks:
Node 26.5.0 added `--experimental-import-text` (nodejs/node#62300). On Node >= 26.5.0 nub injects it and its preload defers `with { type: "text" }` to Node's native translator; below 26.5 nub's own `loadTextImport` short-circuit stays the mechanism. Output is byte-identical across the boundary (both `TextDecoder`, default export, BOM stripped).

The preload defers only when signaled (`__NUB_NATIVE_IMPORT_TEXT`, set iff the flag is in the final inject set) AND Node >= 26.5.0, so `--no-experimental-import-text` and version downgrades fall back to nub's impl. Verified on Node 26.5.0 / 26.2.0 / 24.4.0; tiers pinned in `version_tiers.rs`.

### DIFF
--- a/crates/nub-cli/src/cli.rs
+++ b/crates/nub-cli/src/cli.rs
@@ -3723,6 +3723,10 @@ fn build_script_command(
         aug.apply_localstorage_env(|k, v| {
             command.env(k, v);
         });
+        // Import Text native-defer signal (Node ≥ 26.5.0); inherited by the subtree.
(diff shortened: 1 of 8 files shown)


colinhacks:
Superseded by #395, which shipped the same behavior via native feature detection.

PR #398, opened 2026-07-09
Intersect injected flags with the Node binary's accepted set

colinhacks:
nub injects `--experimental-*` flags to unflag Node features early. Several of those bands are open-ended (`[lo, ∞)`, e.g. `--experimental-import-text` on `[26.5.0, ∞)`) on the assumption that once a flag exists Node keeps accepting it forever.

That assumption is false. Node keeps *some* unflagged flags as accepted no-ops (`--experimental-fetch`, `--experimental-modules`) but **hard-removes others**: `--experimental-policy` and `--experimental-network-imports` are gone, and `--experimental-permission` was accepted through Node 23.11 then deleted at 24.0 (it stabilized to `--permission`). Passing a removed flag is a `node: bad option` startup abort (exit 9), before any code runs. So an open-ended band would crash nub on whatever future Node drops the flag — and runtime feature-detection can't help, since the flag is parsed before any preload loads.

## Fix

Make injection self-correcting by intersecting the version-band set with the target Node's *actual* accepted-flag set:

- `discovery::accepted_env_flags(node_path)` probes `process.allowedNodeEnvironmentFlags` once and caches it per `(path, mtime)` in `~/.cache/nub/node-env-flags.json`. Amortized cost ≈ 0 (one probe per binary, ever; spawn-free on the warm path).
- `compute_inject_flags` gained a Stage-4 intersection: a flag the running Node doesn't accept is dropped. `None` (probe unavailable) preserves the prior version-band behavior, so a failed probe never regresses.
- Sound because nub injects flags via argv **and** propagates them to children via `NODE_OPTIONS`, so every injected flag is already envvar-allowed — exactly what `allowedNodeEnvironmentFlags` enumerates. A `host_node_accepts_every_injected_flag` test guards the invariant (the intersection drops nothing on a current, supported Node).

A future Node that removes a flag now needs no nub release — the probe stops offering it and Stage 4 drops it.

## Verification

- Empirically confirmed the removal behavior against installed Node 18.20–26.5 and `node_options.cc` (no-op vs hard-removed flags; the `--experimental-permission` → `--permission` deletion at 24.0).
- e2e: `--experimental-import-text` is injected on Node 26.5 (present in its probed set) and not on 26.2 (absent); warm runs hit the cache with no re-probe.
- `cargo clippy --all-targets --all-features -D warnings`, `cargo fmt --check`, and the `nub-core` node test suite pass.

Refs #395.

### DIFF
--- a/crates/nub-cli/src/cli.rs
+++ b/crates/nub-cli/src/cli.rs
@@ -3634,6 +3634,7 @@ fn build_script_command(
     );
     let aug = nub_core::node::spawn::compute_augmentation_env(
         &nub_binary,
+        node.path.as_std_path(),
         node.version,
         compat_mode,
         pnp_ctx.as_ref().map(|c| c.pnp_cjs.as_path()),
@@ -4558,11 +4559,13 @@ fn run_watch(file: &str, args: &[String]) -> Result<i32> {
     let mut node_args = vec!["--watch".to_string(), "--watch-preserve-output".to_string()];
 
     let node_options = env::var("NODE_OPTIONS").ok();
+    let accepted = nub_core::node::discovery::accepted_env_flags(node.path.as_std_path());
     let inject = nub_core::node::flags::compute_inject_flags(
         node.version.clone(),
         args,
         node_options.as_deref(),
         false,
+        accepted.as_ref(),
     );
     for flag in &inject {
         node_args.push(flag.to_string());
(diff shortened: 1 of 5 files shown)


colinhacks:
Shipped in v0.4.6: https://github.com/nubjs/nub/releases/tag/v0.4.6

PR #400, opened 2026-07-10
feat(runtime): backport --experimental-detect-module for ambiguous ESM .js

colinhacks:
Injects `--experimental-detect-module` where the flag exists and is still required, so an ambiguous ESM `.js` (ESM syntax, no `package.json` `"type"` field) runs as ESM under nub on older Node that refuses it bare.

Band `[20.10.0, 20.19.0) ∪ [21.1.0, 22.7.0)`: introduced 21.1.0 (nodejs/node#50096), backported 20.10.0, default-on 20.19.0 / 22.7.0 (nodejs/node#53619); never on 18.x/19.x/21.0.0. Tight-banded (like sqlite/wasm) so a future removal cannot crash.

Additive: CJS `.js` unaffected; `--node`/`NODE_COMPAT` stay unaugmented. `nub amb.js` goes rc 1 → rc 0 on Node 20.11, green on 22.15/26.5. Unit tests pin the boundaries; an integration test runs it on real 20.11.0. First of two PRs.

### DIFF
--- a/crates/nub-cli/tests/version_tiers.rs
+++ b/crates/nub-cli/tests/version_tiers.rs
@@ -258,6 +258,33 @@ fn import_text_works_via_native_on_26_5() {
     );
 }
 
+/// Module syntax detection backport on the COMPAT tier. Node 20.11.0 sits in the
+/// `--experimental-detect-module` injected band ([20.10, 20.19), before the 20.19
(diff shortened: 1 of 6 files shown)


colinhacks:
Folding into the single flag-detection PR.

PR #401, opened 2026-07-10
feat(runtime): structured-probe auto-backport of experimental Node flags

colinhacks:
Replaces the hand-maintained version bands for experimental-flag injection with a per-binary capability probe. nub asks each resolved Node which `--experimental-*` flags it actually accepts — via `internalBinding('options')` under `--expose-internals`, folded into the existing version-detection spawn and cached in `node-discovery.json` — then injects nub's wanted enablers wherever the binary accepts them. The probe is the version gate: a flag is injected where it exists and dropped where it doesn't, so a future removal, the 21.x eventsource hole, and patched/lying-version builds all fall out for free. This closes the exit-9-on-removal blind spot, where injecting a flag a binary rejects aborts the process before user code runs.

## Policy seam

The injection policy is a single swappable function (`flags::experimental_inject_set`). The default is an allowlist: inject `accepted ∩ wanted-enablers`, where the wanted set is the feature matrix's `Unflag` enablers plus `--experimental-detect-module`. Every candidate is guarded by arity (`type ∈ {kNoOp, kBoolean}`) and NODE_OPTIONS-legality (`env == 0`), so a value-taker is never bare-injected into argv or NODE_OPTIONS. Widening to an accepted-minus-denylist policy later is a one-function change.

Hard-excluded from the wanted set: `--experimental-shadow-realm` (the harmony denylist, #246 — crashes embedded Node/Electron on a V8 snapshot-hash mismatch), `--experimental-webstorage` (spawn.rs owns it, paired with `--localstorage-file`), and strip/transform-types (nub's transpiler owns `.ts`). When the probe is unavailable — an old accessor, a hardened build, a spawn failure — injection fails safe to the matrix version bands (today's behavior).

## detect-module

detect-module now falls out of the mechanism with no hand-coded band: an ambiguous ESM `.js` runs as ESM wherever the binary accepts the flag. A `version_tiers` integration test on Node 20.11 is the headline verification — bare 20.11 refuses the ambiguous file (exit 1), nub probes it, sees the flag accepted, and injects it. A differential fixture backs the compat-tier fix.

## Cache

`node-discovery.json` gains a per-binary accepted-flags record with a `flags_probed` marker. The reader is tolerant — pre-feature entries re-probe; a probed-but-unintrospectable build is negative-cached so it never spawn-storms. Writes are atomic (temp + rename), since discovery now has two writers on the same entry.

## Verification

Probe validated on Node 18.19 / 20.11 / 22.14 / 22.15 / 24.4 / 26.5. End-to-end, the injected set is exactly the accepted wanted enablers per binary, and the hazards (shadow-realm, ffi, vfs, print-required-tla, async-context-frame, strip/transform-types) are never injected even on a 26.5 that accepts them. 133 nub-core lib tests and 11 `version_tiers` tests pass; `clippy --all-targets --all-features` and `fmt` clean.

Three fresh-context reviews (correctness, impact-analysis, adversarial fail-safe/value-taker) are folded in.

### DIFF
--- a/crates/nub-cli/src/cli.rs
+++ b/crates/nub-cli/src/cli.rs
@@ -3632,11 +3632,17 @@ fn build_script_command(
         cmd.split_whitespace()
             .chain(args.iter().map(String::as_str)),
     );
+    // Probe which experimental flags this Node accepts (cache-amortized). Skip on
+    // compat mode, where compute_augmentation_env injects nothing anyway.
+    let accepted = (!compat_mode)
+        .then(|| nub_core::node::discovery::accepted_experimental_flags(node.path.as_std_path()))
+        .flatten();
     let aug = nub_core::node::spawn::compute_augmentation_env(
         &nub_binary,
         node.version,
         compat_mode,
         pnp_ctx.as_ref().map(|c| c.pnp_cjs.as_path()),
+        accepted.as_ref(),
     );
 
     let mut command = StdCommand::new(shell);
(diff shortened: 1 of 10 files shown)


colinhacks:
Superseded. The runtime auto-inject mechanism is scratched: auto-injecting flags nub hasn't individually vetted is an unacceptable risk regardless of allowlist/denylist framing, and the hand-vetted feature matrix is the safety boundary. Replaced by hand-vetted matrix bands plus a scheduled CI flag-drift snapshot guard, in a new PR. Branch kept at flag-detect-probe for reference.

PR #402, opened 2026-07-10
feat(node): matrix bands for detect-module/ffi/vfs/stream-iter + flag-drift snapshot CI

colinhacks:
Two related changes to how nub tracks Node's experimental flags, in one PR.

## Part A — four hand-vetted feature-matrix bands

Adds `Unflag` bands to the Node feature matrix so nub injects each flag on exactly the versions that accept it and still require it. Every floor was pinned empirically (probed `internalBinding('options')` plus a `node:<module>` load test) and corroborated against the Node changelog:

| flag | band | evidence |
|---|---|---|
| `--experimental-detect-module` | `[20.10.0, 20.19.0)` ∪ `[21.1.0, 22.7.0)` | intro 21.1.0 (#50096), backported 20.10.0; default-on 20.19.0 & 22.7.0 (#53619); the 21.0.0 release predates the flag, so the band excludes it — injecting there is a `bad option` crash, the same EOL-line hole as eventsource |
| `--experimental-ffi` (`node:ffi`) | `[26.1.0, ∞)` | added 26.1.0 (#62072); absent 26.0.0 |
| `--experimental-vfs` (`node:vfs`) | `[26.4.0, ∞)` | added 26.4.0 (#63115); absent 26.3.0 |
| `--experimental-stream-iter` (`node:stream/iter`) | `[25.9.0, ∞)` | added 25.9.0 (#62066); absent 25.8.1 |

The three module flags are default-off through Node 27 nightly (each module throws `ERR_UNKNOWN_BUILTIN_MODULE` without its flag), so the bands are open-ended; all four are `kBoolean`/NODE_OPTIONS-legal and snapshot-neutral (module gates, not V8 harmony flags). Band boundaries are covered by the pure-Rust `unflag_set_matches_known_boundaries` unit test. Two `version_tiers` integration tests plus fixtures verify end-to-end that nub injects the flags — an ambiguous ESM `.js` runs as ESM on Node 20.11, and `node:ffi`/`vfs`/`stream-iter` import on Node 26.5 (bare Node rejects all four). CI side-installs 20.11.0 and 26.5.0 so both tests run there.

## Part B — scheduled flag-drift snapshot guard

A CI/build-time guard (no runtime probe — the matrix stays hand-vetted) that warns when Node changes its `--experimental-*` flag set:

- `scripts/node-flag-inventory.mjs` dumps the running Node's experimental-flag set (`name → [OptionType, envVarSettings]`) via `internalBinding('options')`.
- `scripts/check-node-flag-drift.mjs` diffs a fresh capture against the committed snapshot and lists added/removed/changed flags.
- `scripts/node-flag-inventory.snapshot.json` pins Node 26.5.0 (42 flags).
- `.github/workflows/node-flag-drift.yml` runs a deterministic per-PR `snapshot-check` (Node pinned to the snapshot version, must match) and a scheduled `nightly-drift` job over Node nightly and RC. The nightly job fails loudly with the added/removed flags so a new experimental flag is vetted — and hand-added to the matrix if nub should support it — before nub ever claims it.

Verified against real Node 26.2/26.5 and a live Node 27 nightly: the nightly check correctly flags `--experimental-dtls` (new in 27) as drift to vet, and the introspection accessor still works on nightly.

## Verification

All floors probed on real binaries across Node 18.19–27-nightly. Gates green locally: `unflag_set_matches_known_boundaries` and the full `feature_matrix`/`version_tiers`/`node::` suites pass; `clippy --all-targets --all-features -- -D warnings` clean; `fmt --check` clean on both the default and 1.97.0 toolchains. Two fresh-context reviews (correctness, impact-analysis) folded in.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -618,13 +618,14 @@ jobs:
           node-version: ${{ matrix.node }}
           cache: 'pnpm'
       # Side-install the off-matrix tier representatives for the version_tiers
-      # integration test (compat 22.13 and unsupported 18.18). actions/setup-node
-      # only manages one PATH-active Node per job; these two go into private
-      # dirs and are exposed to the test via TEST_NODE_BIN_<MAJOR>_<MINOR>_<PATCH>
-      # — the env-var prefix deliberately avoids `NUB_*` (brand-boundary rule
-      # applies even to test-only knobs). Linux-only because the tarball URL
-      # paths differ per OS and the version_tiers contract is runtime-specific,
-      # not OS-specific.
+      # integration test: compat 22.13, unsupported 18.18, detect-module in-band
+      # 20.11 (flag still required there), and 26.5 (carries the node:ffi/vfs/
+      # stream-iter enabler modules). actions/setup-node only manages one
+      # PATH-active Node per job; these go into private dirs and are exposed to
+      # the test via TEST_NODE_BIN_<MAJOR>_<MINOR>_<PATCH> — the env-var prefix
+      # deliberately avoids `NUB_*` (brand-boundary rule applies even to test-only
+      # knobs). Linux-only because the tarball URL paths differ per OS and the
+      # version_tiers contract is runtime-specific, not OS-specific.
       - name: Install side Node binaries for version_tiers
         if: runner.os == 'Linux'
         shell: bash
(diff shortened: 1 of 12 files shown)


colinhacks:
Shipped in v0.4.6: https://github.com/nubjs/nub/releases/tag/v0.4.6

PR #412, opened 2026-07-10
fix(linker): exec native-executable bins directly instead of wrapping in node

### LINKED ISSUE #394: `nubx esbuild` errored out, should not generate a wrapper script
`esbuild` version: `0.28.1`

Since `esbuild` is a native program and not a JS script, but `nub` auto generates a script to run `esbuild` using Node in `node_modules/.bin`.

colinhacks:
`.bin/esbuild` ran `node <native-binary>` and failed: esbuild's postinstall swaps its JS launcher for the native binary, but bins are linked before build scripts run, so the `node` shim was never regenerated.

- `create_bin_shim` detects native magic bytes (ELF/Mach-O/PE) → direct-exec (symlink on Unix, `cmd`/`ps1`/`sh` wrappers on Windows), not `node`.
- Bins re-linked after build scripts run (shared `link_all_bins`).
- `nubx`'s resolver execs a native target directly (runtime safety net).

Reproduced with `esbuild@0.28.1`; after fix both `.bin` and `nubx` return `0.28.1`. Unit tests added; clippy + tests green.

Closes #394


colinhacks:
Shipped in v0.4.7: https://github.com/nubjs/nub/releases/tag/v0.4.7

PR #438, opened 2026-07-10
sandbox(docs): fix Windows crash-leak reasoning + macOS hardlink-to-secret parity

colinhacks:
Two accuracy fixes to `LIMITATIONS.md`, no engine change.

- Windows crash-leak: dropped the "profile is deleted" justification (wrong — a hard `TerminateProcess` kill also skips `ProfileGuard`'s `Drop`, leaking the profile). The orphan is still inert, but because the AppContainer SID is unique per run.
- Added a macOS parity note next to the Linux hardlink-to-secret residual — Seatbelt is path-pattern based like Landlock, so the same alias applies.

Refs the sandbox epic.


A000000009:
cd najmcoin-dapp
git init
git add .
git commit -m "NajmCoin DApp"

PR #444, opened 2026-07-12
feat(resolver): support pnpm namedRegistries in pnpm-compat mode

### LINKED ISSUE #440: `nub install` fails with dependencies using named registries
When installing dependencies via [named registries](https://pnpm.io/settings#namedregistries) directly, or even when installing dependencies that themselves use named registries `nub install` fails.

Direct:
```
➜ nub install
ERR_NUB_NO_MATCHING_VERSION

  ╰─▶ no version of @work/compression matches range `work:0.x.x`
  help: available versions: 0.0.18, 0.0.17, 0.0.16, 0.0.15, 0.0.14
```

Sub-Dependencies:
```
➜ nub install
ERR_NUB_NO_MATCHING_VERSION

  ╰─▶ no version of @work/constants matches range `work:1.x.x`
  help: chain: @work/exceptions@1.1.167 > @work/constants
        available versions: 1.1.668, 1.1.667, 1.1.666, 1.1.665, 1.1.664
```

They are defined like so:

```yaml
namedRegistries:
    work: https://npm.work.net/
```

in either the global or local pnpm configuration (for example `~/Library/Preferences/pnpm/config.yaml` and `./pnpm-workspace.yaml`)

which then lets you do this in the `package.json`:

```json
{
    "dependencies": {
        "@work/constants": "work:1.x.x"
    }
}
```

colinhacks:
Closes #440

Implements pnpm's `namedRegistries` — a `<alias>: <url>` map (from `pnpm-workspace.yaml` or the global pnpm `config.yaml`) that lets a dependency spec use the alias as a protocol prefix, e.g. `"@work/constants": "work:1.x.x"` resolves `@work/constants@1.x.x` from the `work` registry. Direct and transitive specs both resolve; the built-in `gh:` alias points at GitHub Packages. Auth rides existing `//host/:_authToken=` npmrc entries.

Before this, a `work:1.x.x` spec fell through to the semver resolver against the default registry — the reporter's `no version of @work/compression matches range work:0.x.x`.

## Scope

Active **only when pnpm is the incumbent PM** — gated on a new `EngineContext.named_registries_enabled` flag that defaults `false`, so standalone aube is byte-for-byte unchanged on the default path (fork-discipline). nub sets it when the project is pnpm-owned or fresh.

Spec forms mirror pnpm 11: `<alias>:<range>`, `<alias>:@owner/name@<range>`, `<alias>:<name>@<range>`, `<alias>:<tag>`. An alias colliding with a built-in scheme (`npm:`, `jsr:`, `workspace:`, …) is shadowed by that scheme. Invalid alias URLs warn-drop (pnpm hard-errors — see below).

## Mechanism

- Config: `discover_named_registries` reads the alias map (built-in `gh` < global `config.yaml` < workspace yaml), gated on the pnpm flag.
- Resolve: `preprocess_task` recognizes an `<alias>:` prefix, rewrites the range, and records the package → registry-URL route on the client. A per-client routing table, consulted by the single URL-derivation chokepoint, routes packument + tarball + auth to the aliased host.
- Lockfile: named-registry packages force-persist their tarball URL so a frozen install fetches from the right host without re-resolving; the frozen tarball-URL verify is skipped on host mismatch (content is still SHA-512-verified, and a missing integrity on that path is refused).

## Decisions for review

- **Gate breadth (recommend-only).** The feature is enabled for pnpm-incumbent **or fresh** projects (`read_branded_pnpm_config`), matching how every other pnpm-branded config read is gated. If you'd rather restrict it to strictly-pnpm-owned projects (exclude fresh), that's a `Role::Pnpm` gate instead — happy to switch.
- **Invalid-URL handling.** Warn-drop + skip, rather than pnpm's hard `INVALID_NAMED_REGISTRY_URL`, to keep install resilient. Easy to flip to a hard error.

## Known limitations (documented in-code)

- The route table is keyed by package name, so one project resolving the same package name from two different registries (one aliased, one default) can't be represented — both route to the alias. pnpm encodes the registry into the depPath; narrow same-name-two-registries case.
- A `gitBranchLockfile` merge doesn't run resolve-finalize, so a *custom* (non-`gh:`) named registry serving standard `/-/name-ver.tgz` tarball paths could lose its persisted tarball URL on merge. `gh:` is unaffected (non-derivable `/download/` URLs).

## Follow-ups (not in this PR)

- `nub add gh:@acme/foo` package.json write-path — this PR covers the install/resolve path only.
- Parser matches pnpm on all realistic inputs; minor divergences on malformed bodies (empty version → `latest` vs `*`) and a hardcoded default tag remain.

## Testing

- Unit: spec parser (9 cases incl. reserved-scheme shadowing), `discover_named_registries` (gate off/on, precedence, URL warn-drop), the frozen verify Verify/Skip/Refuse truth table.
- Behavioral e2e: `work:` routes to the aliased host under pnpm and stays inert under npm (gate off).
- Not exercised: a live frozen round-trip against a real aliased registry (no live registry available) — CI runs the full suite.


colinhacks:
Shipped in v0.4.9: https://github.com/nubjs/nub/releases/tag/v0.4.9

PR #446, opened 2026-07-12
feat(pm): version-gate the Expo GVS eject on SDK 56

colinhacks:
Expo's Metro became global-virtual-store-aware in **SDK 56** (On-demand Filesystem). Below 56 it crawls by realpath and can't reach the machine-global store — the same break as bare `react-native`. Expo was left on the shared store unconditionally, so pre-56 projects broke.

Now ejects GVS (project-local store) when the root manifest declares `expo` below major 56; keeps the shared store at 56+. Unfloorable ranges eject (safe direction). Nub-side only.

Limitation: root-manifest only (workspace-member Expo undetected, same scope as `vite_compat`; escape hatch `disableGlobalVirtualStoreForPackages[]=expo`). Unit-tested; clippy/fmt clean; docs updated.

### DIFF
--- a/crates/nub-cli/src/pm_engine/expo_compat.rs
+++ b/crates/nub-cli/src/pm_engine/expo_compat.rs
@@ -1,0 +1,122 @@
+//! Version-gated GVS eject for Expo.
+//!
+//! Expo gained global-virtual-store compatibility in **SDK 56** via its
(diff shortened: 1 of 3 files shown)


colinhacks:
Shipped in v0.4.10: https://github.com/nubjs/nub/releases/tag/v0.4.10

PR #448, opened 2026-07-12
feat(run): cross-platform script shell via deno_task_shell

### LINKED ISSUE #399: Cross platform shell for task runner

colinhacks:
Implements pnpm's `shellEmulator` via `deno_task_shell` (MIT): `nub run` script bodies run through a bundled POSIX-subset shell **by default on every platform**, so POSIX-isms (`rm -rf`, `&&`, `$VAR`) work on Windows without a system `sh`.

- Default-on is a deliberate divergence from pnpm (defaults off).
- Opt out: `shell-emulator=false` → native `sh`/`cmd`; `--script-shell <path>` overrides to a real shell.
- The augmentation env (PATH shim, `NODE_OPTIONS`, `npm_*`) reaches both paths, so `node`/`tsc` stays transpiled. deno's `KillSignal` handles SIGTERM/SIGINT.
- Scope: `nub run` only.

Behavior change; warrants a minor release.

Closes #399

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -277,7 +277,7 @@ dependencies = [
  "diffy",
  "glob",
  "hex",
- "junction",
+ "junction 2.0.0",
  "libc",
  "miette",
  "pathdiff",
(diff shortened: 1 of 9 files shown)


colinhacks:
Closing this. `nub run` will route script bodies through a real POSIX `sh` process rather than an in-process interpreter — full POSIX coverage, and it fits the sandbox model (which confines a spawned child process). The cross-platform-scripts goal from #399 carries into that work.

colinhacks:
Superseded by #559: `nub run` now spawns a real POSIX shell — the system `/bin/sh` on macOS/Linux, a bundled busybox-w32 on Windows.

PR #458, opened 2026-07-13
fix(import): run the peer pass in aube's own import; cover the bun parser

### LINKED ISSUE #453: nub import writes pnpm-lock.yaml without peer suffixes, breaking store-resident peers under GVS
## Summary

`nub import` (bun.lock → pnpm-lock.yaml) writes snapshots **without peer suffixes or mirrored peer edges**. nub then treats the result as a pnpm-lock incumbent — a format it assumes already carries peer resolution — so under the isolated/GVS layout a store-resident package gets no peer siblings in its private `node_modules`, and its peers fail to resolve **at runtime**.

## Environment

nub 0.4.11, macOS arm64, Node 24.

## Repro (real-world)

A Vite project using `@cloudflare/vite-plugin` (declares `peerDependencies: vite`), previously on bun:

1. `nub import` → "Imported 831 packages from bun.lock to pnpm-lock.yaml"
2. `nub install` (isolated default)
3. `nub run build`

**Actual:**

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'vite' imported from
~/.cache/nub/pm/store/@cloudflare+vite-plugin@1.26.0-a6eba333c4b73934/node_modules/@cloudflare/vite-plugin/dist/index.mjs
```

The imported lockfile shows the bare version — no `(vite@…)` suffix:

```yaml
'@cloudflare/vite-plugin':
  specifier: ^1.26.0
  version: 1.26.0        # organic pnpm-lock would be 1.26.0(vite@x.y.z)
```

**Expected:** peers resolved/suffixed like an organic pnpm-lock (or like installing with `bun.lock` kept as the incumbent, where the peer pass runs).

## Why

The install path runs its peer-context pass only for suffix-less incumbents (npm/bun); pnpm-lock incumbents skip it because "those lockfiles already carry peer suffixes". `nub import` produces a pnpm-lock that violates that assumption, and nothing re-establishes the peer edges afterwards.

## Workaround

`nub install --no-frozen-lockfile` immediately after the import — the fresh resolution computes peer suffixes and the build goes green (at the cost of within-range version bumps).

## Suggested fix

Run the peer-context pass as part of `nub import` when converting from a suffix-less format (bun/npm/yarn), so the written pnpm-lock satisfies the invariant the install path assumes. Alternatively, tag import-generated lockfiles so the install path knows to run the pass once.

afonsojramos:
Rebased after #471 landed the nub-cli half of this independently — this PR is now the complementary delta it left open:

1. **`aube import` (→ aube-lock.yaml) still wrote suffix-less lockfiles** from npm/bun sources. Same #453 failure class: install treats aube-lock.yaml as an already peer-resolved incumbent and skips its peer pass, so store-resident packages get no sibling peer links at runtime. This mirrors #471's exact pass, gating (`Npm | NpmShrinkwrap | Bun`), and rationale (no `filter_graph`, yarn excluded) at that call site.
2. **bun-parser regression test**: #471's test exercises the npm lockfile parser only; bun's peer-dependency plumbing is a separate parse path. Added `import_writes_peer_suffixes_for_bun_lock_source` with a real published peer pair (ajv-keywords → ajv, pinned integrities, offline), asserting both the suffixed snapshot key and the mirrored peer edge.

All three import tests green (`import_converts_package_lock_to_pnpm_lock`, #471's `import_writes_peer_suffixes_for_suffixless_source`, and the new bun one).

Fixes #453

### DIFF
--- a/crates/nub-cli/src/pm_engine/install_family.rs
+++ b/crates/nub-cli/src/pm_engine/install_family.rs
@@ -816,7 +816,7 @@ fn import_to_pnpm_lock(force: bool) -> miette::Result<String> {
         }
     };
 
-    let (mut graph, kind) = match aube_lockfile::parse_for_import(&root, &manifest) {
+    let (graph, kind) = match aube_lockfile::parse_for_import(&root, &manifest) {
         Ok(pair) => pair,
         Err(aube_lockfile::Error::NotFound(_)) => {
             restore(&backup);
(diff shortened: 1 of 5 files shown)


colinhacks:

PR #466, opened 2026-07-14
fix: filter runtime env keys across watch restarts

colinhacks:
## Summary

- keep env-file runtime-control keys filtered across Node watch restarts
- preserve ambient runtime values and user/PnP preload ordering
- fail closed on invalid guard state and verify the guard in the embedded runtime

## Verification

- focused watch and runtime-cache tests
- scoped nub-cli/nub-core clippy with all targets and features
- built-binary watch fixture covering late env-file edits
- direct malformed-state guard probes


colinhacks:

PR #467, opened 2026-07-14
fix(pm): repair incomplete version cache entries

colinhacks:
## Summary

- replace an incomplete package-manager version entry from verified staging
- preserve valid concurrent winners and clean quarantined invalid entries
- validate the expected launcher before and after placement

## Verification

- `cargo fmt --check`
- focused `nub-core` provision tests: 13 passed, 1 network-only ignored
- `scripts/rust-build.sh build -p nub-cli`
- local-registry corrupt-cache fixture
- `scripts/rust-build.sh clippy -p nub-core --all-targets -- -D warnings`


colinhacks:

PR #470, opened 2026-07-14
fix(node): harden temporary PATH shim directories

colinhacks:
## Summary

- create each temporary `node` PATH shim in an exclusive randomized private directory
- validate held directory and node identities before reuse and cleanup, including full Windows file IDs
- make stale cleanup bounded and nonrecursive while preserving legacy shim compatibility

## Validation

- `scripts/rust-build.sh test -p nub-core`
- selected concurrent child, independent process, recursive workspace, and aggregate-output integration tests
- real-binary randomized mode, same-PID legacy collision, nested Node, parallel workspace, and cleanup probes
- `scripts/rust-build.sh clippy --all-targets --all-features -- -D warnings`
- Windows GNU `nub-core` library and test-target checks
- `cargo fmt --check`
- `git diff --check`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2714,6 +2714,7 @@ dependencies = [
  "dirs-next",
  "dotenvy",
  "flate2",
+ "getrandom 0.3.4",
  "glob-match",
  "libc",
  "liblzma",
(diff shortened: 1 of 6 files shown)


colinhacks:

PR #471, opened 2026-07-14
fix(import): run peer-context pass when importing bun/npm to pnpm-lock

### LINKED ISSUE #453: nub import writes pnpm-lock.yaml without peer suffixes, breaking store-resident peers under GVS

colinhacks:
`nub import` wrote pnpm-lock.yaml without peer-context suffixes. The install path skips its peer pass for pnpm-lock incumbents, so imported locks left peer-dependent packages with no sibling under the isolated store layout — failing at runtime with `Cannot find package`.

Fix: run the peer pass (hoist → apply_peer_contexts → remove) in the import path for npm/npm-shrinkwrap/bun before writing. Yarn stays excluded, matching the install gate.

Verified: offline regression test; e2e on a real bun.lock now writes the `(vite@…)` suffix and the runtime error is gone.

Closes #453

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2667,6 +2667,7 @@ dependencies = [
  "aube-lockfile",
  "aube-manifest",
  "aube-registry",
+ "aube-resolver",
  "aube-scripts",
  "aube-settings",
  "aube-store",
--- a/crates/nub-cli/Cargo.toml
+++ b/crates/nub-cli/Cargo.toml
@@ -40,6 +40,11 @@ aube-lockfile = { path = "../../vendor/aube/crates/aube-lockfile" }
 # materialization) via `aube_linker::set_disk_materialize_expand_hook`; a pure
 # library crate (no CLI/tokio surface), same lean-graph rationale as aube-lockfile.
 aube-linker = { path = "../../vendor/aube/crates/aube-linker" }
+# aube-resolver: `nub import` runs the peer-context pass (hoist_auto_installed_peers
+# + apply_peer_contexts) on a bun/npm-parsed graph before writing pnpm-lock, so the
+# imported lockfile carries peer suffixes the install path assumes pnpm-locks have
+# (#453). Pure offline graph transforms; a lean library crate already in the graph.
+aube-resolver = { path = "../../vendor/aube/crates/aube-resolver" }
 aube-manifest = { path = "../../vendor/aube/crates/aube-manifest" }
 # default-features = false drops aube's default set (`mimalloc`, `config-tui`),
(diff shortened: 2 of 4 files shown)


afonsojramos:
hey @colinhacks! #458 had this fix open since yesterday 😅 But this PR is still incomplete I reckon: aube import still writes suffix-less aube-lock.yaml from npm/bun sources, and the test only exercises the npm parser. So I've rebased #458 down to exactly that remainder, plus a bun-parser regression test.

colinhacks:

PR #472, opened 2026-07-14
fix(runtime): auto-select async hook tier for a preload-detected foreign loader

### LINKED ISSUE #460: resolveSync ERR_METHOD_NOT_IMPLEMENTED crash: tsx inside a nested nub run on Node < 24.12
## Summary

On the Node band where the sync customization-hook stubs throw (< 24.12 per the comment in `runtime/preload-common.cjs`), running **tsx inside a nested `nub run`** can crash with:

```
Error [ERR_METHOD_NOT_IMPLEMENTED]: The resolveSync() method is not implemented
    at nextStep (node:internal/modules/customization_hooks:189:26)
    at resolve (~/.cache/nub/runtime-0.4.11-2107008e/preload-common.cjs:451:14)
    at nextStep (node:internal/modules/customization_hooks:189:26)
    at resolveWithHooks (node:internal/modules/customization_hooks:395:10)
```

Node.js v24.10.0, nub 0.4.11, macOS arm64.

## Two independent production hits (same repo, different paths)

1. `nub run precommit` → the script chains `nub run i18n:check` → `tsx scripts/i18n-check.ts` → the crash above, killing the pre-commit hook. Works when the same script runs via a **single-level** `nub run`, and works on Node ≥ 24.17. Workaround shipped: replace `tsx <file>` with `nub <file>` (which also happens to be the nub-native answer).
2. Playwright (running through nubx in CI) `globalSetup` spawning `tsx` inherited nub's `NODE_OPTIONS` preload and installed conflicting module resolvers. Workaround shipped: spawn the local tsx binary with `NODE_OPTIONS` deleted from the child env.

## Analysis

`preload-common.cjs` already anticipates this class: the `resolve` hook's catch checks `isAsyncLoaderSyncStub(err)` and falls back to `resolveViaParentRequire`, and the matcher does cover `ERR_METHOD_NOT_IMPLEMENTED` + `resolveSync`. So in the crashing case either:

- the fallback `resolveViaParentRequire` returned nothing for the specifier in play (a relative / TS-extension specifier from inside tsx's loader seems likely — parent `require` can't resolve `.ts` without hooks), and the stub error was rethrown, or
- the throw happened on a path that doesn't route through that catch (e.g. the `resolveWithHooks` frame above nub's own `nextStep` call, when tsx's registered hooks call back through the chain).

A trivial nested repro (`nub run outer` → `nub run inner` → `tsx t.ts` with a plain console.log, tsx 4.23.0, Node 24.10.0) does **not** trigger it — the failing script imported more (fs/path graph walking over a large repo), so the crash needs a resolution shape the minimal case doesn't hit. I can gather more from the affected repo's history if useful; the full stack above is from the live failure.

## Suggestions

- Harden the stub recovery for the rethrow path (when the fallback misses, degrading to Node's async resolution or to a transpile-aware fallback for `.ts` specifiers would keep tsx-under-nub alive on the affected Node band).
- Consider not exporting the preload into child `NODE_OPTIONS` for processes that aren't nub-spawned Node scripts, or de-duplicating hook registration when a nested nub context is detected — both production hits were "nub's hook + another tool's hook in one chain".
- Docs note: on Node < 24.12, prefer `nub <file.ts>` over `tsx <file.ts>` inside nub-run scripts.

colinhacks:
On Node 22.15.0–24.11.0 the async `module.register` loader's `resolveSync`/`loadSync` are unimplemented stubs (fixed in 24.11.1, nodejs/node#59666). nub's sync `module.registerHooks` fast tier forces resolution synchronous and hits those stubs when a foreign loader (tsx/ts-node) is active → `ERR_METHOD_NOT_IMPLEMENTED`.

nub already forces its async tier when the launcher argv scan sees a tsx child (`__NUB_FORCE_ASYNC_TIER`), but that scan is blind to tsx spawned via a nested `nub run` (the reported case), a `child_process` spawn (Playwright `globalSetup`), or a shell wrapper — nub never sees the tsx argv, so the process crashes.

## Fix

Detect the foreign loader intrinsically at preload from the process's own `--import`/`--loader` flags — scanning both `execArgv` and `NODE_OPTIONS` — on the broken band, and register nub's hooks via the async loader-worker tier so both loaders compose. Because it reads the process's own flags, it covers every spawn shape the launcher scan cannot see. Conservative by design: any such flag on the band takes the async tier (a worker-startup cost on a shrinking Node band, never a crash).

A mid-flight sync→async swap on foreign registration was prototyped and does not work — the loader worker cannot be registered from inside another loader's registration on this band — so preload-time selection is the viable point.

## Validation

- Real tsx, both crash reproductions (nested `nub run` + `child_process` spawn), fixed on Node 22.16 and 24.10.
- node-matrix green on 22.16 / 24.10 (broken) and 24.14 (fixed); new Scenario C guards both delivery channels and fails without the fix on the band.
- Fresh-context correctness + impact-analysis review: no false-positive risk (the `NODE_OPTIONS` scan rejects nub's own `--experimental-import-text`; nub never injects `--import`/`--loader` on the broken band).

## Follow-up (separate)

The Rust argv-scan apparatus (`child_hosts_async_loader` / `force_async_tier_env`) is now redundant with the intrinsic detection (strictly broader). Left in place to keep this diff tight.

Closes #460

https://claude.ai/code/session_013KZKJdgxVzFjyGxASsyhCy

### DIFF
--- a/runtime/preload-common.cjs
+++ b/runtime/preload-common.cjs
@@ -231,6 +231,56 @@ function userAsyncLoaderActive() {
   return __userAsyncLoaderRegistered || cliAsyncLoaderPresent();
 }
 
+// The Node band where the async `module.register` loader's `resolveSync`/`loadSync`
+// are unimplemented stubs that throw `ERR_METHOD_NOT_IMPLEMENTED`: 22.15.0 ..= 24.11.0
+// (fixed in 24.11.1, nodejs/node#59666). Mirrors the Rust `node_hook_compose_broken`
+// (spawn.rs) for every RELEASE version; a pre-release like `22.15.0-rc.1` reads as in-band
+// here (the numeric parse ignores the `-rc` tag) where the Rust semver sorts it just below
+// the 22.15.0 floor — a harmless over-selection (the async tier is always correct and the
+// two signals are OR'd), never a crash. Outside this band a foreign async loader composes
+// with nub's sync hooks natively, so the fast tier stays.
+function nodeHookComposeBroken() {
+  const p = String(process.versions.node).split(".");
+  const maj = parseInt(p[0], 10) || 0;
+  const min = parseInt(p[1], 10) || 0;
+  const pat = parseInt(p[2], 10) || 0;
+  const geFloor = maj > 22 || (maj === 22 && min >= 15);
+  const leCeil = maj < 24 || (maj === 24 && (min < 11 || (min === 11 && pat === 0)));
+  return geFloor && leCeil;
+}
+
+// Does a foreign async ESM loader flag (`--import` / `--loader` / `--experimental-loader`)
+// ride in THIS process's own startup flags, via EITHER channel? tsx/ts-node deliver their
+// loader through one of two paths, and they land in different places:
+//   • re-exec argv → `process.execArgv` (tsx's bin re-execs `node --import <loader>`), and
+//   • `NODE_OPTIONS="--import tsx/esm"` → `process.env.NODE_OPTIONS` (NOT hoisted into
+//     execArgv — verified on Node 24.x), the common CI/shell-config delivery.
+// Both must be scanned. nub's own fast-tier injection is `--require` (never `--import`/
+// `--loader`) on this band, and its compat-tier `--import preload.mjs` lives below 22.15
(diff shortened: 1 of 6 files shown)


colinhacks:

PR #474, opened 2026-07-14
feat(node): resolve the Node version from .tool-versions (asdf/mise)

colinhacks:
Resolve the project's Node version from `.tool-versions` (the asdf/mise file), so asdf/mise-driven projects that pin `nodejs` there provision the right Node under nub without a redundant `.node-version`.

## Resolution order

`.tool-versions` slots at precedence **#4** — below the Node-specific pin files, above the `engines.node` range:

1. `package.json#devEngines.runtime`
2. `.node-version`
3. `.nvmrc`
4. **`.tool-versions`** (new)
5. `package.json#engines.node`

The ordering is a specificity-of-intent gradient: a polyglot pin's `nodejs` line is one tool among many, so a deliberately-added `.node-version` wins when a project carries both. A `.tool-versions`-only project — the asdf/mise norm — never hits the conflict.

## Behavior

- Reads the `nodejs` (asdf plugin key) or `node` (mise alias) line; strips inline `#` comments; takes the first version on a multi-version fallback line (`nodejs 20 18` → `20`).
- Reuses `VersionPin` parsing, so aliases resolve and unmodelable specifiers (`system`, `ref:<sha>`) fall through to the next source exactly as an unparseable `.node-version` does.
- Only the Node line is read; every other tool in the file is ignored (nub provisions Node, not the whole toolchain).
- The source label, provenance line (`resolved from .tool-versions`), and disagreement warnings come through `walk_up_for_pin` with no extra plumbing.

## Scope

`.tool-versions` only; `mise.toml` (TOML, mise-only) is out of scope. The setup-nub GitHub Action provisions via nub, so it inherits this automatically once a release ships.

## Testing

Unit: 3 tests in `discovery.rs` (precedence vs `.node-version`/`.nvmrc`, beats `engines.node`, the `node`-key/inline-comment/multi-version/`system` shapes). Full `node::discovery` suite green; `clippy --all-targets --all-features -D warnings` and `fmt --check` clean.

E2e against the dev binary (`nub node which`): `.tool-versions nodejs 26.4.0` → `resolved from .tool-versions (26.4.0)`; adding `.nvmrc`/`.node-version` flips the winner correctly; `nodejs system` falls through to node on PATH.

Docs updated: runtime overview + the GitHub Action resolution list.

### DIFF
--- a/crates/nub-cli/src/cli.rs
+++ b/crates/nub-cli/src/cli.rs
@@ -6382,7 +6382,7 @@ fn discover_node_for_status(cwd: &Path) -> Result<nub_core::node::discovery::Res
             "pinned Node version {pin} not found\n\
              \x20\x20Active shell Node: {shell_version} (does not satisfy the pin)\n\
              \x20\x20Provision it: nub node install {pin} (or run a file — nub installs the pin on demand)\n\
-             \x20\x20The pin comes from .node-version / .nvmrc / engines.node / devEngines.runtime."
+             \x20\x20The pin comes from .node-version / .nvmrc / .tool-versions / engines.node / devEngines.runtime."
         ),
         other => anyhow::Error::new(other),
     })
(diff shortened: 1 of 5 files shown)


colinhacks:

PR #475, opened 2026-07-14
fix(pm): eject curated project-context packages from the global virtual store

### LINKED ISSUE #457: nub install "simple-git-hooks" fails; looks for package.json in the wrong directory
Issue is what it states on the tin ☝️ .

Reproducer:
```
mkdir /tmp/mytest
cd /tmp/mytest
npm init -y
nub i simple-git-hooks
```
Result:
```
nub i simple-git-hooks
Resolving simple-git-hooks@latest...
  + simple-git-hooks@2.13.1 (specifier: ^2.13.1)
Updated package.json
WARN defaultTrust: running build scripts for 1 default-trusted package(s): simple-git-hooks@2.13.1 code=WARN_NUB_DEFAULT_TRUST_BUILDS count=1 packages=["simple-git-hooks@2.13.1"]
node:fs:1734
  const stats = binding.stat(
                        ^

Error: ENOENT: no such file or directory, stat '/home/jlicht/.cache/nub/pm/store/simple-git-hooks@2.13.1-4787829ad8d51ce6/package.json'
    at Object.statSync (node:fs:1734:25)
    at _getPackageJson (/home/jlicht/.cache/nub/pm/store/simple-git-hooks@2.13.1-4787829ad8d51ce6/node_modules/simple-git-hooks/simple-git-hooks.js:232:13)
    at checkSimpleGitHooksInDependencies (/home/jlicht/.cache/nub/pm/store/simple-git-hooks@2.13.1-4787829ad8d51ce6/node_modules/simple-git-hooks/simple-git-hooks.js:129:34)
    at postinstall (/home/jlicht/.cache/nub/pm/store/simple-git-hooks@2.13.1-4787829ad8d51ce6/node_modules/simple-git-hooks/postinstall.js:15:9)
    at Object.<anonymous> (/home/jlicht/.cache/nub/pm/store/simple-git-hooks@2.13.1-4787829ad8d51ce6/node_modules/simple-git-hooks/postinstall.js:23:1)
    at Module._compile (node:internal/modules/cjs/loader:1871:14)
    at Object..js (node:internal/modules/cjs/loader:2002:10)
    at Module.load (node:internal/modules/cjs/loader:1594:32)
    at Module._load (node:internal/modules/cjs/loader:1396:12)
    at module_._load (/home/jlicht/.cache/nub/runtime-0.4.7-b1ba5736/preload-common.cjs:1014:30) {
  errno: -2,
  code: 'ENOENT',
  syscall: 'stat',
  path: '/home/jlicht/.cache/nub/pm/store/simple-git-hooks@2.13.1-4787829ad8d51ce6/package.json'
}

Node.js v24.18.0
  × lifecycle script postinstall failed for simple-git-hooks@2.13.1: script `postinstall` exited with code 1

```

FWIW, I can actually find what I assume to be the correct `package.json` at:
`/home/jlicht/.cache/nub/pm/store/simple-git-hooks@2.13.1-4787829ad8d51ce6/node_modules/simple-git-hooks/package.json`

Expectation: it works

colinhacks:
`nub i simple-git-hooks` crashes: its `postinstall` runs inside the global virtual store, detached from any project. simple-git-hooks walks up from `cwd` to find a project root, lands on the per-package store wrapper (which has no `package.json`), and throws ENOENT. The same package builds cleanly with GVS disabled, where it materializes project-local — so ejection is the fix.

This adds `NUB_PROJECT_CONTEXT_EJECT`, a curated list of packages whose build reads or mutates the *consuming* project, as an additional disk-materialize seed. Every listed name present in the resolved graph unions into the existing phantom-eject seed set and flows through the same importer-closure + disk-materialize path, so the package materializes project-local (a real project above `cwd`) instead of symlinked into the shared store.

**Curated, not blanket.** Self-contained builds — node-gyp native compiles, prebuilt-binary downloaders like esbuild — read only their own dir and share output across projects via the side-effects cache, so they stay symlinked in the store (built once, shared). Only the project-context class ejects.

Notes on the list:

- Git-hook installers (simple-git-hooks, lefthook, pre-commit, …) are leaves — the closure is the package itself, so ejecting them is free.
- `vue-demi` / `@intlify/*` bridges carry real importer closures in Vue projects. They are included because their build rewrites which output to expose based on the consuming project's installed Vue major, so their built content is project-specific — the eject is a correctness requirement, not only the #457 crash.
- `msw` and `install-peers` are included conservatively; they may already resolve the consuming project via `INIT_CWD`, but ejecting them is safe and removes the dependence on that.

The curated seed is injected inside the disk-materialize expand hook, past aube's `disk_materialize_packages` settings fold, so a warm/upgrade tree would otherwise keep an identical install-state fingerprint and skip the relink — leaving the fix inert on existing installs. A stable, order-independent token of the list is folded into the install-state fingerprint so the initial upgrade and any future list edit invalidate the warm tree and force the relink.

Verified end-to-end: `nub i simple-git-hooks` in a fresh project exits 0 and materializes project-local (was ENOENT); a bare re-`install` hits the up-to-date fast path with no re-materialize (no wipe-loop); `nub i esbuild` stays symlinked into the global store. Unit tests cover the eject, the stays-symlinked contract, and the fingerprint token's order-independence and edit-sensitivity.

Closes #457

### DIFF
--- a/crates/nub-cli/src/dynamic_phantom.rs
+++ b/crates/nub-cli/src/dynamic_phantom.rs
@@ -82,11 +82,17 @@ fn eject_disabled(raw: Option<&str>) -> bool {
 /// hash — this seam is what makes it invalidate the warm tree.
 ///
 /// For users the token is CONSTANT-ON: the dead on/off toggle is gone, so it folds
-/// only [`PHANTOM_SCANNER_VERSION`]. That fold is what makes a scanner-logic bump
-/// COMPLETE rather than a half-fix — the bump re-scans content into a new sidecar
-/// path, but on a warm tree with an unchanged lockfile aube would SKIP the link
-/// phase and never apply the improved verdict; changing this token forces the link
-/// to re-run so the consumer picks up the new-version sidecars.
+/// [`PHANTOM_SCANNER_VERSION`] plus the curated-eject list token
+/// ([`crate::pm_engine::phantom_closure::project_context_eject_token`]). The scanner
+/// fold makes a scanner-logic bump COMPLETE rather than a half-fix — the bump
+/// re-scans content into a new sidecar path, but on a warm tree with an unchanged
+/// lockfile aube would SKIP the link phase and never apply the improved verdict;
+/// changing this token forces the link to re-run so the consumer picks up the
+/// new-version sidecars. The curated-eject fold does the same for the #457 list: its
+/// members are injected inside the expand hook, past aube's `disk_materialize_packages`
+/// settings fold, so folding the list token here is what invalidates a warm tree on
+/// the initial ship AND on any future list edit (else the stale symlinked shape is
+/// accepted and #457 stays unfixed on existing installs).
 ///
 /// The token still branches on [`enabled`] SOLELY for the internal A/B seam: when
 /// an agent flips [`INTERNAL_EJECT_DISABLE_VAR`] the token changes, so a warm tree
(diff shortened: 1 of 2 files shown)


colinhacks:

colinhacks:
Shipped in v0.4.13: https://github.com/nubjs/nub/releases/tag/v0.4.13

PR #476, opened 2026-07-15
fix(pm): resolve the global store dir under %LOCALAPPDATA% on Windows

### LINKED ISSUE #451: Global store path mismatch on Windows
The `nub store path` command yields `C:\Users\<USER>\.local/share\nub\store\v1` on Windows but all packages in `node_modules/.store` resolve to `C:\Users\<USER>\AppData\Local\nub\pm\store\*`.

This mismatches the behavior described in [the documentation](https://nubjs.com/docs/install/virtual-store?section=one-symlink-per-package#one-symlink-per-package).

colinhacks:
On Windows `nub store path` printed a `…\.local/share\nub\store` path, splitting the CAS store from the cache tier at `…\AppData\Local\nub\pm`.

`nub_data_dir()` — source of the `storeDir` default and the phantom scanner's store path — read only `XDG_DATA_HOME` before `~/.local/share`, with no `%LOCALAPPDATA%` branch, shadowing the engine's Windows-aware `store_dir()` fallback. Fires on plain cmd/PowerShell.

Adds the `%LOCALAPPDATA%` precedence via an env-injected pure helper unit-tested off Windows. Unix unchanged; `fmt`/`clippy --all-targets`/test green. On-disk Windows placement gated by the Windows CI leg.

Closes #451

### DIFF
--- a/crates/nub-cli/src/pm_engine/mod.rs
+++ b/crates/nub-cli/src/pm_engine/mod.rs
@@ -2298,15 +2298,48 @@ fn nub_lockfile_present(dir: &Path) -> bool {
         || dir.join(use_align::NUB_LEGACY_LOCKFILE).is_file()
 }
 
-/// Nub's XDG data root (`$XDG_DATA_HOME/nub` or `~/.local/share/nub`), the
-/// data-dir sibling of `nub_core::node::discovery::cache_dir`.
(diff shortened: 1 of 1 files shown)


colinhacks:
Shipped in v0.4.13: https://github.com/nubjs/nub/releases/tag/v0.4.13

PR #499, opened 2026-07-21
feat(init): ship nub init — TypeScript-first project scaffold

### LINKED ISSUE #487: [Feature Request] support `init` command as what `pnpm` or `bun` or `deno` did

colinhacks:
Replaces the reserved `init` stub with the real command: scaffolds `package.json` (nub identity pin + `@nubjs/types`/`@types/node` devDeps), `tsconfig.json`, `index.ts`, `.gitignore`, `README.md`, then runs `git init` and `nub install` by default. Prompts for name/language/git in a TTY; `-y`, `--js`, `--name`, `--no-git`, `--no-install`, `--force`; non-TTY takes defaults. Positionals rejected with a `nubx create-<template>` hint. Docs at `/docs/init`.

Closes #487

Verified: e2e scaffold + install + run + `tsc --noEmit` on the generated project; PTY-driven prompts; 8 tests; clippy/fmt green.

https://claude.ai/code/session_01EGwuVYchyCCkGFzAik5c1y


crates/nub-cli/src/init.rs
```diff
@@ -0,0 +1,295 @@
...
+    // ── refuse-don't-clobber: every conflicting target named, then abort ─
+    let mut files: Vec<(&str, String)> =
+        vec![("package.json", manifest_json(&name, entry, typescript))];
+    if typescript {
+        files.push(("tsconfig.json", TSCONFIG.to_string()));
+    }
+    files.push((entry, "console.log(\"Hello from Nub\");\n".to_string()));
+    files.push((
+        ".gitignore",
+        "node_modules\n.env*\n*.log\n.DS_Store\n".to_string(),
```
afonsojramos, line 123: Minor: the scaffolded `.gitignore` uses `.env*`, which also ignores `.env.example`. `bun init` lists the env files explicitly (`.env`, `.env.*.local`); alternatively, a one-line `!.env.example` negation would do the same here. (Fair note: create-next-app shares the `.env*` behavior, so there's precedent either way.)

```suggestion
        "node_modules\n.env\n.env.local\n.env.*.local\n*.log\n.DS_Store\n".to_string(),
```

crates/nub-cli/src/init.rs
```diff
@@ -0,0 +1,295 @@
...
+}
+
+/// The scaffolded manifest, in display order. Identity fields come from the
+/// same writer `nub pm pin` uses so the two surfaces can't drift; both the pin
+/// and the `@nubjs/types` range derive from the running version (nub-types is
+/// release-locked to the binary).
+fn manifest_json(name: &str, entry: &str, typescript: bool) -> String {
+    let ver = env!("CARGO_PKG_VERSION");
+    let mut root = Map::new();
+    root.insert("name".into(), Value::String(name.into()));
```
afonsojramos, line 176: Optional hardening: no `"private": true` in the scaffolded manifest. tbh the current implementation matches the majority convention (`npm init -y`, `pnpm init`, `yarn init` v1 and v4 all omit it), so it's defensible as-is, but `bun init` does write it, and for an app-shaped scaffold the field is cheap insurance against an accidental `npm publish`.

site/content/docs/init.mdx
```diff
@@ -0,0 +1,109 @@
...
+    "lib": ["es2024"],
+    "types": ["node", "@nubjs/types"],
+    "strict": true,
+    "noEmit": true
+    // ...
+  }
+}
+```
+
+The generated project is not locked to Nub: everything runs on plain Node too (the `devEngines` policy is `warn`, not a hard block), and swapping the start script to a plain `node` invocation is the only change a non-Nub user would make.
```
afonsojramos, line 72: Small accuracy caveat: "swapping the start script to a plain `node` invocation is the only change a non-Nub user would make" holds only on Node ≥22.18 (type stripping, unflagged in 23.6, backported to 22.18). Verified locally: the scaffolded `index.ts` runs on Node 24.18 and 22.22, but fails on Node 20.20 with `ERR_UNKNOWN_FILE_EXTENSION`. Since nub's own floor is 18.19, users on 18/20 would need a newer Node or the `--js` variant.

crates/nub-cli/src/init.rs
```diff
@@ -0,0 +1,295 @@
...
+        .chars()
+        .map(|c| if c.is_whitespace() { '-' } else { c })
+        .filter(|c| c.is_ascii_alphanumeric() || matches!(c, '-' | '_' | '.' | '~'))
+        .collect();
+    while s.starts_with(['.', '_', '-']) {
+        s.remove(0);
+    }
+    s.truncate(214);
+    s
+}
```
afonsojramos, line 264: `sanitize_name` silently mangles scoped names: `--name @scope/pkg` comes out as `"name": "scopepkg"` (the `@` and `/` are dropped by this filter). Verified against a local build of this branch.

Scoped names need to be first-class npm citizens (`npm init --scope=@myorg` yields `@myorg/<dir>`), and since `init` is the only PM scaffold that accepts arbitrary `--name` input, it's also the only one that needs to accept scope. 

Could we instead honor the `@[a-z0-9-_.~]+/[a-z0-9-_.~]+` shape? Sanitize per-part, and reject a bare `@scope` with no package part (returning `scope` would be the same silent-mangling bug in a smaller coat):

```suggestion
/// npm-valid name from arbitrary input. A scoped name (`@scope/pkg`)
/// sanitizes per-part, preserving `@` and `/`; a bare `@scope` with no
/// package part is invalid (empty result = the caller re-prompts/falls
/// back). Anything else sanitizes as one token.
fn sanitize_name(raw: &str) -> String {
    let raw = raw.trim();
    if raw.starts_with('@') {
        return match raw.strip_prefix('@').and_then(|r| r.split_once('/')) {
            Some((scope, pkg)) => {
                let (scope, pkg) = (sanitize_part(scope), sanitize_part(pkg));
                if scope.is_empty() || pkg.is_empty() {
                    String::new()
                } else {
                    let mut s = format!("@{scope}/{pkg}");
                    s.truncate(214);
                    s
                }
            }
            None => String::new(),
        };
    }
    sanitize_part(raw)
}

/// One name segment: lowercase, whitespace → `-`, keep `[a-z0-9-_.~]`, no
/// leading `.`/`_`/`-`.
fn sanitize_part(raw: &str) -> String {
    let mut s: String = raw
        .to_lowercase()
        .chars()
        .map(|c| if c.is_whitespace() { '-' } else { c })
        .filter(|c| c.is_ascii_alphanumeric() || matches!(c, '-' | '_' | '.' | '~'))
        .collect();
    while s.starts_with(['.', '_', '-']) {
        s.remove(0);
    }
    s.truncate(214);
    s
}
```

Empty-from-sanitize stays the "invalid" signal, so the interactive validator and the `--name` bail in `run_init` keep working unchanged. Test cases for the new shapes are in a separate suggestion on the sanitize test below.

crates/nub-cli/src/init.rs
```diff
@@ -0,0 +1,295 @@
...
+mod tests {
+    use super::*;
+
+    #[test]
+    fn sanitize_lowercases_hyphenates_and_strips_leading_punctuation() {
+        assert_eq!(sanitize_name("My App"), "my-app");
+        assert_eq!(sanitize_name(".hidden"), "hidden");
+        assert_eq!(sanitize_name("__weird$name!"), "weirdname");
+        assert_eq!(sanitize_name("ok-name_1.2~x"), "ok-name_1.2~x");
+        assert_eq!(sanitize_name("🎉🎉"), "");
```
afonsojramos, line 276: If the scope-aware shape lands, these cover it alongside the existing cases — including the two invalid shapes that must stay empty (a bare `@scope` is not a name, and must not become `scope`):

```suggestion
        assert_eq!(sanitize_name("🎉🎉"), "");
        assert_eq!(sanitize_name("@scope/pkg"), "@scope/pkg");
        assert_eq!(sanitize_name("@My Org/My App"), "@my-org/my-app");
        assert_eq!(sanitize_name("@scope"), "");
        assert_eq!(sanitize_name("@/pkg"), "");
```

afonsojramos (commented):
Sorry for prying, but you opened this PR at the same time as mine and I decided to take a look 👀
Looks awesome tbh! Left just four comments, three of which are genuinely minor. The only higher priority one is the scoped-name mangling: `--name @scope/pkg` silently becomes scopepkg. The .env.example negation is a one-liner worth taking; the private: true and Node-version-doc points are take-it-or-leave-it.

colinhacks:

PR #500, opened 2026-07-21
feat(run): fall back to node_modules/.bin on a script miss

afonsojramos:
## Summary

`nub run <name>` now falls back to `node_modules/.bin/<name>` when the name matches no script, so locally installed CLIs run without a pass-through script. `nub run astro dev` works in a project with no `"astro": "astro"` boilerplate — matching `bun run` and aube's own documented contract ("If no script matches, aube falls back to `node_modules/.bin/<name>`"), which nub's Rust `run` path had not implemented.

```bash
nub run astro dev        # no script needed; args forward as with a script
nub run tsc --noEmit
```

Behavior, in short:

- A script always wins over a same-named bin.
- Resolution walks up to the workspace root, so hoisted installs work; Yarn PnP trees resolve through the PnP bin-runner `nub exec` already uses.
- Local only — a bin miss never reaches the registry (that stays `nubx`/`dlx` territory).
- A bin is not a lifecycle: no `pre`/`post` hooks, no `npm_lifecycle_*` env. `--if-present` treats a resolvable bin as present. `/regex/` selectors never fall back.
- Double miss keeps exit 1 and the existing `missing script` error, extended to note the `.bin` lookup.

## Design decisions

- **Hook at the two script-miss sites** (single-package selector and `-w` root lookup), trying the bin before `--if-present`, matching aube's order. Resolution and launch reuse the `nub exec` machinery (`find_bin` walk-up, shebang-aware `launch_bin`) — no new resolution logic.
- **PnP probe mode**: the shared `pnp-bin-run.cjs` runner gained a quiet `--probe` (exit 0/1, no output) so a PnP bin miss doesn't shadow the `missing-script` error with the runner's own 127 message.
- **Plain names only**: the fallback refuses path-shaped names, so `nub run ../foo` can't traverse out of `.bin` via the walk-up (found in review; regression test included).
- **Recursive fan-out unchanged**: `-r`/`--filter` keep script-only resolution; per-member bin fallback is a deliberate follow-up.

Session-settled decisions carried from planning: fallback scope limited to single-package + `-w` (user-approved, over including `-r`/`--filter` now); no lifecycle hooks for bins (user-approved, over honoring `pre<name>`/`post<name>`); PnP parity included via the existing runner (user-approved, over node_modules-only); double miss keeps exit 1 (user-approved, over exec's 127 shape).

## Test plan

- 11 new integration tests (written red-first): fallback + arg forwarding, script precedence, non-node (`#!/bin/sh`) bins, walk-up to the workspace root, `--silent` preamble suppression, `--if-present` semantics, no lifecycle hooks, double-miss error shape, regex selectors never falling back, `-w` root bin + root-script precedence, path-traversal guard.
- `cargo test -p nub-core -p nub-cli`: full suite green.
- PnP matrix: 18/18, including 3 new checks — `run-bin` (PnP fallback), `run-bin-miss` (probe-silent double miss keeps the `missing-script` error), `ws-run-bin` (from a member).
- Conformance `run-missing-script` cell still pins exit 1; expectation note updated to record the new requirement (bin lookup must also miss).
- Windows is exercised by the `.cmd`-shaped fixtures on the windows CI leg; `find_bin`'s extension handling is already unit-tested.

## Post-Deploy Monitoring & Validation

No additional operational monitoring required — this is a local CLI behavior change with no service, telemetry, or data-path impact; the conformance suite and PnP matrix pin the behavior on every CI run.

## Documentation

`runner/run.mdx` gains a "Local binaries" section (precedence, no-registry boundary, PnP, `--if-present`, the deliberate `pnpm run`/`npm run` deviation), and `runner/exec.mdx` notes when `exec` remains the explicit form.

### DIFF
--- a/crates/nub-cli/src/cli.rs
+++ b/crates/nub-cli/src/cli.rs
@@ -2803,13 +2803,24 @@ fn run_script(
             nub_core::workspace::detect::detect_project(&ws_root).ok_or_else(|| {
                 anyhow::anyhow!("--workspace-root: no package.json at {}", ws_root.display())
             })?;
-        let cmd = nub_core::workspace::scripts::resolve_script(&root_project.manifest, script)
-            .ok_or_else(|| {
-                anyhow::anyhow!(
-                    "missing script: \"{script}\" in the workspace root\n\nAvailable scripts:\n{}",
+        let cmd = match nub_core::workspace::scripts::resolve_script(&root_project.manifest, script)
+        {
+            Some(cmd) => cmd,
+            None => {
+                // The same `.bin` fallback as the single-package path, resolved
+                // against the workspace root's chain (see `run_selected_scripts`).
+                if let Some(code) =
+                    run_bin_fallback_for_plain_name(script, &ws_root, compat_mode, args)?
+                {
+                    return Ok(code);
+                }
+                let bin_note = bin_miss_note(script);
+                anyhow::bail!(
+                    "missing script: \"{script}\" in the workspace root{bin_note}\n\nAvailable scripts:\n{}",
                     list_scripts(&root_project.manifest)
-                )
-            })?;
+                );
+            }
+        };
         let exec = ScriptExecOpts {
             ignore_scripts: ws.ignore_scripts,
             script_shell: ws.script_shell.as_deref(),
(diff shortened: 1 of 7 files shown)


afonsojramos:
@colinhacks can you take a look?

colinhacks:
Thanks for the thorough work here — clean implementation and tests.

We're going to keep a clear separation between the subcommands: `nub run` stays for `package.json` scripts, and running a locally installed binary is what `nub exec`/`nubx` are for. A bin fallback in `nub run` blurs that line and diverges from `pnpm run`, which we mirror deliberately — so we'll pass on this one.

For the `nub run astro dev` case, `nubx astro dev` (or `nub exec astro dev`) does the same thing and keeps it explicit that you're executing a binary. Appreciate the effort, and hope to see more PRs from you.

PR #501, opened 2026-07-21
Fall back to webpki roots and system DNS on Android (no-JVM builds)

### LINKED ISSUE #490: Android (Termux): all network operations abort — rustls-platform-verifier needs a JVM (ndk-context not initialized)
## Summary

When nub is built for `target_os = "android"` and run in Termux (Android CLI environment, no JVM), **every operation that performs an HTTPS request aborts** with exit code 134:

```
thread 'tokio-rt-worker' panicked at .../ndk-context-0.1.1/src/lib.rs:72:30:
android context was not initialized
Aborted
```

Affected commands (100% reproducible):
- `nubx -y cowsay@1.5.0 hello` (registry fetch fallback)
- `nub add is-odd`
- `nub install` in a project whose deps must be fetched from a registry

Local-only paths work fine: running .ts/.js files, `nub run` (incl. workspace `-r`/`--filter`), `nubx` resolving a local `node_modules/.bin`, `nub remove`, `nub node which/ls`.

## Environment

- Device: Android (aarch64), Termux userland (bionic libc, **no JavaVM**)
- nub v0.4.13, built from source (commit 4fd083f) with `cargo build --release -p nub-cli --features embed-runtime`
- System Node v26.4.0; network and TLS stack verified working (curl, node, cargo all fetch HTTPS fine)

## Root cause

nub links `rustls-platform-verifier`, which on Android targets the platform certificate store **via JNI** and requires `ndk_context` to be initialized with a JavaVM. That assumption holds for Android *apps*, but not for a CLI process on Termux — there is no JVM, so the first TLS handshake panics.

`strings` on the binary confirms the JNI path is compiled in (`Landroid/content/Context;`, `org/rustls/platformverifier/VerificationResult`).

## Suggested fix

On Android, detect the absence of a JVM / ndk-context and fall back to `webpki-roots` (static Mozilla roots) instead of the platform verifier — the standard approach in the reqwest/rustls ecosystem for non-app Android targets. Alternatively, gate `rustls-platform-verifier` behind a non-Android cfg for CLI builds.

## Related: building for Android at all

Since no android-arm64 prebuilt is published, I built from source. Only one patch was needed: `vendor/aube/crates/aube/src/commands/global.rs` uses `#[cfg(target_os = "linux")]` for `platform_default()`, which leaves it undefined on Android (`target_os = "android"` ≠ `"linux"`). Changing it to `#[cfg(any(target_os = "linux", target_os = "android"))]` fixes the build. With that + the TLS fallback, nub appears fully functional on Termux.

Happy to test any fix on-device.

colinhacks:
On Android without a JVM (Termux), any HTTPS request aborts: rustls-platform-verifier and reqwest 0.13's default-on hickory-dns both require JNI/ndk-context, and `platform_default()` had no `android` cfg arm. Adds compile-time Android arms: webpki-roots-only TLS trust, getaddrinfo DNS, and the linux global-dir default. Byte-identical on all shipped targets.

Residual: `nub publish --provenance` builds its own clients and may still abort on Android.

Clippy/fmt/tests green; Android-only reqwest calls type-checked against pinned 0.13.4; reporter verified the equivalent change on-device.

Closes #490

https://claude.ai/code/session_017uZpUP88BDZdorSBR1H33S

### DIFF
--- a/vendor/aube/crates/aube-registry/src/client/http.rs
+++ b/vendor/aube/crates/aube-registry/src/client/http.rs
@@ -193,8 +193,11 @@ fn build_http_client_inner(
         // In-process DNS caching via hickory-dns. The system resolver
         // does not cache and uses a thread pool for `getaddrinfo`,
         // which serializes the first cold lookup per origin. hickory
-        // resolves async + caches for the process lifetime.
-        .hickory_dns(true)
+        // resolves async + caches for the process lifetime. Not on
+        // Android: hickory reads the OS DNS config through JNI and
+        // aborts without a JVM (Termux/CLI builds), so the getaddrinfo
+        // resolver stays in place there.
+        .hickory_dns(cfg!(not(target_os = "android")))
         // `strict-ssl=false` disables cert validation entirely. This
         // is a security hole on purpose: corporate registries should
         // prefer per-registry `ca` / `cafile` so validation stays on.
--- a/vendor/aube/crates/aube-util/src/http/mod.rs
+++ b/vendor/aube/crates/aube-util/src/http/mod.rs
@@ -20,23 +20,37 @@ pub mod ticket_cache;
 /// reqwest's rustls-platform-verifier OS trust store active.
 ///
 /// reqwest 0.13 can merge extra roots with the platform verifier on Unix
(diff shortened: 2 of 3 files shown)


colinhacks:

PR #502, opened 2026-07-21
pm: exclude workspace link packages from dedupe diff

### LINKED ISSUE #494: dedupe repeatedly reports workspace link packages removed without changing pnpm-lock.yaml
## Summary

   When I run `nub dedupe` in a pnpm workspace, it repeatedly reports workspace `link:` packages as removed, although `pnpm-lock.yaml` remains byte-identical.

   `nub dedupe --check` also consistently exits with code 1.

   ## Environment

   - Nub: `v0.4.13`
   - Node: `v24.18.0`
   - Platform: macOS arm64
   - Lockfile: `pnpm-lock.yaml`, version 9

   ## Minimal reproduction

   `package.json`:

   ```json
   {
     "name": "nub-dedupe-repro",
     "private": true,
     "devDependencies": {
       "@repro/a": "workspace:*"
     }
   }
   ```

   `pnpm-workspace.yaml`:

   ```yaml
   packages:
     - packages/*
   ```

   `packages/a/package.json`:

   ```json
   {
     "name": "@repro/a",
     "version": "1.0.0",
     "dependencies": {
       "@repro/b": "workspace:*"
     }
   }
   ```

   `packages/b/package.json`:

   ```json
   {
     "name": "@repro/b",
     "version": "1.0.0"
   }
   ```

   I ran:

   ```sh
   nub install
   shasum -a 256 pnpm-lock.yaml
   nub dedupe --reporter=append-only
   shasum -a 256 pnpm-lock.yaml
   nub dedupe --reporter=append-only
   shasum -a 256 pnpm-lock.yaml
   ```

   ## Actual result

   Both dedupe runs report:

   ```text
     - @repro/a@link+b076d3df9e79b8ac
     - @repro/b@link+20eb2d0e310de41f
   Dedupe: 2 removed, 0 added (net -2 packages)
   Wrote pnpm-lock.yaml
   ```

   The SHA-256 remains identical before and after each run:

   ```text
   d2fd984a49e2727635ecc2e3af0affe80db19dc18681aeb17ea09936247fcdf9
   ```

   Running:

   ```sh
   nub dedupe --check --reporter=append-only
   ```

   reports the same links and exits with code 1:

   ```text
   × dedupe --check: lockfile is not deduped
   ```

   ## Expected result

   After dedupe completes:

   - A subsequent `nub dedupe` should report no changes.
   - `nub dedupe --check` should exit with code 0.
   - Workspace links should not be reported as removed when the lockfile remains unchanged.

colinhacks:
The lockfile parser keys workspace `link:` deps as `<name>@link+<hash>`; a fresh resolve binds them by version and records no package entry, and the writer never serializes them. Dedupe's raw key diff therefore reported the links as removed on every run while the lockfile stayed byte-identical, and `dedupe --check` exited 1 permanently.

Fix: diff only entries the writer serializes (filter `link:`/`exec:` from both sides); derive the net count from the filtered sets.

Closes #494

Verified on the issue's fixture: dedupe reports "already deduped", lockfile unchanged, `--check` exits 0 — matching pnpm 10.15.1. Adds an offline workspace regression test.

### DIFF
--- a/crates/nub-cli/tests/pm_verbs.rs
+++ b/crates/nub-cli/tests/pm_verbs.rs
@@ -512,6 +512,73 @@ fn yarn_gate_refuses_mutating_verbs_and_names_the_remedy() {
     );
 }
 
+/// Workspace `link:` deps must not surface in dedupe's diff (#494): the
+/// lockfile parser synthesizes `<name>@link+<hash>` package entries that a
+/// fresh resolve never produces, so dedupe reported every workspace link as
+/// "removed" on each run — while the lockfile stayed byte-identical (the
+/// writer never serializes links) and `--check` failed forever. Offline by
+/// construction: the fixture's only deps are workspace members.
+#[test]
+fn dedupe_ignores_workspace_links_and_check_passes() {
+    let dir = pm_tmpdir("dedupe-ws");
+    std::fs::write(
+        dir.join("package.json"),
+        r#"{"name":"fixture","private":true,"devDependencies":{"@repro/a":"workspace:*"}}"#,
+    )
+    .unwrap();
+    std::fs::write(
+        dir.join("pnpm-workspace.yaml"),
+        "packages:\n  - packages/*\n",
+    )
+    .unwrap();
+    for (rel, manifest) in [
+        (
+            "packages/a",
+            r#"{"name":"@repro/a","version":"1.0.0","dependencies":{"@repro/b":"workspace:*"}}"#,
+        ),
+        ("packages/b", r#"{"name":"@repro/b","version":"1.0.0"}"#),
+    ] {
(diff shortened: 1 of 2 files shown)


colinhacks:
The broad exclusion is intentional: `Link | Exec` exactly mirrors the writer's skip set for `packages:`/`snapshots:` (`pnpm/write.rs:384`, `:737`), so the diff compares precisely what can reach the lockfile. A target-path change to an `exec:`/non-member `link:` dep only moves an importer-level `version:` line, which this diff never compared. 16226782a5 rescopes the doc comment: the parse-vs-resolve asymmetry is attributed to workspace-member links only; `exec:` is excluded on writer parity alone.

colinhacks:

PR #503, opened 2026-07-21
pm: run approved build scripts during approve-builds

### LINKED ISSUE #498: nub approve-builds should trigger rebuild automatically, like pnpm
Currently, after running `nub approve-builds -g`, users must manually run `nub -C <global-install-dir> rebuild` to apply the approved builds. This is inconvenient. It would be better if `nub approve-builds` automatically triggers a rebuild of the affected environments, similar to how pnpm handles approve-builds.

colinhacks:
Closes #498.

`approve-builds` now runs a rebuild scoped to the just-approved packages in the same invocation, matching pnpm: interactive selections get pnpm's default-No confirmation first; `--all` and positional approvals build immediately; the global path stays write-only (pnpm 11 removed `approve-builds --global`).

Also adds `allowBuilds` to `INSTALL_SHAPE_FIELDS` — an approval written to package.json previously left the warm install fast path up-to-date, so the hinted `nub install` never ran the scripts.

https://claude.ai/code/session_01UzNyZsp64TeLSjbsVbXeAF


colinhacks:

Issue #463, opened 2026-07-14
nub run orphans its workload on SIGKILL (Playwright webServer default teardown) — proposal: PR_SET_PDEATHSIG on the script child


afonsojramos:
## Summary

`nub run <script>` orphans its workload when the nub process is killed with an unforwardable signal (SIGKILL). Observed in production with Playwright, whose `webServer` teardown **defaults to SIGKILL**: Playwright kills the `nub run preview` wrapper, nub dies instantly, and the `sh -c` → vite subtree keeps running, holding the CI job open until the job-level timeout (~2h observed).

nub 0.4.11, Linux CI (Blacksmith ubuntu-2404) and macOS.

## Not a missing feature: the existing forwarding is correct but can't cover this

I checked before filing: `track_child_group` + the group-leader spawn (`group_on_spawn`) shipped in v0.4.11 (10 references in the tag's `node/spawn.rs`), and its design is right — the exact received signal forwards to the child's process group, covering the `docker stop` case its docs describe. The Playwright case defeats it because **SIGKILL is never delivered to userspace**: the forward thread never runs, so nothing reaches the group.

Workarounds we shipped in the affected repos:
- Give Playwright the real server binary instead of a `nub run` wrapper, and/or
- `gracefulShutdown: { signal: "SIGTERM", timeout: 5000 }` in `playwright.config` (then nub's existing forwarding works as designed).

## Proposed fix: parent-death signal on the script child

On Linux, set `PR_SET_PDEATHSIG(SIGTERM)` in the script child's `pre_exec` (with the standard re-check of `getppid()` after registration to close the race where the parent already died). Then even a SIGKILL on nub tears the workload down, kernel-side, with no userspace forwarding needed. Covers Playwright's default teardown, `docker kill`, CI cancellation, OOM-killer hits on the wrapper.

macOS has no pdeathsig; nearest equivalent is a kqueue `EVFILT_PROC`/`NOTE_EXIT` watcher thread in the child runtime, or simply documenting the limitation there (the high-value environment for this is Linux CI anyway).

Happy to send a PR with the `pre_exec` hook plus a kill-9-the-parent orphan-regression test (we already have a container rig for Linux validation from #453/#458 work).

colinhacks:
Fixed in v0.4.12 (now published): https://github.com/nubjs/nub/releases/tag/v0.4.12

Issue #26, opened 2026-06-19
SIGINT emitted twice on Ctrl+C


dmitri-gb:
Create `signals-test.js` containing:
```js
process.on('SIGINT', () => {
  console.log('Received SIGINT', performance.now());
});

console.log('Waiting...');

setTimeout(() => {
  console.log('Done');  
}, 5000);
```

Run it and interrupt with Ctrl+C...

```
$ node signals-test.js
Waiting...
^CReceived SIGINT 1430.523125
Done
```

```
$ nub signals-test.js
Waiting...
^CReceived SIGINT 1539.929542
Received SIGINT 1542.769334
Done
```

(Encountered on macOS.)

colinhacks:
Fixed in v0.1.6 (now published) — SIGINT is now delivered once on Ctrl+C, matching Node.

PR #504, opened 2026-07-21
fix(spawn): macOS group-reaper watcher so SIGKILL of the nub process group can't orphan the workload

### LINKED ISSUE #480: SIGKILL of nub/nubx process group leaves Node child running
## Summary

Nub places its Node child in a separate process group. If a supervisor kills the nub or nubx command process group with SIGKILL, Nub cannot forward the signal and the Node workload survives, is reparented to PID 1, and continues running.

This matters for process supervisors that send SIGTERM, wait for a grace period, then kill the supervised command group with SIGKILL. It also affects abrupt supervisor or owner death.

## Reproduction

A minimal supervisor can launch Nub as its own process-group leader, then kill that group:

```js
// repro.mjs
import { spawn } from "node:child_process"

const nub = spawn(
  "nub",
  ["-e", "setInterval(() => {}, 1000)"],
  { detached: true, stdio: "ignore" },
)
nub.unref()
console.log(nub.pid)
```

```sh
nub_pid=$(node repro.mjs)
sleep 1
node_pid=$(pgrep -P "$nub_pid")

kill -KILL -- "-$nub_pid"
sleep 1

ps -o pid=,ppid=,pgid=,command= -p "$node_pid"
```

Observed result: Node remains alive with PPID 1 and its own PGID. Cleanup requires a second kill against the child group.

I also reproduced this through `nubx vitest --watch`: killing the Nub command-root PGID left the Node process alive and reparented to PID 1.

## Expected behavior

Killing the supervised nub or nubx process group must not leave the launched workload running, including when Nub itself cannot run signal-forwarding code because it received SIGKILL.

## Cause

In `crates/nub-core/src/node/spawn.rs`, `group_on_spawn` calls `setpgid(0, 0)`, placing the Node child in its own process group. Nub forwards catchable signals to that child group, but SIGKILL cannot be caught or forwarded. A group kill aimed at the Nub command root therefore misses the child group.

## Environment

- Nub 0.4.11
- Node 26.5.0
- macOS arm64

Current main still appears to use the same separate-child-group topology.

## Possible direction

Add parent-death cleanup that does not depend on Nub executing a signal handler, or adjust process-group ownership so a supervisor group kill also reaches the workload while preserving exactly-once interactive signal delivery.

A regression test should launch Nub under a separate command PGID, SIGKILL that PGID, and assert the Node child and all descendants disappear without a second cleanup process.

colinhacks:
On macOS, SIGKILL of nub or its process group orphaned the workload: the child is its own process group (#26), SIGKILL can't be forwarded, and macOS lacks `PR_SET_PDEATHSIG` (#463's Linux backstop).

Fix: a watcher (`__pdeath-watch`) planted in the workload's group holds a pipe whose only write end lives in nub. Nub dying — any means, incl. SIGKILL — delivers EOF; the watcher SIGTERMs the group. Normal teardown writes a disarm byte first, so deliberate background survivors are left alone. Armed at all three `group_on_spawn` sites; Linux/Windows unchanged.

Verified with the issue's repro; 3 regression tests added.

Closes #480

https://claude.ai/code/session_019qrd384BeSgQHua8gqojY9

### DIFF
--- a/crates/nub-cli/src/cli.rs
+++ b/crates/nub-cli/src/cli.rs
@@ -1515,12 +1515,14 @@ fn run_nub() -> Result<i32> {
             _ => {
                 // Check if this is the first positional and matches a subcommand
                 // (nub-native, a verb registered to the embedded PM engine, or
-                // the engine's hidden node-gyp re-entry verb — its lazy shims
-                // re-invoke current_exe() with it mid-lifecycle-script).
+                // a hidden re-entry verb: the engine's lazy node-gyp shims and
+                // the macOS parent-death watcher both re-invoke current_exe()
+                // with theirs).
                 if rest.is_empty()
                     && !arg.starts_with('-')
                     && (SUBCOMMANDS.contains(&arg.as_str())
                         || arg == "__node-gyp-bootstrap"
+                        || (cfg!(unix) && arg == "__pdeath-watch")
                         || crate::pm_engine::lookup_verb(arg).is_some())
                 {
                     subcommand_found = true;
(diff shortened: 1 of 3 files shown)


colinhacks:

PR #508, opened 2026-07-21
node install: stream extraction during download, overlap checksum fetch

### LINKED ISSUE #496: nub node install 26 is slow, but n 26.5.0 is relatively fast
Installing Node.js 26 using `nub node install 26` is noticeably slow, whereas using `n 26.5.0` is much faster. Please consider optimizing the install speed.

colinhacks:
Closes #496.

The install pipeline was serial: checksums GET → full download → SHA-256 → xz decode → extract. Now the `.tar.xz` path extracts while the tarball streams in, and the `SHASUMS256.txt` fetch runs concurrently with the download. The SHA-256 gate moves to before-commit: the streamed tree stays in the quarantine `.tmp-` dir and is renamed into the store only after the checksum verifies; a mismatch wipes it. Windows `.zip` keeps download-then-extract. Kept `.tar.xz` — with decode overlapped, the smaller archive wins at any link speed; dist archives have 2 xz blocks (91% in one), so multithreaded decode was ruled out.

https://claude.ai/code/session_01TujMcGxpD3FUudua4mk97W

### DIFF
--- a/crates/nub-core/src/version_management/download.rs
+++ b/crates/nub-core/src/version_management/download.rs
@@ -356,10 +356,22 @@ fn try_download_extract(
         })
     };
 
+    let join_extractor = |h: std::thread::JoinHandle<Result<PathBuf>>| -> Result<PathBuf> {
+        h.join()
+            .unwrap_or_else(|_| bail!("extractor thread panicked unpacking {source_name}"))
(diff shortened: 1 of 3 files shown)


colinhacks:

Issue #492, opened 2026-07-20
Support neutral `packageExtensions` in Nub-owned (`nub.lock`) projects


ruff-exec:
## Feature request

Please support a neutral, top-level `packageExtensions` field for Nub-owned projects using `nub.lock`.

Nub already supports `pnpm.packageExtensions` when operating in pnpm compatibility mode, but switching to Nub's own identity removes access to this functionality. The documented neutral Nub configuration currently includes fields such as `overrides`, `resolutions`, `patchedDependencies`, and `allowBuilds`, but not `packageExtensions`.

## Motivation: Expo React Server Functions / Components

Expo projects using experimental React Server Functions or React Server Components can encounter a peer graph that cannot currently be represented safely through version overrides alone.

Example environment:

- Nub: 0.4.13
- Expo: 57.0.7
- React Native: 0.86.0
- React: 19.2.3
- React DOM: 19.2.3
- `react-server-dom-webpack`: 19.2.7

Expo SDK 57 and React Native 0.86 require React 19.2.3 because React must match React Native's embedded renderer.

However, `react-server-dom-webpack@19.2.7` declares:

```json
{
  "peerDependencies": {
    "react": "^19.2.7",
    "react-dom": "^19.2.7",
    "webpack": "^5.59.0"
  }
}
```

An npm project can currently install this graph using:

```ini
legacy-peer-deps=true
```

When attempting to use Nub, it correctly rejects this configuration:

```text
Error: nub: `legacy-peer-deps` (npm) is not supported — nub always
resolves peer dependencies; npm's legacy escape hatch would produce
a different peer graph.

remove `legacy-peer-deps` from .npmrc and fix the peer conflict, or
pin the conflicting versions in `overrides`

[ERR_NUB_UNSUPPORTED_CONFIG]
```

An `overrides` version pin does not solve this situation:

- Raising React to 19.2.7 conflicts with React Native's embedded renderer.
- Pinning `react-server-dom-webpack` does not modify its peer metadata.
- Installing another nested React version would introduce multiple interacting React instances.
- Removing all peer validation globally is undesirable.

The application already runs with this combination under npm's `legacy-peer-deps`, but there is no targeted way to document that this specific peer range is intentionally accepted after switching fully to `nub.lock`.

## Existing workaround

Keeping the project pnpm-owned allows Nub to consume:

```yaml
packageExtensions:
  react-server-dom-webpack@19.2.7:
    peerDependencies:
      react: 19.2.3
[...]

PR #509, opened 2026-07-21
fix(pm): enforce + role-scope packageExtensions under nub identity

colinhacks:
Two `packageExtensions` hardening gaps for #492, both confirmed against real pnpm. Both behaviors are gated on new `EngineContext` seams (default off / `None`), so standalone aube's behavior and its `aube-lock.yaml` bytes are unchanged.

**GAP A — drift went undetected under nub.** aube stamped pnpm's `packageExtensionsChecksum` only onto `pnpm-lock.yaml` and never read it back, so editing `packageExtensions` after a lock existed was a no-op — plain `nub install` and `--frozen-lockfile` both exited 0 on a stale graph, where real pnpm errors `ERR_PNPM_LOCKFILE_CONFIG_MISMATCH`. Now, under the `enforce_package_extensions_checksum` posture:

- the checksum is stamped onto the generic `nub.lock` too, not just `pnpm-lock.yaml`;
- `check_package_extensions_drift` compares stored vs freshly-computed checksum, wired into the frozen path (`ERR_NUB_OUTDATED_LOCKFILE`), the prefer path (re-resolve), and `--lockfile-only`;
- the no-churn write guard folds the checksum into its no-op decision, so the stamping write is never suppressed.

**GAP B — compat-role leak.** The engine read top-level `packageExtensions` unconditionally, applying it even under an npm/pnpm/yarn/bun incumbent that ignores a root `packageExtensions` — a lockfile-round-trip divergence against the mirror-the-active-PM policy. Now top-level `packageExtensions` is honored only under nub identity; under a compat role it is dropped with one dim warning (fix clause per PM), via the `embedder_package_extensions` seam that `PackageJson::package_extensions` consumes verbatim.

Migration note: an existing nub project that already applied top-level `packageExtensions` will have its first `--frozen-lockfile` fail once after upgrade (stored `None` vs a freshly-computed checksum); a normal `nub install` restamps it and the frozen install then passes — the intended pnpm-parity behavior.

Tests: drift-method and role-scoping unit tests, the engine-context seam test, the generic-lock stamp test, and a no-churn-guard regression (a checksum-only change must rewrite).

Refs #492

### DIFF
--- a/crates/nub-cli/src/pm_engine/config_scope.rs
+++ b/crates/nub-cli/src/pm_engine/config_scope.rs
@@ -336,6 +336,79 @@ pub(crate) fn scope_overrides(
     (effective, ignored)
 }
 
+/// The manifest's effective `packageExtensions`, scoped to the active `role`,
+/// plus the top-level field to warn about when it's dropped.
+///
+/// Top-level `packageExtensions` is a nub-identity neutral field: NO package
+/// manager reads it from `package.json` root (pnpm reads `pnpm.packageExtensions`
+/// / `pnpm-workspace.yaml`; yarn reads `.yarnrc.yml`; npm and bun have no such
+/// feature). So nub honors the top-level home ONLY under its own identity;
+/// under an npm/pnpm/yarn/bun compat role it applies exactly what the incumbent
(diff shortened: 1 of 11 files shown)


colinhacks:

PR #510, opened 2026-07-21
fix(watch): forward explicit --env-file to the watched Node so restarts re-read it

### LINKED ISSUE #479: --watch restarts keep stale values from explicit --env-file
With `nub --env-file=<file> --watch`, the env file's values are frozen at startup. When the watcher restarts the script, edits to the file are not
picked up: changed values stay stale and keys added after startup never appear. Plain `node --env-file=<file> --watch` re-reads the file on every restart.

Auto-discovered `.env` files are not affected (they live-reload across restarts). Only the explicit `--env-file` / `--env-file-if-exists` flags freeze.

## Repro

`index.js`

```js
console.log('FOO =', process.env.FOO, '| BAR =', process.env.BAR)
```

`custom.env` (named outside the `.env*` cascade so auto-discovery stays out of the picture)

```
FOO=one
```

Run:

```sh
nub --env-file=custom.env --watch --watch-path=. index.js
```

then, from a second shell:

```sh
printf 'FOO=two\nBAR=three\n' > custom.env
```

Observed: the watcher restarts, but the restarted script prints the startup values.

```
FOO = one | BAR = undefined
Completed running 'index.js'. Waiting for file changes before restarting...
Restarting 'index.js'
FOO = one | BAR = undefined
```

Expected: what `node --env-file=custom.env --watch --watch-path=. index.js` prints after the same edit.

```
FOO = one | BAR = undefined
Completed running 'index.js'. Waiting for file changes before restarting...
Restarting 'index.js'
FOO = two | BAR = three
```

The auto-discovery counterpart behaves correctly: drop the flag, rename `custom.env` to `.env`, and the same edit is picked up on the restart.

Versions: nub v0.4.11, node v24.13.0, macOS (Darwin arm64).

colinhacks:
Closes #479.

Explicit `--env-file`/`--env-file-if-exists` values were injected into the long-lived watch supervisor once at startup, so edits neither triggered a restart nor reached restarted children. The paths now forward to the watched Node as real `--env-file*` args — the channel auto-discovered `.env*` files already use (#207) — so Node watches and re-reads them per restart; nub injects only `${VAR}`-expansion-changed keys. Gated on Node accepting the flags (20.6+/22.9+), with the old inject fallback below.

Unit tests + e2e on the issue's repro; differential matches `node --env-file --watch`.

https://claude.ai/code/session_01G96dWWt1soJ7kXnAp3N96X

### DIFF
--- a/crates/nub-cli/src/cli.rs
+++ b/crates/nub-cli/src/cli.rs
@@ -155,6 +155,24 @@ fn emit_ndjson_summary(passed: usize, failed: usize) {
 /// dependency spun up a thread during init (A19). Set once; never mutated after.
 static ENV_FILE_VARS: OnceLock<HashMap<String, String>> = OnceLock::new();
 
+/// Raw (unexpanded, un-stripped) merge of the explicit `--env-file` contents —
+/// the values Node's own `--env-file` parser delivers. The watch path forwards
+/// the explicit files to the watched Node (#479) and compares [`ENV_FILE_VARS`]
+/// against this map to inject only keys whose `${VAR}` expansion changed the
+/// raw value (see [`watch_inject_vars`]); every plain var is left to Node's
+/// re-read so it live-reloads across restarts.
+static ENV_FILE_VARS_RAW: OnceLock<HashMap<String, String>> = OnceLock::new();
+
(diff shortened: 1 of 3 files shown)


colinhacks:

PR #511, opened 2026-07-21
linker: workspace-spanning hoisted planning for nodeLinker=hoisted

### LINKED ISSUE #484: Warn or error on pnpm's `nodeLinker: hoisted` instead of silently mislinking
When a pnpm workspace sets `nodeLinker: hoisted`, `nub install` (0.4.12) neither honours it nor falls back cleanly to the GVS layout: it materialises a per-project peer-variant subtree under a workspace library (~253 modules, including a second copy of `react` at the same version), which breaks React with the two-instance invalid-hook error. On the same repo without the hoisted setting, nub's layout is correct and the full test suite passes, so this is only about the unsupported setting being ignored silently. A warning (or hard error) when the lockfile/workspace requests an unsupported `nodeLinker` would have saved a long debugging session.

Repro: https://github.com/afonsojramos/super-calendar at any commit before [`43e0d0b`](https://github.com/afonsojramos/super-calendar/commit/43e0d0b) (e.g. `git checkout 43e0d0b~1`), then:

```sh
nub install
ls packages/native/node_modules | wc -l   # ~253 entries, including a react copy
nub run test                              # 17 of 31 jest suites fail with invalid-hook errors
```

Environment: nub 0.4.12, node v24.18.0, macOS, `lockfileVersion: '9.0'`.

colinhacks:
Under `nodeLinker=hoisted` with the default `hoistingLimits=none`, the hoisted linker planned placement per physical importer: every workspace member materialized its full transitive closure under its own `node_modules`, so a dependency shared across members (e.g. `react`) was duplicated on disk. Two copies of `react` meant two module identities and the reporter's invalid-hook errors.

Real pnpm's hoisted default runs one hoist over the whole workspace: a dependency at a single version lives once at the workspace-root `node_modules`, and only version conflicts nest under the requiring member. This change makes nub match that.

- `hoistingLimits=none` (default) plans one workspace-spanning tree; a shared dep hoists to the root once, conflicts nest under the member.
- `hoistingLimits=workspaces` now genuinely borders at each member (per-importer trees); `dependencies` is unchanged; single-package projects are unchanged.
- `link:` workspace siblings stay member-local (pnpm's `hoistWorkspacePackages` layout), not hoisted to the root.
- The install-state hash folds a hoisted-layout marker so a tree written by the old per-importer algorithm is relinked rather than treated as current.

This is a latent-bug fix: aube's hoisted mode documents npm/yarn-classic flat-tree parity, npm hoists workspace-wide to the root, and aube's own docs marked the per-importer behavior "currently equivalent to None."

## Verification

Differential against pnpm 11 on the reporter's repo (super-calendar @ 43e0d0b~1):

| metric | before | after | pnpm 11 |
|---|---|---|---|
| physical `react@19.2.7` copies | 7 | 1 | 1 |
| `packages/native/node_modules` entries | 253 | 1 | 1 |
| `require.resolve('react')` root vs native | different | same dir | same dir |

A minimal two-member workspace (shared dep + one conflicting version) matches pnpm on entry counts, physical copy count, and resolve identity from root and each member. Single-package hoisted installs are unchanged.

Tests: workspace-spanning unit cases in `hoisted.rs` (shared hoists to root once; conflict nests under the member; `link:` sibling stays member-local; `from_graph` recompute) plus an on-disk `link_workspace` materialization test in `tests.rs`.

Closes #484

https://claude.ai/code/session_01EDXu8722BkCGvBYYu549sn

### DIFF
--- a/vendor/aube/crates/aube-linker/src/hoisted.rs
+++ b/vendor/aube/crates/aube-linker/src/hoisted.rs
@@ -72,17 +72,28 @@ impl HoistedPlacements {
         hoisting_limits: HoistingLimits,
     ) -> Result<Self, Error> {
         let mut placements = Self::default();
-        for (importer_path, deps) in &graph.importers {
-            if !crate::is_physical_importer(importer_path) {
-                continue;
-            }
-            let importer_dir = if importer_path == "." {
-                root_dir.to_path_buf()
+        let importer_nm = |importer_path: &str| -> PathBuf {
+            if importer_path == "." {
+                root_dir.join(modules_dir_name)
             } else {
-                root_dir.join(importer_path)
-            };
-            let nm = importer_dir.join(modules_dir_name);
-            let plan = plan_importer(&nm, deps, graph, hoisting_limits)?;
+                // Match the link path's lexical `..` collapse so a
+                // `../sibling` importer key recomputes the same on-disk dir.
+                aube_util::path::normalize_lexical(
+                    &root_dir.join(importer_path).join(modules_dir_name),
+                )
+            }
+        };
+        let physical: Vec<(&String, &Vec<DirectDep>)> = graph
+            .importers
+            .iter()
+            .filter(|(p, _)| crate::is_physical_importer(p))
+            .collect();
+
+        // Match the linker: the default (`None`) hoists a multi-importer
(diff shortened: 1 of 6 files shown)


colinhacks:

PR #512, opened 2026-07-22
pm: tri-state interactive update picker

### LINKED ISSUE #483: cli-update interactive colors
It would be nice to have a table and color ui for the interactive update with nub here is examples of bun:

Image

thinking of it would be cool to support "beta" for example effect version (also had the same issue with typescript nighlty) version unless it adds too much code to handle these weird beta version resolution with not many people using this feature

colinhacks:
One row per outdated direct dep in `nub update -i`; space/←→ cycles keep → in-range target → latest, as aligned radio cells with semver-diff coloring. Rows default to keep. A latest below a prerelease pin is never offered. `-i --latest` is the same picker. Gated by a new `Embedder::rich_update_picker` flag: nub opts in, standalone aube unchanged.

Closes #483

Verification: aube suite green (760 passed) + new picker unit tests; driven e2e in a real terminal (mixed picks rewrite manifest/lockfile per row, cancel exits 130 unmutated); clippy + fmt clean.

https://claude.ai/code/session_01AwLJZRoRYeTXcTibVB8dQw


colinhacks:

PR #513, opened 2026-07-22
fix(pm): clamp identity walk-up to nub's cache root for internal scratch installs

### LINKED ISSUE #489: __node-gyp-bootstrap re-entry install walks up past its own cache dir into $HOME, hard-failing on unrelated lockfile ambiguity
## Summary

`nub install` (and other verbs in the mutating install family) can hard-fail with `ERR_NUB_LOCKFILE_AMBIGUOUS` for a reason that has nothing to do with the project being installed: the internal `__node-gyp-bootstrap` re-entry verb (mentioned in #1 — "allowlisted node-gyp builds bootstrap a real node-gyp under `$XDG_CACHE_HOME/nub/pm/tools/`") runs a recursive install *inside that cache directory*. The first time it runs, that scratch directory has a `package.json` + `pnpm-workspace.yaml` but no lockfile of its own yet — and the recursive install's identity resolution responds to that by walking up the filesystem looking for project context, exactly like it would for a real workspace member. It walks straight past `$XDG_CACHE_HOME` and lands in `$HOME`, where it can find unrelated lockfiles from a completely different, unrelated setup and refuse to proceed.

This makes `nub install` unusable — deterministically, on every invocation — on any machine where `$HOME` happens to contain more than one kind of lockfile (e.g. a personal global `package.json` scratch space, leftover `bun.lock`/`package-lock.json` from unrelated experiments), even though the actual project being installed has a single, unambiguous lockfile and a declared `packageManager`.

Confirmed reproducible across every release from **v0.2.10 through v0.4.13** (today's latest) — roughly two weeks / a dozen releases with no change to this path.

## Environment

- nub `v0.4.13` (also reproduced on v0.2.10, v0.4.2, v0.4.11 along the way)
- Linux x86_64 (Fedora-based), Node v26.5.0 (resolved from `.nvmrc`)
- Project: a private Node.js/TypeScript repo with a normal single `pnpm-lock.yaml` + `packageManager: "pnpm@11.10.0+..."` declared in `package.json` — no ambiguity in the project itself.
- `$HOME` (unrelated to my project) happens to contain three different lockfiles from other, unrelated tooling/experiments: `~/pnpm-lock.yaml`, `~/bun.lock`, `~/package-lock.json`, alongside a small `~/package.json`.

## Reproduction

1. In a project whose install needs to bootstrap node-gyp for some dependency (I wasn't able to pin down the exact triggering package from nub's own output — even at `--verbose --verbose --verbose` it doesn't log which lifecycle step requested the bootstrap).
2. Have `$HOME` contain 2+ different lockfile formats (unrelated to the project) — a common state for any dev machine that has ever tried more than one package manager globally.
3. Run `nub install` (no special flags) in the project.

## Actual behavior

```
Error: ERR_NUB_LOCKFILE_AMBIGUOUS

  × multiple lockfiles found: pnpm-lock.yaml, bun.lock, package-lock.json —
  │ cannot tell which package manager owns this project
  help: set the declaration: nub pm use <pm> — or remove the stale lockfile

  × recursive nub install failed while bootstrapping node-gyp (exit 1) — pre-
  │ populate /home/<user>/.cache/nub/pm/tools/node-gyp/v12 or run `nub
  │ install` once while online
```

The `install` fails completely — nothing gets linked, no lockfile is written — even though the ambiguity is 100% in `$HOME`, not in the project.

### I confirmed the exact mechanism by pre-populating the scratch dir myself

```
$ cat ~/.cache/nub/pm/tools/node-gyp/v12/package.json
{"name":"aube-tool-node-gyp","private":true,"dependencies":{"node-gyp":"^12.0.0"}}
$ cat ~/.cache/nub/pm/tools/node-gyp/v12/pnpm-workspace.yaml
(empty)
$ ls ~/.cache/nub/pm/tools/node-gyp/v12/
package.json  pnpm-workspace.yaml        # <- no lockfile, no node_modules yet
```

Running a real (non-nub) `pnpm install` **directly inside that scratch directory** — nothing touched outside `$XDG_CACHE_HOME/nub/pm/tools/node-gyp/v12/` — populates `node_modules/node-gyp` and writes a `pnpm-lock.yaml` there. After that, `nub install` in the actual project succeeds end-to-end (all lifecycle scripts included, no `--ignore-scripts` needed):

```
$ cd ~/.cache/nub/pm/tools/node-gyp/v12 && pnpm install
Packages: +20
dependencies:
+ node-gyp 12.4.0
Done in 952ms using pnpm v11.10.0

$ cd /path/to/project && nub install
nub 0.4.13
✓ resolved 1223 · reused 1223
... (full install completes, prepare hook runs, no error)
```

This isolates the bug precisely: the recursive install **only** fails when that specific scratch directory has no lockfile of its own yet (first-time provisioning, or after nub bumps its pinned node-gyp version and the subdirectory changes, e.g. `v12` → `v13`). Once it has one, this specific trigger never recurs — but the underlying identity-walk-up behavior is still there, waiting for the next internal re-entry verb that hits a fresh scratch directory.

### Why I think this is a bug, not intended behavior

#197 and #199 already established the precedent that lockfile ambiguity should only hard-fail for verbs that actually read/write *the project's own* lockfile (the "mutating install family"), and added a lenient/transient identity-resolution mode specifically so internal/global-scope operations (`dlx`, `nubx`, `create`, `store`/`cache`/`config`, `view`/`bin`/`root`, etc.) don't get caught by an ambiguity that has nothing to do with what they're doing.

The `__node-gyp-bootstrap` re-entry verb seems to have been missed by that fix, and it's arguably the clearest case for the lenient path: it never touches the *calling* project's lockfile at all, and it never needs project discovery in the first place — nub always wants pnpm for this internal scratch install, deterministically, by its own embedder defaults. There's no reason for its identity resolution to escape `$XDG_CACHE_HOME/nub/pm/tools/` at all, let alone escalate all the way to `$HOME` and hard-fail the outer install over lockfiles the user has never associated with this project.

(Related but distinct, for context: #457, fixed in v0.4.12, was also about a lifecycle script running against the wrong directory context — a different bug, but the same general area of "internal/store-relative execution context" having rough edges.)

## Suggested fix

Route `__node-gyp-bootstrap`'s recursive `install` through the same lenient/transient identity-resolution session introduced in #197 — or, more targeted: skip identity discovery entirely for this specific re-entry verb and hardcode it to nub's own default pnpm identity, since it's always installing the same internal `node-gyp` scratch package regardless of what project triggered it. There's no ambiguity to resolve here in the first place; the walk-up should never have been reached.

Happy to test a fix against this repro if useful.

colinhacks:
The node-gyp bootstrap's recursive `nub install` runs in a fresh cache scratch dir with no lockfile, so the identity walk-up climbed past the cache root and could hard-fail on unrelated lockfile ambiguity in `$HOME` (missed by the #197/#199 lenient-identity scoping — this re-entry runs the strict install family).

The walk now stops at nub's PM cache root: installs inside it are nub-internal by construction and never inherit ambient identity.

Closes #489

Verified: regression test fails pre-fix / passes post-fix; e2e bootstrap install completes under an ambiguous `$HOME`; clippy/fmt/pm_identity green.

https://claude.ai/code/session_01Y8uY2mCdhEHhWzw6WuMfrR

### DIFF
--- a/crates/nub-cli/src/pm_engine/mod.rs
+++ b/crates/nub-cli/src/pm_engine/mod.rs
@@ -1465,6 +1465,22 @@ fn resolve_identity_walk_up(
     strictness: IdentityStrictness,
 ) -> Result<Option<DetectedLockfile>> {
     use aube_lockfile::ResolvedLockfileKind;
+    // The walk never escapes nub's own PM cache root. Installs running inside
+    // it are nub-internal by construction (the node-gyp bootstrap's recursive
+    // install, dlx scratch dirs) and must not inherit identity from whatever
+    // sits above the cache — unbounded, a first-run bootstrap dir (manifest,
+    // no lockfile yet) walked into $HOME and hard-failed the outer install on
+    // unrelated-lockfile ambiguity (#489). The root is kept in whichever
+    // spelling is an ancestor of `cwd` (raw, or canonicalized for the
+    // symlinked-temp-dir case) so the containment test stays consistent as
+    // `dir` pops.
+    let clamp = aube_store::dirs::cache_dir().and_then(|root| {
+        if cwd.starts_with(&root) {
+            return Some(root);
+        }
+        let canon = std::fs::canonicalize(&root).ok()?;
+        cwd.starts_with(&canon).then_some(canon)
+    });
     let mut dir = cwd.to_path_buf();
     for _ in 0..16 {
         match aube_lockfile::resolve_project_lockfile_kind(&dir) {
@@ -1493,6 +1509,11 @@ fn resolve_identity_walk_up(
             Err(err) => return Err(identity_error(err)),
         }
         if !dir.pop() {
+            break;
+        }
+        if let Some(root) = &clamp
+            && !dir.starts_with(root)
+        {
             break;
         }
     }
--- a/crates/nub-cli/tests/pm_identity.rs
+++ b/crates/nub-cli/tests/pm_identity.rs
@@ -336,6 +336,72 @@ fn global_scope_commands_ignore_multi_lockfile_ambiguity() {
     );
 }
 
+/// The internal-re-entry corner of the #197/#199 class (#489): the node-gyp
+/// bootstrap runs a recursive `nub install` inside the PM cache
+/// (`$XDG_CACHE_HOME/nub/pm/tools/node-gyp/<bucket>`), whose first run has a
+/// manifest but no lockfile yet. The identity walk-up must stop at nub's cache
+/// root instead of climbing into ancestor dirs ($HOME on a real machine) and
+/// hard-failing on lockfile ambiguity that has nothing to do with the install.
+#[test]
+fn cache_scratch_installs_never_inherit_ambient_identity() {
+    // Ambiguous ANCESTOR of the cache root — two lockfile families, as a
+    // $HOME with leftovers from unrelated experiments would have.
+    let dir = project("cache-clamp", r#"{"name":"app","version":"1.0.0"}"#);
+    std::fs::write(
+        dir.join("package-lock.json"),
+        r#"{"name":"app","version":"1.0.0","lockfileVersion":3,"requires":true,"packages":{}}"#,
+    )
+    .unwrap();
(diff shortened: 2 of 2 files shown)


colinhacks:

PR #515, opened 2026-07-22
fix(init): preserve scoped package names in sanitize_name

afonsojramos:
## Summary

`nub init --name @scope/pkg` no longer mangles the name to `scopepkg` — scoped names are sanitized per-part and preserved (`@My Org/My App` → `@my-org/my-app`), and a bare `@scope` with no package part is rejected as invalid instead of silently becoming `scope`. Raised in review on #499 after that PR had merged.

```bash
nub init --name "@myorg/api"   # before: "name": "myorgapi"  →  now: "name": "@myorg/api"
nub init --name "@myorg"       # before: "name": "myorg"     →  now: error: not a valid package name
```

Two smaller scaffold fixes ride along, also from the #499 review:

- The scaffolded `.gitignore` no longer ignores `.env.example` — `.env*` is replaced with the explicit bun-style list (`.env`, `.env.local`, `.env.*.local`), which also keeps shared `.env.production`/`.env.development` files tracked.
- `init.mdx` now notes that swapping the start script to plain `node` requires Node ≥22.18 (type stripping), and that `--name` preserves scopes.

Empty-from-sanitize stays the invalid signal, so the interactive validator and the `--name` bail in `run_init` are unchanged; no call-site changes.

## Test plan

- New unit test `sanitize_scoped_names_sanitize_per_part_and_reject_bare_scopes`: scope preserved, per-part sanitization, bare `@scope` → invalid, `@/pkg` → invalid, emoji-scope → invalid. Existing sanitize tests unchanged and passing.
- Full `init_cmd` integration suite (6 tests) green; `cargo test -p nub-cli` green; clippy and fmt clean.
- Manual smoke: `--name "@My Org/My App"` scaffolds `"name": "@my-org/my-app"`; `--name "@scope"` errors cleanly; scaffolded `.gitignore` verified with `git check-ignore` (`.env`, `.env.local`, `.env.development.local` ignored; `.env.example`, `.env.production` tracked).

## Post-Deploy Monitoring & Validation

No additional operational monitoring required — local CLI scaffold behavior with no service, telemetry, or data-path impact; unit and integration tests pin the behavior in CI.

Related: #499


afonsojramos:
@colinhacks this one is quite small really

colinhacks:

PR #516, opened 2026-07-22
pm: accept dist-tag specs in update args

colinhacks:
`nub update typescript@beta` (any registry dist-tag) now resolves through the tag, matching pnpm — previously only `latest` was accepted. A non-`latest` tag writes an exact pin (`~5.3.0` → `"6.0.0-beta"`); `@latest` and version/range args keep the operator-gluing rewrite. Nonexistent tags fail with `ERR_NUB_NO_MATCHING_VERSION`; protocol/alias/URL specs stay rejected.

Refs #483

Verification: differential vs pnpm 10.15.1 (`chalk@latest` → `^5.6.2`, `semver@7.6.0` → `~7.6.0`, `typescript@beta` → `6.0.0-beta`, all matching); aube suite green (754), clippy + fmt clean.

https://claude.ai/code/session_01AwLJZRoRYeTXcTibVB8dQw


colinhacks:

PR #517, opened 2026-07-22
fix(spawn): dispatch the parent-death watcher verb above argv0 detection

colinhacks:
Trunk CI has been red on `Test (macos-14, node 24)` since #504: 87/208 integration tests failing with `EAGAIN (os error 35)` on spawn, the process table exhausted.

The watcher re-invokes `current_exe()` with the hidden `__pdeath-watch` verb, dispatched only on the `nub` argv0 arm. Under nub's own PATH shim `current_exe()` is `node`, so `run_as_node()` ran it as a script — spawning a workload plus another watcher per level, each then blocked in `GroupReaper::drop` on the one below.

Fix: dispatch it at the top of `cli::run()`, above `Argv0::detect()`.

Peak watchers over one integration run, same host: 3516 and climbing before, 1 after. Full suite after: 208 passed.

Refs #504, #480

### DIFF
--- a/crates/nub-cli/src/cli.rs
+++ b/crates/nub-cli/src/cli.rs
@@ -1079,6 +1079,26 @@ pub enum ColorWhen {
 
 /// Top-level entry point. Returns the process exit code.
 pub fn run() -> Result<i32> {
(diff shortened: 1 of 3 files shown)


colinhacks:

PR #519, opened 2026-07-22
aube: sync vendored engine to upstream v1.32.0

colinhacks:
Syncs the vendored engine from aube v1.25.1 to v1.32.0 — 133 upstream commits.

## Merge base

`vendor/aube` was at v1.25.1 while the `nub-fork` merge venue was at v1.23.0: an earlier sync went straight into `vendor/aube` and bypassed the venue, so git had been merging against a two-release-stale base.

Re-parenting the reconciliation snapshot onto upstream's v1.25.1 release commit:

| base | conflicted files | hunks |
| --- | --- | --- |
| v1.23.0 (stale) | 72 | 205 |
| v1.25.1 (correct) | 23 | 48 |

`Cargo.lock`, the changelogs, and generated docs auto-merge at the correct base. `vendor/aube/UPSTREAM` now records the base commit so the next bump reads it instead of inferring it from a version string.

## Resolution

Preserved: GVS/phantom linking, force-materialization, concurrent OSV and trust-policy gating, the `defaultTrust` floor, embedder-profile plumbing, nub-identity config scoping, pnpm-11 `namedRegistries`, patch-group range resolution.

Grafted from upstream: plain unified patch parsing (jdx/aube#1021), `file:`/`portal:` stale-materialization removal, proxy passthrough to lifecycle scripts (ordered so nub's env overlay still applies last), `rustls`/`hickory-dns` feature gating with nub's Android carve-out, stale `(patch_hash=…)` stripping, hash-only patch decoration (jdx/aube#1035), the berry patch-selector superset, and upstream's install cancellation points re-sited into nub's concurrent structure.

Seven further defects were found by the compiler, not by conflict markers — a new required `parse_one` parameter, upstream's new `access.rs` written against a signature nub had changed, a `TrustExcludeRule` field rename under a nub-added method, a new progress `Mode::Events` variant, and two new required struct fields.

## Two pre-existing bugs fixed

Both predate this merge; both are fixed here because the merge widens them.

1. `decide_with_floor` dropped `source_key`, so the build-approval warning seam disagreed with the runner about `file:`/git/tarball-backed packages — a bare-name `allowBuilds` approval vouched for bytes the runner refuses. Upstream v1.32 moves all three seams to `decide_package_with_git_repository`, which would widen the gap to two missing keys.
2. The pnpm writer required `patched_dependencies` to contain the source key, but nub's pnpm reader leaves that map empty by design, so every parse and re-emit dropped `(patch_hash=…)` suffixes and changed dep-path identity.

## nub-side

jdx/aube#1068 and #1069 feature-gated the TLS backend and the publish stack, so `crates/nub-cli/Cargo.toml` now names what it needs:

- `aube` → `rustls`, `publish`. `publish` gates `commands::publish`, which `pm_engine/publish_family.rs` calls directly; `rustls` is the crate's only TLS backend. Without them nub-cli does not compile.
- `aube-registry` → `hickory-dns`. Default-preserving: this crate called `.hickory_dns(cfg!(not(target_os = "android")))` unconditionally before v1.32.
- `hickory-dns` is deliberately left off the `aube` crate — reqwest's feature flips the default resolver for every client in the binary.

## Verification

- aube: `cargo check --workspace --all-targets`, `clippy --all-targets --all-features -- -D warnings`, 2,590 tests — all green.
- nub: `check -p nub-cli --all-targets`, `clippy --all-targets --all-features -- -D warnings`, `fmt --check`, and 443 tests across the 15 sync-relevant binaries — all green. The full suite was not run in one pass (`pm_two_mode` alone takes 41 minutes); CI covers it.
- `aube-util`'s new cfg partition checked under no-default-features, `rustls`, and the Windows target. The Android branch is exhaustive and mutually exclusive by construction.
- The 18 files failing `cargo fmt --check` inside `vendor/aube` are an identical set before and after this merge. nub's root workspace excludes `vendor/aube`, so they do not gate nub CI.

## Patch handling brought to pnpm parity

Two further pnpm-compat gaps, both confirmed against real pnpm 10.15.1 rather than inferred.

**Patches now apply to npm-aliased installs.** `patchedDependencies: { "is-odd@3.0.1": … }` did not patch `odd-alias: npm:is-odd@3.0.1`; pnpm does, because it resolves patches against the resolved manifest's name, where the alias is not a distinct node. The rule lives in one place — `PatchGroups::resolve_package`, called by both the resolver and the pnpm writer, keyed by `spec_key()` so the linker's `Patches` map, the graph hash, and the delta fingerprint all inherit it. The no-churn write guard carried a second copy of the matching logic and now routes through the same entry point.

Resolution is by registry name only. A key naming the *alias* matches nothing, because pnpm can never match one either.

aube v1.32.0 has the same gap, and additionally records `(patch_hash=…)` on the alias's dep path — so its lockfile claims a patched identity the install does not produce. Reported upstream as jdx/aube discussion #1082.

Fixing this exposed a second, pre-existing reader bug: `pnpm/read.rs` baked the write-time `(patch_hash=…)` marker into parsed dep paths, so a parsed patched graph could never hash-match a fresh resolve and the no-churn guard could never fire. The reader now strips the marker and the writer re-derives it — one owner per direction, writer output unchanged.

**An unused patch key now fails the install.** A `patchedDependencies` key matching no installed package was silently ignored; pnpm exits 1 with no `node_modules` and no lockfile, even when the dead key sits beside keys that do apply. Nub now matches, with the message text byte-identical to pnpm's, and the gate in the same place pnpm runs it (inside the resolution step, so a warm install whose resolution is skipped does not re-check).

`allowUnusedPatches` downgrades it to a warning. All seven candidate spellings were probed against pnpm 10.15.1: this is a manifest and workspace-YAML setting and `.npmrc` is not a home for it, so nub reads it from exactly the homes `patchedDependencies` uses. The existing config gating then gives the brand boundary for free — a pnpm-incumbent project reads `pnpm.allowUnusedPatches`, a nub-identity project reads the neutral top-level key and ignores the branded one, the same shape `allowBuilds` already has. pnpm 10's deprecated `allowNonAppliedPatches` spelling is accepted.

**This fails installs that previously succeeded.** A typo, a removed dependency, or a version bump that outran a patch selector now fails on the next resolving install. That is the intended parity, but it is the user-visible blast radius of this PR.

## Flagged for review

- The `decide_with_floor` fix is a security-posture change riding in a sync PR.
- The unused-patch-key gate above can fail installs that work today.
- One new upstream bats test (`registry dist.shasum with no dist.integrity is derived`) is skipped: it greps `aube-lock.yaml`, which nub does not write. The derivation it covers is present in the merged resolver and unit-covered.
- Open PRs touching `vendor/aube/**` (#512, #442, #419) overlap the upstream delta and will conflict; whichever merges second resolves.


colinhacks:

PR #529, opened 2026-07-22
fix(pm): export npm_config_user_agent to dlx/create/exec children

colinhacks:
dlx/exec child spawns set `npm_node_execpath` but never `npm_config_user_agent`, so create-* scaffolders under `nub create` fell back to npm-mode output. pnpm exports the UA to these children as it does to lifecycle scripts. `runtime::apply_child_env` now exports the role-aware UA; `nub create vue` prints `nub install` / `nub run dev`. Regression test on the dlx path.

https://claude.ai/code/session_01XeVYy7YT1M3wrVzrwwirbS


colinhacks:

PR #530, opened 2026-07-22
fix(pm): key the GVS engine fingerprint off the project's Node, not the shell's

colinhacks:
## The bug

nub's embedder profile sets `runtime_switching: false` — nub owns Node provisioning, so aube's runtime resolver stays inert and `runtime::current()` carries no version. `engines::effective_node_version` then fell through to its ambient `node --version` PATH probe, and that version's **major** is what `engine_name_default` folds into the GVS (global virtual store) subdir hash of every subtree containing a build-allowed package.

Meanwhile lifecycle scripts already run under the project's **pinned** Node: `apply_lifecycle_augmentation` resolves it via `discover_node` (pin-aware — `.nvmrc` / `.node-version` / `engines`) and pins `npm_node_execpath` to it. So dependency build scripts compiled against the project's Node while the store key recorded the developer's shell Node — two majors could share one virtual-store entry, and the same project rehashed whenever the shell Node changed.

## Empirical reproduction

One pinned project (`.nvmrc` → `22.15.0`, one dependency with a build script), installed twice with two different ambient shell Node majors on PATH.

```
# before (main): the GVS subdir hash tracks the SHELL node, not the pin
ambient node v22.15.0  → node_modules/.store/core-js@3.39.0 -> …/store/core-js@3.39.0-e481e56a79d379fc
ambient node v26.5.0   → node_modules/.store/core-js@3.39.0 -> …/store/core-js@3.39.0-b9c157bb9ee7a614   # ❌ different entry, same pin
lockfile sha256: 0d2735… (identical across both)

# after: the hash tracks the project's pinned node, stable across shells
ambient node v22.15.0  → …/store/core-js@3.39.0-e481e56a79d379fc
ambient node v26.5.0   → …/store/core-js@3.39.0-e481e56a79d379fc   # ✓ same entry
lockfile sha256: 0d2735… (unchanged)
```

The lockfile bytes are identical in every arm, before and after: `graph_identity_hash` is deliberately engine-agnostic (`engine: None`), so the engine taint never reaches the recorded lockfile identity.

## The fix

Carry the resolved version across the existing per-invocation embedder seam. A new `EngineContext::runtime_node_version: Option<String>` is set beside `runtime_node_bin` from the same `discover_node` call, and consulted by `effective_node_version` at the resolver's tier: below an explicit `.npmrc` `node-version` override (pnpm's validation-only knob keeps winning), above the PATH probe.

The version is published **above** `apply_lifecycle_augmentation`'s early returns, since `--node` / `NODE_COMPAT` disable augmentation but not version provisioning — the store key should still reflect the project's pinned Node in compat mode. It is left unset when discovery itself fails, so the PATH probe still answers for the bare `node` those scripts would fall back to.

This also corrects the `engines.node` check, which aube's own doc comment says must validate the Node scripts will actually run on, not whatever PATH happened to carry. The check is warn-only unless the project sets `engine-strict`.

### Why not `seed_embedder_node`

The v1.32 sync added `runtime::seed_embedder_node`, but it is the wrong mechanism here: it fabricates a full `RuntimeContext` whose provenance is hardcoded `Mise` (so `nub doctor` would report the Node as "provided by mise"), derives `node_bin` from a single bin dir when nub's shim dir and provisioned node are different paths, re-spawns `node --version` though nub already holds the version, and only reaches `install` / `ci`. The `EngineContext` seam runs once per command from `engine_session`, so it covers every engine-routed verb, and it carries the version nub already resolved.

## Invalidation cost

One-time, and cheap: a **re-link** of the affected subtrees. No re-download, and no re-run of build scripts. The content-addressable store is keyed by integrity, so the package bytes are already present and the subtree re-materializes under the corrected hash from the CAS; the side-effects cache is keyed independently of the engine name, so previously-built outputs are restored rather than rebuilt. Verified against an isolated cache: after flipping the engine and clearing `node_modules`, `nub install --offline` exits 0 with `side-effects-cache: restored` in the log, and the CAS is unchanged in size.

## Fork discipline

Default-preserving for standalone aube: `runtime_node_version` defaults to `None`, the new tier is skipped, and every existing aube path is byte-for-byte unchanged (the `default_is_upstream_neutral` test carries the new field).

## Relationship to the v1.32 aube sync (#519)

The bug persists unchanged after the v1.32 sync — `seed_embedder_node` arriving upstream does not fix it unless nub opts in — so this is not made redundant by #519. `effective_node_version` and the two touched `EngineContext` neighbors are byte-identical on both branches, so this merges cleanly with #519 in either landing order.

## Tests

`engines.rs` gains one unit test pinning the contract: with `runtime_node_version` set, `effective_node_version` returns it rather than the PATH probe, and the explicit `.npmrc` override still outranks it. The existing upstream-neutrality assertion is extended to cover the new field.

### DIFF
--- a/crates/nub-cli/src/pm_engine/mod.rs
+++ b/crates/nub-cli/src/pm_engine/mod.rs
@@ -1385,18 +1385,31 @@ fn augmentation_to_lifecycle_overlay(
 /// Install nub's runtime augmentation onto the engine's lifecycle-script spawn
 /// env (via aube's generic [`aube::set_script_settings`] overlay), so dependency
 /// build scripts run under the project's provisioned + augmented Node — the same
-/// env `nub run` / `nub exec` give scripts. No-op (overlay stays default-empty,
-/// behavior preserved) when augmentation can't be computed (compat / re-entrant
-/// / broken install). Called once per command from [`engine_session`].
+/// env `nub run` / `nub exec` give scripts. The overlay stays default-empty
+/// (behavior preserved) when augmentation can't be computed (compat /
+/// re-entrant / broken install); the resolved Node *version* is published to the
+/// engine either way. Called once per command from [`engine_session`].
 fn apply_lifecycle_augmentation(cwd: &Path) {
-    let Ok(nub_binary) = nub_core::node::spawn::current_nub_binary() else {
-        return;
-    };
     // The project's Node — pin-aware (`.nvmrc`/`.node-version`/`engines`), NOT
     // the ambient PATH node. This resolved version drives flag injection and its
     // path pins npm_node_execpath. Mirrors build_script_command's discovery.
-    let node = nub_core::node::discovery::discover_node(cwd)
-        .unwrap_or_else(|_| nub_core::node::discovery::ResolvedNode::fallback());
+    let discovered = nub_core::node::discovery::discover_node(cwd);
+    // Published ABOVE the early returns: the engine keys its GVS virtual-store
+    // paths — and validates `engines.node` — off this version, so it has to name
+    // the Node dependency build scripts actually compile against. That stays the
+    // project's pin under `--node` / NODE_COMPAT, which disable augmentation but
+    // not version provisioning. Otherwise aube probes the ambient PATH node and
+    // two majors silently share one store entry. Deliberately left unset when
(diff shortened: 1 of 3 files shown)


colinhacks:

PR #531, opened 2026-07-22
fix(pm): strip time: from the catch-up integrity-rewrite lockfile

### LINKED ISSUE #520: pm: catch-up integrity rewrite writes a time: block into non-time-based lockfiles
`install/mod.rs` (~L2707) builds `lockfile_graph_for_integrity_rewrite` from a `graph.clone()` taken *after* the write block, so it still carries `times`. On that path a lockfile written outside `resolution-mode=time-based` gets a `time:` block, defeating the strip the main write path applies.

pnpm only persists `time:` under time-based resolution, so this is a pnpm-parity leak.

Pre-existing; surfaced during the aube v1.32 sync (#519), which made it more visible by moving the main write into an overlapped task.

colinhacks:
The catch-up integrity rewrite built its write graph from a clone of the resolved graph taken *after* the main write, so it still carried `graph.times` and re-emitted a top-level `time:` block. pnpm persists `time:` only under `resolution-mode=time-based`, so on that path a non-time-based lockfile got a `time:` block the main write had already stripped — a pnpm-parity leak.

## Fix

Route the rewrite through `lockfile_graph_for_write` with the same `persist_times` decision the main write uses, hoisted to one source of truth so the two writes cannot diverge on the `time:` axis. The `--lockfile-only` path already routed through the same helper; this was the last write site that bypassed it.

## Verified

- `cargo test -p aube-lockfile --lib` and `cargo test --workspace --lib` green; `cargo clippy --workspace --all-targets --all-features -- -D warnings` clean.
- New unit test pins both branches of the strip helper: a times-carrying graph serializes `time:`-free when not persisting and keeps it under time-based.

The full end-to-end trigger — a required (non-optional) package with a host-mismatched `os`/`cpu` plus a times-populating setting like `minimumReleaseAge` — needs a mock registry the hermetic harness lacks, so the guard sits at the write-graph level.

Closes #520

### DIFF
--- a/vendor/aube/crates/aube/src/commands/install/lockfile_dir.rs
+++ b/vendor/aube/crates/aube/src/commands/install/lockfile_dir.rs
@@ -116,3 +116,37 @@ pub(super) fn write_lockfile_dir_remapped(
     remapped.importers.insert(importer_key.to_string(), deps);
     aube_lockfile::write_lockfile_as(lockfile_dir, &remapped, manifest, kind)
 }
+
+#[cfg(test)]
+mod tests {
+    use super::lockfile_graph_for_write;
+    use std::borrow::Cow;
+
+    /// The resolver keeps publish times in `graph.times` whenever
+    /// minimumReleaseAge / trustPolicy / the defaultTrust floor is active,
+    /// but pnpm serializes a `time:` block only under time-based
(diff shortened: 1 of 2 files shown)


colinhacks:

PR #532, opened 2026-07-22
aube: split embedder node paths, add explicit version and provenance

colinhacks:
The embedder seam supplies Node from one directory: the seed derives `bin_dir/node`, which drives both `NODE` and `npm_node_execpath`. A wrapping host needs them distinct: shim on `PATH`, `NODE` at the shim, and `npm_node_execpath` at the real Node, which node-gyp reads for Node's prefix.

- A new `EmbedderRuntime` and `seed_embedder_runtime` keep the three paths independent, plus a supplied version (no probe) and explicit provenance.
- The runtime context gains `node_execpath`, falling back to `node_bin`.
- Provenance gains an `Embedder` variant; the seed previously reported every host as mise.
- The split reaches both spawn paths: `apply_child_env` (exec/dlx/create) and the dependency lifecycle-script env via `ScriptSettings.node_execpath` — the path where node-gyp actually runs, as a postinstall. Same `None`-falls-back-to-the-program contract at every layer.

Default-preserving: new fields default to none/off; `seed_embedder_node` delegates unchanged, and every current caller leaves `node_execpath` unset, so both env vars still name one binary.

### DIFF
--- a/vendor/aube/crates/aube-scripts/src/lib.rs
+++ b/vendor/aube/crates/aube-scripts/src/lib.rs
@@ -43,9 +43,16 @@ pub struct ScriptSettings {
     /// system one while project-local binaries still win. `None` when
     /// no runtime switching is active.
     pub node_bin_dir: Option<PathBuf>,
-    /// The resolved node executable, exported as `npm_node_execpath` /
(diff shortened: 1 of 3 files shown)


colinhacks:

PR #533, opened 2026-07-22
fix(runtime): resolve the dev preload from the source root; cover the augmentation seam

### LINKED ISSUE #528: test: lifecycle-script augmentation is untested end-to-end, and local runs can silently disable it
No test asserts a script spawned by a real `nub install` sees nub's augmentation. Both halves are unit-tested over hand-built structs and never joined (`lifecycle_overlay_carries_full_augmentation`, aube's `env_overlay_keys_are_applied`); the `configure_script_settings` → spawn seam is uncovered.

Local runs may also not exercise it: `find_preload` walks 5 parents for `runtime/preload.mjs`, and `scripts/rust-build.sh` sends non-diverged worktrees to `~/.cache/nub/shared-target`, which has no `runtime/` ancestor. CI uses in-repo `./target` and does augment. Found via #524.

colinhacks:
`find_preload` located `runtime/preload.mjs` only by walking up from the nub binary. A dev build routed to a target dir with no `runtime/` ancestor — the shared cross-worktree dir `scripts/rust-build.sh` uses on the fast path — found no preload, so `compute_augmentation_env` short-circuited and lifecycle scripts ran wholly un-augmented; CI (in-repo `./target`, which has `runtime/` as a sibling) augmented. A silent local/CI split, and the gap that let a lifecycle hang survive the whole suite during the v1.32 aube sync.

## Fix

Add a second, layout-agnostic route in `find_preload`: the source tree that compiled the binary (`CARGO_MANIFEST_DIR` → `<repo>/runtime`). It is gated to the non-embed build, so the shipped binary — built with `embed-runtime`, which resolves its preload from the embedded blob — never reaches it and is unchanged.

## Test

A new integration test (`crates/nub-cli/tests/pm_augment.rs`) runs a real offline install and asserts a root `postinstall` sees nub's preload in `NODE_OPTIONS` and the node-shim dir leading `PATH` — the augmentation seam (`compute_augmentation_env` → overlay → spawn) that was only unit-tested in halves. It fails loudly, not vacuously, when the build cannot locate its preload — the exact gap that let the hang through 2,672 aube + 443 nub tests.

## Verified

Measured both directions in a clean env (no ambient shim): a binary with no `runtime/` ancestor does not augment before the fix and augments after; a binary with a `runtime/` ancestor augments either way. The new test passes when augmentation is available and fails with "harness cannot exercise augmentation" when `find_preload` is forced to `None`. `clippy --all-targets --all-features -D warnings`, nub-core (364) and nub-cli `--bins` (321) unit tests all green.

Closes #528

### DIFF
--- a/crates/nub-cli/tests/pm_augment.rs
+++ b/crates/nub-cli/tests/pm_augment.rs
@@ -1,0 +1,150 @@
+//! The lifecycle-augmentation seam, end-to-end through the real binary.
+//!
+//! `nub install` must run a project's lifecycle scripts under nub's runtime
+//! augmentation — nub's preload in `NODE_OPTIONS` and the node-shim dir leading
+//! `PATH`, so a build script's `node`/`$NODE child.js` re-enters nub augmented and
+//! node-gyp compiles against the provisioned Node. Both halves of that seam
+//! (`augmentation_to_lifecycle_overlay` in pm_engine, aube's env-overlay
+//! application in aube-scripts) are unit-tested in isolation over hand-built
+//! structs; nothing joined `compute_augmentation_env` → the overlay → a real
+//! spawn. That uncovered join is what let a lifecycle hang survive 2,672 aube +
+//! 443 nub tests during the v1.32 sync (#528). This test closes it by observing a
+//! real root `postinstall`'s environment.
+//!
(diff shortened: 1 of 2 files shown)


colinhacks:

PR #534, opened 2026-07-23
fix(pm): approve-builds surfaces and approves local-source deps

### LINKED ISSUE #527: pm: approve-builds reports nothing to approve for local-source deps the install warns about
`nub install` emits `WARN_NUB_IGNORED_BUILD_SCRIPTS` for local-source dependencies (`file:./dep`, `file:./dep.tgz`) with build scripts, but `nub approve-builds --all` then reports "No ignored builds to approve" — so there is no way to approve them.

Reproduced on both the aube v1.32 sync branch and a binary built from main, so it is pre-existing and not sync-related.

Found while verifying #519.

colinhacks:
`nub install` warns `WARN_NUB_IGNORED_BUILD_SCRIPTS` for local-source dependencies (`file:./dep`, `file:*.tgz`, git) that ship a build script, but `nub approve-builds` / `nub ignored-builds` then reported nothing to approve — a dead-end with no way to enable those builds.

## Root cause

The warning and `approve-builds` derive the ignored set from different sources. The install warning enumerates the materialized tree (recorded in install state); `collect_ignored` instead classified each dep from its `(name, version)` index in the global store. A source-backed dep has no store index — its content is imported by dep_path, not a registry integrity — so the lookup always missed and the dep was dropped. Approving also needs the source key: a bare package name never authorizes a source-backed build.

## Fix

- `collect_ignored` now also surfaces source-backed deps from the install-recorded unreviewed set — the same set the warning enumerates — keyed by the source approval key (`dep@file:./dep`), so the two paths agree by construction.
- `approve-builds` writes that source key to `allowBuilds` (so the policy authorizes the build) while still targeting the follow-up rebuild by package name.
- `ignored-builds` and the picker show the source key, matching the warning and pnpm 11.

Mirrors pnpm 11, which lists `dep@file:./dep` in `ignored-builds` and approves by the source key.

## Verified

Fixture with a `file:` directory dep and a `file:*.tgz` dep, each with a `postinstall`: before, `ignored-builds` and `approve-builds --all` both reported nothing; after, both are listed, `approve-builds --all` writes `dep@file:./dep: true`, and the next install runs the postinstall (marker written) and clears the warning. Registry-dep approval is unchanged. Adds an e2e test covering the contract.

## Known limitation (follow-up)

approve-builds' in-invocation rebuild does not build a source-backed dep immediately: its virtual-store path folds in build state, so approving shifts the path and the scoped rebuild can't reach the pre-approval tree. The build runs on the next install. Registry deps are unaffected — this is a separate, deeper linker issue.

Closes #527

### DIFF
--- a/vendor/aube/crates/aube/src/commands/approve_builds.rs
+++ b/vendor/aube/crates/aube/src/commands/approve_builds.rs
@@ -92,14 +92,26 @@ fn run_project(cwd: &Path, all: bool, packages: Vec<String>) -> miette::Result<V
         return Ok(Vec::new());
     }
 
+    // A source-backed package (`file:`/tarball/git/…) is authorized by its
+    // source key, never its bare name, so the `allowBuilds` write must use
+    // `approval_key`. The follow-up rebuild instead matches deps by graph
+    // `name` (pnpm's `rebuild <name>` contract), so the two lists diverge
+    // for source-backed deps and must be threaded separately.
+    let entries = selected_entries(&ignored, &selected);
+    let approval_keys = dedupe(entries.iter().map(|e| e.approval_key.clone()).collect());
+    let display = entries
+        .iter()
+        .map(|e| e.display_spec())
+        .collect::<Vec<_>>()
+        .join(", ");
+
     // The picker only toggles names; scripts running is a separate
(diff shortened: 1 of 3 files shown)


colinhacks:

PR #537, opened 2026-07-23
fix(pm): key the side-effects cache on the Node engine, not just the platform

colinhacks:
The side-effects cache stored post-build package directories at `<name>@<version>/<os>-<arch>/<input-hash>`. The input hash fingerprints the package *before* its lifecycle scripts run, so nothing in that key described the Node the artifacts were built with. A native addon compiled under one `NODE_MODULE_VERSION` was a cache hit for an install pinned to a different Node major, and the whole post-build directory — `build/Release/*.node` included — was restored into it.

That is a broken install, not cache churn. The virtual store already segregates on `<os>-<arch>-node<major>` via `graph_hash::engine_name_default`, so the two caches disagreed about whether the engine mattered and artifacts leaked across a boundary the store path was busy keeping apart.

## Repro

A node-gyp addon registered with `NODE_MODULE` (not N-API, so ABI-sensitive), installed under Node 22.15.0 and then Node 26.5.0 in one project with `allowBuilds` on and an isolated store:

```
$ echo 22.15.0 > .nvmrc && nub install
side-effects-cache: saved    .../abi-probe@1.0.0/macos-aarch64/e8c24316…

$ rm -rf node_modules && echo 26.5.0 > .nvmrc && nub install
side-effects-cache: restored .../abi-probe@1.0.0/macos-aarch64/e8c24316…
```

Same path, one compile. The Node 22 build, the cache entry, and the Node 26 store copy of `abi_probe.node` all share sha256 `d925f2c284788219…`. Loading it under Node 26:

```
Error: The module '.../abi-probe/build/Release/abi_probe.node'
was compiled against a different Node.js version using
NODE_MODULE_VERSION 127. This version of Node.js requires
NODE_MODULE_VERSION 147.
```

After the change the two arms land at `macos-arm64-node22` and `macos-arm64-node26`, compile separately, and both load. Re-installing under Node 22 with the store entry wiped still logs `side-effects-cache: restored` and runs node-gyp zero times, so the cache is segregated rather than disabled.

## Change

Entries key on `engine_name_default`, the same fingerprint the virtual store uses, so both caches segregate on identical axes. The cache root and the resolved Node version travel together as `SideEffectsCacheLocation`, so no call site can name one without the other — `install` threads the version it already resolved for the engines check, `rebuild` resolves its own. An unresolved version falls back to the platform alone rather than naming an engine nobody observed.

Factoring `platform_name()` out of `engine_name_default` gives that fallback one shared spelling of `<os>-<arch>`. The output of `engine_name_default` is byte-identical, so virtual-store paths do not move.

## Fork discipline

This is not default-preserving: standalone aube's cache paths change too. It lands under the latent-bug-fix clause. Restoring an ABI-mismatched native build is a defect for standalone aube on exactly the same terms, and there is no posture under which the old key is correct, so the fix is unconditional rather than embedder-gated.

## Invalidation

Existing entries are orphaned by the path change, so affected native modules rebuild once. That is the intended cost — the old entries are ABI-ambiguous and must not be trusted. Nothing else invalidates: lockfiles are untouched, and the unchanged `engine_name_default` output means the virtual store does not re-key.

Refs #530

## Notes

The side-effects path's arch spelling changes from Rust's `aarch64` to Node's `arm64`, since the segment now comes from `engine_name_default` rather than raw `std::env::consts::ARCH`. Existing entries orphan on the engine segment regardless, so this costs nothing extra, and it is what makes the two caches segregate on identical axes.

Under standalone aube, `rebuild` cannot reach `runtime::current()` (it never enters a `runtime::scope`), so with runtime switching plus a version pin it resolves a different Node than `install` and primes an entry `install` will not read. Nub is unaffected: both verbs route through `engine_session`, which publishes the discovered version. The prior behavior was worse, since both wrote the same platform-only path, so `rebuild` could prime a wrong-ABI entry that `install` would then restore; the mismatch now degrades to lost priming rather than a bad restore.

### DIFF
--- a/vendor/aube/crates/aube-lockfile/src/graph_hash.rs
+++ b/vendor/aube/crates/aube-lockfile/src/graph_hash.rs
@@ -63,14 +63,24 @@ pub struct EngineName(pub String);
 /// store directories look familiar next to `process.arch` output.
 /// Libc detection is a known gap (TODO: musl vs glibc).
 pub fn engine_name_default(node_version: &str) -> EngineName {
-    let os = std::env::consts::OS;
-    let arch = node_arch(std::env::consts::ARCH);
     let major = node_version
         .trim_start_matches('v')
         .split('.')
         .next()
         .unwrap_or("");
-    EngineName(format!("{os}-{arch}-node{major}"))
+    EngineName(format!("{}-node{major}", platform_name()))
+}
+
+/// The `<os>-<arch>` prefix of [`engine_name_default`] on its own, for
+/// callers that key build artifacts but could not resolve a Node
+/// version: they degrade to the platform axis instead of naming an
+/// engine nobody observed.
+pub fn platform_name() -> String {
+    format!(
+        "{}-{}",
+        std::env::consts::OS,
+        node_arch(std::env::consts::ARCH)
+    )
 }
 
 /// Map Rust `std::env::consts::ARCH` values to Node's `process.arch`
--- a/vendor/aube/crates/aube-settings/settings.toml
+++ b/vendor/aube/crates/aube-settings/settings.toml
@@ -2195,7 +2195,10 @@ default = "true"
 docs = """
 When an allowlisted dependency runs lifecycle scripts, aube snapshots
 the post-build package directory under the cache dir keyed by
-`(name, version, input hash)`. Future installs with the same inputs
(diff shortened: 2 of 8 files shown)


colinhacks:

PR #540, opened 2026-07-23
skills(aube-bump): record the patch and build-approval fork delta

colinhacks:
Four fixes landed in `vendor/aube` after the v1.32.0 bump that stock aube does not have: alias patch inheritance (#519), unused-patch-key failure (#519), and source-backed build-approval surfacing (#534).

The next bump has to keep them, so they belong in the skill's delta inventory. The build-approval one changes standalone-aube behavior, so it will show as fork delta rather than converge — noted inline.

Docs only.


colinhacks:

PR #542, opened 2026-07-23
fix(pm): fold the Node engine into the install freshness and delta gates

### LINKED ISSUE #539: install delta filter is engine-blind: a Node major switch skips the lifecycle phase
After a Node major switch, a repeat `nub install` can leave a native addon compiled against the previous Node in place. The install-delta filter skips the lifecycle phase for the package entirely, so the side-effects cache is never consulted and the addon is never rebuilt.

Requires the shared virtual store to be off (`enableGlobalVirtualStore=false`) — the default under CI, for `dlx`, and for compat-disabled packages. With the shared store on, the virtual-store directory is engine-keyed, so a Node switch yields a fresh directory and the rebuild happens correctly.

## Mechanism

`run_dep_lifecycle_scripts` filters packages by `selected_dep_paths` (`lifecycle.rs:440-444`) before constructing a cache entry, so `restore_if_available` never runs. That set comes from `lifecycle_delta_filter` (`finalize.rs:77-121`, wired at `:232-244`), which is enabled by default on any install with prior state. Its three inputs are all engine-blind:

- `dep_build_policy_hash` — build-policy plus floor fingerprint (`finalize.rs:62-75`); `DefaultTrustFloor::fingerprint` hashes four booleans (`default_trust.rs:149-169`).
- leaf and subtree hashes — package content only (`delta.rs:218-226`).
- the install-state settings fingerprint — `hash_settings` enumerates node_linker, hoist family, modules_dir, import method, GVS, lockfile, dedupe, symlink, and disk_materialize, but no Node version (`state.rs:1348-1460`).

So a switch from Node 26 to Node 22 with an unchanged lockfile selects nothing, and the Node-26 `.node` survives in `node_modules`.

## Reproducing

Install a `NODE_MODULE`-registered (non-N-API) node-gyp addon under one Node major with the shared virtual store off, switch to another major, and install again without `--force`. `require()` then fails with `ERR_DLOPEN_FAILED` and a `NODE_MODULE_VERSION` mismatch. Passing `--force`, or changing the dependency build policy, widens the filter to a full scan and the rebuild happens correctly.

## Scope

Not addressed by #530, #537, or #538. Those made the virtual-store path, the side-effects cache path, and the cache marker engine-aware; this is the remaining layer where the engine is not an input, and it sits upstream of all three — the fixed code is never reached.

The likely fix is folding the engine into `hash_settings` or `dep_build_policy_hash`. That forces a relink on every Node switch, so it warrants its own blast-radius pass rather than riding along with the cache work.

Refs #530, #537, #538

colinhacks:
## The bug

A Node major switch could leave a native addon compiled against the previous ABI in place, with the install reporting success.

Two gates sit upstream of the side-effects cache, and both were engine-blind:

- `check_needs_install` decides the tree is current from `hash_settings`, which enumerates `node_linker`, the hoist family, `modules_dir`, import method and the rest, but carried no Node input. A plain repeat install after a switch short-circuited before any lifecycle work.
- `lifecycle_delta_filter` then narrows the phase to packages whose `dep_path` changed, gated on `dep_build_policy_hash` — also engine-free. So even a full install skipped the unchanged addon.

The cache layers keyed by engine in #537 and #538 sit downstream of both, so on a plain switch they were never reached.

## Reproduction

One project, one native addon (`NODE_MODULE`-registered, so ABI-sensitive), default layout. Nothing changes between the two installs except the Node on `PATH` — no manifest edit, no `--force`, no policy change.

```
STEP 1: node v22.15.0 (NODE_MODULE_VERSION=127)
  side-effects-cache:   saved
  ran N scripts:        allowBuilds: ran 1 dep lifecycle
  artifact sha256:      a3731d6d406e3615
  require('abi-probe'): LOADED OK: pong

STEP 2: node v26.5.0 (NODE_MODULE_VERSION=147)
  needs-install reason: (none)              # short-circuited
  delta verdict:        (none)              # the delta filter never ran
  side-effects-cache:   (none)              # never consulted
  ran N scripts:        (none)
  artifact sha256:      a3731d6d406e3615    # still the ABI-127 build
  require('abi-probe'):
      Error: The module '.../abi_probe.node' was compiled against a different
      Node.js version using NODE_MODULE_VERSION 127. This version of Node.js
      requires NODE_MODULE_VERSION 147.
```

`install rc=0` throughout. That the cache is never consulted in step 2 is what shows the freshness check short-circuits ahead of the delta filter: fixing only the delta filter leaves the behavior unchanged.

## After the fix

The same fixture, reduced to the gates themselves: a dependency whose `postinstall` stamps `process.versions.modules`, so the recorded major shows directly whether the build re-ran. Nothing changes between arms but the Node on `PATH`.

```
                          base (main)          with this change
STEP 1  node v22.15.0
  ran N scripts:          ran 1 dep lifecycle  ran 1 dep lifecycle
  built by MODULE_VERSION 127                  127

STEP 2  node v26.5.0
  settings/Node changed:  (silent)             install settings or the active
                                               Node version have changed
  ran N scripts:          (none)               ran 1 dep lifecycle
  built by MODULE_VERSION 127  <- stale        147  <- rebuilt
```

## The fix

Fold the engine name into both gates. `hash_settings` resolves it through the existing override-aware ctx, so an `.npmrc` `node-version` is still honored, and `dep_build_policy_hash` takes it alongside the policy and floor fingerprints.

Granularity is the engine name (`<os>-<arch>-node<major>`), not the raw version, so a `22.15.0` to `22.15.1` bump does not invalidate a warm tree. The same string now identifies the engine in all four layers: the virtual-store path, the side-effects cache path, its marker, and these two gates.

`dep-build-policy-v1` becomes `-v2`, so the first install after upgrading runs one full eligible-build scan.

The settings-change reason now reads "install settings or the active Node version have changed" — a Node switch reaching that branch would otherwise report an `.npmrc` edit that never happened.

## Cost

A Node major switch now invalidates the warm install state, so the next install re-links and revisits build-allowed packages instead of reporting the tree current. That is the point: the prior behavior reported success over a broken addon. Within a major, nothing changes.

## Fork discipline

Not default-preserving — the two hashes move for standalone aube too, which is the latent-bug-fix clause: an install that leaves a wrong-ABI addon in place and exits 0 is broken on any host, not only under the embedder.

Closes #539
Refs #530, #537, #538

## Scope of the settings-hash half

The freshness hash folds the engine only when the embedder owns Node provisioning. That gate is load-bearing rather than conservative: `hash_settings` is computed once by the freshness check, before `runtime::ensure` resolves the runtime, and once by the state write, after it. Under runtime switching those two points see different values of `runtime::current()`, so a resolver-derived engine would make the two hashes disagree permanently — every install would report the settings changed and rewrite a value the next check could never match. An embedder that provisions Node publishes its resolved version before any install code runs, so both sites read one value.

Standalone aube therefore keeps this hash byte-for-byte, and its delta gate still tightens: `dep_build_policy_hash` is computed after `ensure`, where the resolver has settled. What stays open there is a pin change that alters the major without touching any hashed setting; closing it needs the runtime resolved before the fast path, which is a larger reordering than this fix.

Not covered by a test: that the fold stays absent under runtime switching. Asserting it means mutating the process-global engine context, which would race the engine-context tests in the same binary — a flake this repo treats as a defect. The gate is structural instead.

### DIFF
--- a/vendor/aube/crates/aube/src/commands/install/finalize.rs
+++ b/vendor/aube/crates/aube/src/commands/install/finalize.rs
@@ -59,18 +59,31 @@ pub(super) struct FinalizePhaseInput<'a> {
     pub(super) phase_timings: &'a mut InstallPhaseTimings,
 }
 
+/// Gates whether the next install may narrow its lifecycle phase to changed
+/// packages. The engine belongs in it because build scripts compile against a
+/// specific Node ABI: without it, switching Node majors leaves the prior
+/// install's state looking current, the delta filter selects nothing, and a
+/// native addon built for the old major survives untouched — the cache layers
+/// downstream never get consulted. Engine-name granularity (major, os, arch)
+/// rather than the raw version keeps a patch bump from forcing a full scan.
 fn dep_build_policy_hash(
     build_policy: &aube_scripts::BuildPolicy,
     default_trust_floor: &super::default_trust::DefaultTrustFloor,
+    node_version: Option<&str>,
 ) -> String {
     let policy = build_policy.fingerprint();
     let floor = default_trust_floor.fingerprint();
+    let engine = node_version.map_or_else(aube_lockfile::graph_hash::platform_name, |v| {
+        aube_lockfile::graph_hash::engine_name_default(v).0
+    });
     let mut hasher = blake3::Hasher::new();
-    hasher.update(b"dep-build-policy-v1");
+    hasher.update(b"dep-build-policy-v2");
     hasher.update(&(policy.len() as u64).to_le_bytes());
     hasher.update(policy.as_bytes());
     hasher.update(&(floor.len() as u64).to_le_bytes());
     hasher.update(floor.as_bytes());
+    hasher.update(&(engine.len() as u64).to_le_bytes());
+    hasher.update(engine.as_bytes());
     hasher.finalize().to_hex().to_string()
 }
 
@@ -228,7 +241,8 @@ pub(super) async fn run_finalize_phase(input: FinalizePhaseInput<'_>) -> miette:
     }
 
     let filtered_install = !workspace_filter_empty || dep_selection.is_filtered();
-    let dep_build_policy_hash = dep_build_policy_hash(build_policy, default_trust_floor);
+    let dep_build_policy_hash =
+        dep_build_policy_hash(build_policy, default_trust_floor, node_version);
     let lifecycle_delta_filter = if ignore_scripts {
         None
     } else {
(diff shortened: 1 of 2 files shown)


colinhacks:

PR #543, opened 2026-07-24
pm: report linked files through the embedder events reporter, and roll back legacy lockfiles

### LINKED ISSUE #521: pm: linked-file count is not reported through aube's embedder Events reporter
nub's `files_linked` counter drives the live `N files` count during linking. aube v1.32 added an embedder events mode (`Mode::Events`, jdx/aube#1036) whose `InstallProgressSnapshot` has no field for it, so an embedder consuming events gets resolve/fetch progress but no linking progress.

Not reachable from nub's own CLI today (nub renders its own TTY line). Fixing it needs a new field on aube's stable embedding API, i.e. an upstream request.

Found during the aube v1.32 sync (#519).

### LINKED ISSUE #522: pm: add_to_project cancellation rollback misses nub's legacy lockfile basenames
aube v1.32's new embedder entrypoint `add_to_project` restores only the primary `lockfile_path` on cancellation. nub's `--no-save` path additionally snapshots and restores `lockfile_legacy_basenames` via `snapshot_legacy_lockfiles`.

Under nub identity, a cancelled embedded add whose install migrated the legacy name would leave the legacy file deleted.

Latent: `add_to_project` has zero call sites in this repo. Found during the aube v1.32 sync (#519).

colinhacks:
Two fixes to aube's embedding API — the surface nub vendors but does not itself consume. nub calls `aube::commands::<verb>::run()` directly and never selects Events mode, so neither bug is reachable from nub's CLI today; both are fixed so the vendored tree is correct if that surface is ever wired.

## Linked-file count reaches the events reporter

`InstallProgressSnapshot` gains `files_linked`. The counter is the linker's own `Arc`, reached through `link_progress_counter()`, so a host consuming events sees the same live linking signal nub's TTY line already renders.

`link.rs` previously handed that counter to the linker only when the embedder opted into the TTY progress UX. That would have pinned the new field at zero for exactly the host Events mode exists to serve — a field that always reports `0` is worse than no field — so it now also passes it under Events. Standalone aube selects neither, leaving its link pass byte-identical.

**Known limitation, stated rather than hidden:** Events mode has no heartbeat, so nothing emits between the linking phase start and `finish()`. A host reads `0` on the Linking snapshot and the true total on Complete. The value is loaded per snapshot rather than latched, so it starts ticking the moment any mid-link emission exists; adding an Events heartbeat thread is a larger change and deliberately out of scope here.

## Cancellation restores legacy lockfiles

`add_to_project`'s rollback restored only the current-name lockfile. A write during the add migrates a legacy basename — writes the current name, deletes the legacy file — so a cancelled embedded add could leave the host's checked-in legacy lockfile deleted.

It now snapshots and restores the same set `--no-save` does, through the existing `no_save::snapshot_legacy_lockfiles`, so all three rollback paths agree rather than drifting apart. Every path is attempted before an error surfaces, so one unwritable file cannot strand the rest half-restored. Standalone aube's profile carries no legacy basenames, making the loop inert there.

## Verification

- aube: `cargo clippy --workspace --all-targets --all-features -- -D warnings` clean; `cargo test --workspace --lib` 2,612 passed / 0 failed; the new `embed_legacy_lockfile` test passes.
- nub: `check -p nub-cli --all-targets`, `clippy --all-targets --all-features -- -D warnings`, and `fmt --check` all clean.
- The `files_linked` test bumps the counter through `link_progress_counter()` — the linker's only handle — so a duplicate atomic in `EventState` would read `0` and fail it.
- Tests are unit-level by necessity: nub's CLI never selects Events mode and `add_to_project` has no nub call sites, so an end-to-end through nub is unreachable for either path.

Closes #521
Closes #522

### DIFF
--- a/vendor/aube/crates/aube/src/commands/add/mod.rs
+++ b/vendor/aube/crates/aube/src/commands/add/mod.rs
@@ -80,6 +80,13 @@ pub async fn add_to_project(
         .wrap_err("failed to snapshot package.json before embedded add")?;
     let lockfile_path = no_save::lockfile_path_for_project(lock.project_dir())?;
     let original_lockfile = no_save::snapshot_lockfile(&lockfile_path)?;
+    // The rollback set must match what `--no-save` snapshots (`no_save::Snapshot`
+    // and the filtered add's root snapshot): the current-name lockfile *plus*
+    // every pre-rename legacy basename. A write during the add MIGRATES the
+    // legacy name (writes the current name, deletes the legacy file), so
+    // restoring only `lockfile_path` would leave the host's checked-in legacy
+    // lockfile deleted. Empty — hence inert — for standalone aube.
+    let original_legacy = no_save::snapshot_legacy_lockfiles(&lockfile_path)?;
     let mutation_control = options.control.clone();
     let mutation_result = install::control::scope(options.control.clone(), async {
         tokio::select! {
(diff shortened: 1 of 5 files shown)


colinhacks:

PR #545, opened 2026-07-24
test(pm): pin that approve-builds runs a local-source dep's build immediately

colinhacks:
PR #534 shipped a "known limitation" in a test comment: that a source-backed (`file:`) dependency, unlike a registry dep, could not build during the `approve-builds` invocation and only built on the next install. It is false. A `file:` dep builds during `approve-builds`, exactly like a registry dep.

## Why it's stable

`LocalSource::dep_path` derives a directory dep's virtual-store cell from the normalized path alone (`source.rs`) — no build state enters it — and `is_globally_shareable` covers only git and remote tarballs, so a `file:` dep never enters the global store whose subdir folds in the build-state graph hash. Approval therefore cannot move the cell, and the scoped rebuild inside `approve-builds` reaches it.

## Verified against the built binary

Across three fresh fixtures, each with a uniquely-named `file:` dependency carrying a `postinstall`: the marker is absent after `install` and present immediately after `approve-builds`.

A unique name per fixture is load-bearing here — `file:` deps key the global virtual store by `name@version`, so a fixture reusing a name from an earlier run links that run's already-built cell and the build is (correctly) skipped, which reads as a false confirmation of the limitation.

## Change

The `#527` e2e test asserted the marker only after the follow-up install, so nothing pinned the real behavior and the false claim sat unchallenged in a comment. It now asserts the build ran during `approve-builds`, keeps the survives-a-reinstall assertion, and records why the path is stable. Test-only; `cargo test -p aube --test e2e` green.

Refs #534

### DIFF
--- a/vendor/aube/crates/aube/tests/e2e.rs
+++ b/vendor/aube/crates/aube/tests/e2e.rs
@@ -233,7 +233,8 @@ fn approve_builds_surfaces_and_runs_a_local_source_dep() {
     // set, never in the store the enumeration read. This pins the whole
     // contract: the dep is listed, approving it writes the *source* key
     // (a bare name never authorizes a source-backed build), and the
-    // build then runs on reinstall while the warning clears.
+    // build runs during that same `approve-builds` call, then survives
+    // the next install while the warning clears.
     let _guard = e2e_lock();
     let sbx = Sandbox::new();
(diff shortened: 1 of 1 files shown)


colinhacks:

PR #541, opened 2026-07-23
Trim the default trust build warning

colinhacks:
## Summary
- shorten the human-facing `defaultTrust` build disclosure
- keep the stable warning code while omitting duplicate count and package fields
- update the captured output and warning-channel regression check

## Validation
- `cargo clippy --all-targets --all-features -- -D warnings`
- `cargo fmt -p nub-cli -- --check`
- `cargo test -p nub-cli --bin nub default_trust_disclosure_does_not_repeat_count_and_packages`
- real `msgpackr-extract@3.0.4` install emitted the compact warning and completed successfully

`cargo fmt --check` also reports existing formatting drift under `vendor/aube/**` on the base commit.

PR #546, opened 2026-07-24
refactor(pm): document the default-trust log-collapse invariant

colinhacks:
Follow-up polish on the `format_line` collapse from #541. No behavior change.

- Document why the collapse is correct: `format_line` must run before `present::rewrite` so it matches the pre-rewrite `WARN_AUBE_*` code — reordering silently breaks it. Undocumented until now.
- Drop the redundant `message.clone()` via `Cow`.
- Rename the test and cover two untested branches: other warnings keep their fields; the disclosure survives an upstream reword.

Local test run relied on CI — an environment limit was killing local builds.


colinhacks:

PR #547, opened 2026-07-24
skills(ad-hoc-test): warn that reused fixture package identities link stale built cells

colinhacks:
nub's global virtual store persists built package cells across runs, so an install/build-approval fixture reusing a dependency's `name@version` can link a prior run's already-built cell instead of building the current one — a silent false confirmation.

Verified while checking an approve-builds claim: a `file:./dirdep` dep named `dirdep@1.0.0` linked a cell carrying a marker the current fixture never wrote. Guidance: unique package name per run, or wipe the store, and confirm `node_modules/<dep>/` is yours.

Skill docs only.


colinhacks:

PR #548, opened 2026-07-24
fix(pm): make every remaining install cache layer engine-aware

colinhacks:
Three install cache layers still ignored the Node the artifacts are compiled by, so a native addon built under one Node major could be served to, or shared with, an install running a different major. Each is confirmed by a differential fixture (`better-sqlite3`, two Node majors) and pinned by a unit test. This continues the engine-keying work in #530, #537, #538, and #542 — the layers those closed all sit downstream of these.

## 1. GVS fold was blind to `defaultTrust`-floored builds

The virtual-store engine fingerprint folded in only for an explicitly allowed package, but the lifecycle phase also builds packages the `defaultTrust` floor clears. nub enables that floor by default, and its curated list is exactly the native-ABI packages (`better-sqlite3`, `sharp`, `bcrypt`, `canvas`, `node-pty`, …). So a floor-built addon got an engine-agnostic, machine-global, cross-project store path — two majors overwriting each other's `build/Release/*.node` there. Worse than the earlier layers: the victim install reports "Already up to date", so it is not repairable by a plain `nub install`, only `--force`.

```
# before: floor-built better-sqlite3 shares one bucket across majors
floor node22.15.0 -> better-sqlite3@12.11.1-4c55e1943396abea
floor node26.5.0  -> better-sqlite3@12.11.1-4c55e1943396abea

# after: segregated, and identical to an explicitly-allowed build
floor node22.15.0 -> better-sqlite3@12.11.1-aeb6e8ed40c40bc4
floor node26.5.0  -> better-sqlite3@12.11.1-527b3af3c84991cb
```

`build_may_key_engine` over-approximates the lifecycle gate on purpose: the floor's dynamic checks (advisory vetting, the cooling window) only ever narrow what builds, and are unknown at prewarm time where these hashes are first computed — and the link phase reuses the prewarm's hashes wholesale, so the two sites must agree. Threaded via one `default_trust_enabled` bool. Gated on that flag, so standalone aube — floor off by default — is byte-for-byte unchanged.

## 2. The validation-only `node-version` override keyed the ABI caches

`.npmrc node-version` is documented as validation-only (pnpm semantics — it never switches the runtime scripts run on), yet it was the top tier of `effective_node_version`, which feeds every ABI key. So it named a Node that never compiled the artifacts.

```
# real node 26, override node-version=22.15.0
before -> better-sqlite3@12.11.1-aeb6e8ed40c40bc4   # node22's bucket — wrong
after  -> better-sqlite3@12.11.1-527b3af3c84991cb   # real node26's bucket
```

`build_node_version` drops the override tier and feeds the six ABI-keying sites; `effective_node_version` keeps the override for the `engines.node` check and `doctor`, where it is meant to win. A justified latent-bug fix, not default-preserving: keying build artifacts off the override contradicts aube's own documented semantics on any host.

## 3. Workspace-member cwd flipped the engine key against root-anchored state

Install anchors its state and store at the workspace root, but nub published the engine from `discover_node(cwd)` — the process cwd. A member with its own `.nvmrc` then keyed a different major than the root, thrashing the warm path (a #542 regression on the freshness half).

```
# workspace root .nvmrc=22, member .nvmrc=26, install run from the member
before -> better-sqlite3@12.11.1-527b3af3c84991cb   # member's node26 — thrash
after  -> better-sqlite3@12.11.1-aeb6e8ed40c40bc4   # root's node22 — stable
```

Discovery is anchored at `detect_project(cwd).workspace_root` — the same directory aube keys state at, resolved by the same walk-up rule. nub-side only; standalone aube unaffected. A member install now builds the one shared tree against the root's Node, which is also the more correct behavior.

## Migration

Each fix moves a store path for the affected projects, so the next install re-materializes and rebuilds those subtrees once — no re-download (the CAS is integrity-keyed). Correct: the old entries were ABI-ambiguous.

Refs #530, #537, #538, #542

### DIFF
--- a/crates/nub-cli/src/pm_engine/mod.rs
+++ b/crates/nub-cli/src/pm_engine/mod.rs
@@ -1390,10 +1390,21 @@ fn augmentation_to_lifecycle_overlay(
 /// re-entrant / broken install); the resolved Node *version* is published to the
 /// engine either way. Called once per command from [`engine_session`].
 fn apply_lifecycle_augmentation(cwd: &Path) {
+    // Anchor Node discovery at the workspace root — the directory aube keys its
+    // install state and virtual store at (`dirs::workspace_or_project_root`). An
+    // install from a member materializes the ONE shared tree for the whole
+    // workspace, so the Node its build scripts compile against — and the engine
+    // that keys the ABI caches — must be the root's pin, not whichever member the
+    // shell sits in. Discovering from the raw cwd instead lets a member's own
+    // `.nvmrc` flip the engine key against a root-anchored state file, thrashing
+    // the warm path. `detect_project` walks up by the same rule aube's root does.
+    let anchor = nub_core::workspace::detect::detect_project(cwd)
+        .map(|p| p.workspace_root.unwrap_or(p.root))
+        .unwrap_or_else(|| cwd.to_path_buf());
     // The project's Node — pin-aware (`.nvmrc`/`.node-version`/`engines`), NOT
     // the ambient PATH node. This resolved version drives flag injection and its
     // path pins npm_node_execpath. Mirrors build_script_command's discovery.
-    let discovered = nub_core::node::discovery::discover_node(cwd);
+    let discovered = nub_core::node::discovery::discover_node(&anchor);
     // Published ABOVE the early returns: the engine keys its GVS virtual-store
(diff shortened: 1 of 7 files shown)


colinhacks:

PR #553, opened 2026-07-24
perf(pm): restore cold_path() hints in the vendored aube hot loops

colinhacks:
Restores the `cold_path()` hints #442 reverted. Upstream aube pins MSRV 1.91 for mise's distro packaging, but nub vendors aube and builds on stable, so that floor doesn't bind us.

- 22 `core::hint::cold_path()` hints on the rare arms of the hot install loops (linker fallbacks, resolver cache-misses, subset-parser bails, tarball rejects). Inert.
- `vendor/aube/Cargo.toml` MSRV → 1.95, documented as a deliberate fork delta (kept on every bump; noted in the `aube-bump` skill).
- nub root MSRV → 1.95 (nub-cli depends on aube; nub-native already needed it). CI MSRV Check runs root + native on 1.95.0; nub-phantom (aube-free) keeps 1.93.0.

Builds clean on stable locally. Refs #442.


colinhacks:

PR #554, opened 2026-07-24
test(pm): pin engine-key invariants #548 relies on

colinhacks:
Two regression tests pinning invariants [#548](https://github.com/nubjs/nub/issues/548) relies on but left uncovered. Tests only — no behavior change.

## Prewarm ≡ link engine-fold agreement

The GVS prewarm and the link phase hash the same graph independently, and link reuses the prewarm's hashes on a key-set match without comparing values. If the two `allow` closures diverge, link silently serves prewarm's hashes and the engine keying is wrong with no failure.

`engine_fold_agreement_tests` (`vendor/aube/crates/aube/src/commands/install/mod.rs`) pins it two ways:
- a source-level check that both phases gate the fold on identical `build_may_key_engine` arguments — fails if one closure is edited and not the other;
- a hash-level check that both hashers produce byte-identical `node_hash` for a `defaultTrust`-floor-built native package and its dependent (the case #548 fixed), with an engine-off control so the equality can't pass vacuously.

## Workspace-root Node anchor

`apply_lifecycle_augmentation` anchors Node discovery at the workspace root, so a member install keys the ABI caches off the root's pin. The anchor is factored into `lifecycle_node_anchor` (behavior-preserving), and `a_member_install_resolves_the_workspace_roots_node_pin` (`crates/nub-cli/src/pm_engine/mod.rs`) asserts a workspace with a root pin and a different member pin resolves the root's pin from a member cwd, with the raw-cwd member pin as the control.

Both tests were verified to fail when the invariant is broken (one `allow` closure changed; the anchor pointed at the raw cwd) and pass on revert.

Refs #548

### DIFF
--- a/crates/nub-cli/src/pm_engine/mod.rs
+++ b/crates/nub-cli/src/pm_engine/mod.rs
@@ -1382,6 +1382,23 @@ fn augmentation_to_lifecycle_overlay(
     (overlay, prepends)
 }
 
+/// The directory lifecycle Node discovery is anchored at: the workspace root
+/// when `cwd` sits in a member, else the project root, else `cwd` itself.
+///
+/// aube keys its install state and virtual store at the workspace root
+/// (`dirs::workspace_or_project_root`), and an install from a member
+/// materializes the ONE shared tree for the whole workspace — so the Node its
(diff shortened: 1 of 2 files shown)


colinhacks:

PR #555, opened 2026-07-24
fix(run): repair a nub-installed tree built for a different Node major

colinhacks:
Native addons compile against one Node ABI (`NODE_MODULE_VERSION`). When a project's Node major changes between install and run — a bumped `.nvmrc`/`engines.node`, or a switched Node — `node_modules` still holds addons built for the old ABI. `nub run` didn't notice: the pre-run freshness walk in `verify_deps` compares declared deps against what resolves on disk, and a wrong-ABI addon is present and version-satisfying, so the walk reports the tree fresh and the run crashes with a raw error from inside a dependency.

## Before / after

Node 22 → 26 switch, a dependency with a native addon (captured from the dev binary):

```
$ nub run probe            # under Node 26; tree was installed under Node 22
Error: The module '…/probe.node' was compiled against a different Node.js
version using NODE_MODULE_VERSION 127. This version of Node.js requires
NODE_MODULE_VERSION 147. …  code: 'ERR_DLOPEN_FAILED'
```

```
$ nub run probe
nub: dependencies were installed for Node 22, this run uses Node 26. Reinstalling — native addons are built per Node major.
nub 0.5.0 · ✓ installed 1 package in 1.9s
addon loaded: ok
```

A second run under Node 26 is clean (stamp matches, no reinstall); `NUB_VERIFY_DEPS_BEFORE_RUN=off` suppresses the repair and the raw crash returns.

## Detection

Each successful nub install writes `node_modules/.nub-engine` = `<os>-<arch>-node<major>` — the engine string the store and side-effects cache already key builds on. The run path reads that stamp and compares it to the engine it is about to use.

Hot-path cost: the stamp read comes first and the current engine is resolved lazily, so a tree nub did not install pays a single failed file read — no Node resolution. A nub-installed tree resolves the current engine spawn-free from the discovery cache, falling back only to the resolution the launch is about to perform anyway. There is no `node_modules` walk.

## Repair

On a mismatch, a plain `nub install` — the command the warning path already tells users to run. The engine folds into the install's own freshness/delta hash, so a present, satisfying lockfile is not re-resolved (the no-churn write leaves it untouched) while the native dep is relinked from its per-engine store build, or rebuilt when this engine was never built before. It prints what it is doing, and `Policy::Off` skips it along with the rest of the check.

## Posture

`verify_deps` deliberately maps `install`→`warn`, with the rationale that nub won't reshape a tree another PM installed. That is scoped to *other* PMs' trees. Only a nub install writes the stamp, so a mismatch here is nub redoing nub's own work, not reshaping npm/pnpm/yarn's — a narrow carve-out, not a reversal. Ordinary staleness and foreign trees keep today's warn behavior.

Refs #548

### DIFF
--- a/crates/nub-cli/src/install_engine.rs
+++ b/crates/nub-cli/src/install_engine.rs
@@ -1,0 +1,152 @@
+//! Which Node engine a nub install built `node_modules` for (issue #548).
+//!
+//! A native addon is compiled against exactly one `NODE_MODULE_VERSION`, so a
+//! tree whose lifecycle scripts ran under Node 22 dies with
+//! `ERR_DLOPEN_FAILED … NODE_MODULE_VERSION 127 … requires 147` the moment the
+//! project's Node moves to 26. The pre-run freshness walk cannot see that: the
+//! wrong-ABI package is present and its version satisfies the manifest, so the
+//! walk is engine-blind by construction. Nub therefore records, beside the tree
+//! it built, WHICH engine built it, and the run path compares that stamp
+//! against the engine it is about to run under (see
+//! [`verify_deps`](crate::verify_deps)).
+//!
+//! Only a successful nub install writes the stamp, which is what scopes the
+//! auto-repair: nub redoes nub's OWN work, and never reshapes a tree
+//! npm/pnpm/yarn owns. The recorded value is the engine's own artifact-keying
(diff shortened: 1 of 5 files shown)


colinhacks:

PR #551, opened 2026-07-24
ci: bun-style canary channel — build and publish every code push to main

colinhacks:
Every code push to `main` builds the 8-platform matrix with an uncommitted `<base>-canary.<yyyymmdd>.<run#>` stamp, publishes all ten packages under the npm `canary` dist-tag, and recreates the rolling `canary` GitHub prerelease. The publish lives in `release.yml` (npm trusted publishing binds one workflow file per package); tag-path jobs are gated to `refs/tags/`, so a main push cannot touch `latest`. Docs-only and release-bump pushes skip; glibc-floor + smoke gates still run; `releases/latest` never sees the prerelease. `make version` stamping moved to `scripts/set-version.mjs`.

https://claude.ai/code/session_01KbVuigZQd5uHVjYcusfFMo

PR #557, opened 2026-07-24
feat(upgrade): canary channel consumers — nub upgrade --canary/--stable, installer canary arg

colinhacks:
The consumer half of the canary channel #551 publishes — every install method can now reach the rolling canary and return to stable.

- `nub upgrade --canary` / `--stable`, mirroring bun: `--canary` opts in, a canary build's bare `nub upgrade` stays on canary, `--stable` (or an explicit `--version`) returns to stable, and an already-current canary short-circuits instead of re-downloading. Self-owned installs swap in place from `releases/download/canary/` (sha256-verified, same atomic swap as stable); npm installs ride the `canary` dist-tag; Homebrew stays stable-only and refuses with the installer one-liner.
- winget-owned binaries become their own advisory channel: `nub upgrade` prints `winget upgrade --id Nubjs.Nub` instead of the generic unknown-channel npm hint, and never spawns it — winget cannot replace the running exe and records the upgrade as done anyway (microsoft/winget-cli#5235); a self-swap would strand winget's version tracking (winget-cli discussion #3304). Dormant until nub is published to winget.
- `install.sh` / `install.ps1` accept `canary` as the version argument (the rolling tag needs no `releases/latest` resolution); synced to `site/public`.
- `stats/download-counts/report.mjs` skips the canary tag when computing deltas — snapshots don't carry the prerelease flag the #551 filters key on, so a tag guard is the only implementable belt-and-suspenders there.
- FAQ entry: installing and leaving the canary channel.

Tests: channel decision table, canary URL shapes, clap flag conflicts, winget path detection, and `file://`-fixture e2e for the rolling-tag download/verify/swap plus the already-current short-circuit.

Refs #551

### DIFF
--- a/crates/nub-cli/src/cli.rs
+++ b/crates/nub-cli/src/cli.rs
@@ -882,8 +882,17 @@ pub enum Command {
     /// Upgrade Nub to the latest version.
     Upgrade {
         /// Target version (default: latest).
-        #[arg(long)]
+        #[arg(long, conflicts_with_all = ["canary", "stable"])]
         version: Option<String>,
+
+        /// Upgrade to the latest canary build (rebuilt from every commit).
+        #[arg(long, conflicts_with = "stable")]
+        canary: bool,
+
+        /// Upgrade to the latest stable release — the default on stable
+        /// builds; on a canary build this opts back out of the canary channel.
+        #[arg(long)]
+        stable: bool,
 
         /// Show what would happen without performing the upgrade.
         #[arg(long)]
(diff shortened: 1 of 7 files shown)


colinhacks:
Verified against the live canary (`Canary 0.5.0-canary.20260725.95`, published by the fixed pipeline after #558): `install.sh canary` installs a binary reporting the canary version; a receipt-marked install's `nub upgrade --canary` resolves the release name, downloads from `releases/download/canary/`, sha-verifies, swaps atomically, and recreates `nubx`; `releases/latest` still returns `v0.5.0`; npm dist-tags are `latest: 0.5.0`, `canary: 0.5.0-canary.20260725.95`; GitHub's auto release notes pick `v0.5.0` as previous tag, not `canary`.

colinhacks:

PR #558, opened 2026-07-24
fix(release): grant the canary jobs the permissions the workflow baseline denies

colinhacks:
Two #551 follow-ups that keep the canary pipeline from running:

- The maiden canary run ([30131216108](https://github.com/nubjs/nub/actions/runs/30131216108)) failed at the npm publish with E404 on PUT: the workflow baseline is `permissions: {}` and `canary-publish-npm` declares no permissions block, so it lacks `id-token: write` and the OIDC trusted-publishing PUT goes out unauthenticated. `canary-release` has the same gap for its contents writes. Both jobs now carry the grants of the tag-path jobs they mirror.
- zizmor turned red on #551's merge: the pre-publish gate's `case "${{ github.ref }}"` is an error-level template-injection finding. Replaced with the ambient `$GITHUB_REF`.

Refs #551

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1149,7 +1149,7 @@ jobs:
     env:
       VERSION: ${{ needs.verify.outputs.canary_version }}
     steps:
-      - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd  # v5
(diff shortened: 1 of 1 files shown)


colinhacks:

PR #559, opened 2026-07-25
feat(run): route script bodies through a real POSIX shell (bundled busybox on Windows)

### LINKED ISSUE #399: Cross platform shell for task runner

colinhacks:
`nub run` executes package-script bodies through a real POSIX `sh` on every platform: the system `/bin/sh` on macOS/Linux (already the case) and a bundled busybox-w32 `sh` on Windows, replacing the `cmd /d /s /c` default. One POSIX script body now runs the same on macOS, Linux, and Windows. Spawning a real shell process (rather than an in-process interpreter, the approach explored in the now-closed #448) gives full POSIX coverage and keeps the shell a normal child process that nub spawns and confines like any other.

## What changed

- **`build_script_command`** (`crates/nub-cli/src/cli.rs`): the Windows default resolves a busybox sidecar next to `nub.exe` and invokes `busybox sh -c <body>` with POSIX-`sh` argument escaping. The `--script-shell` flag and `.npmrc` `script-shell` still take precedence, unchanged. On Windows the child gets `TMPDIR` and a best-effort `/tmp` directory for busybox's POSIX temp conventions.
- **Vendored busybox** under `vendor/busybox-w32/` — checksum-pinned prebuilt busybox-w32 `1.38.0-FRP-6075-g169694ebd`, verified byte-for-byte against upstream's published `SHA256SUM` before committing:
  - x64 `busybox64.exe` — `07bb1e5b095b00d68a695481f9240879f33c5724b40aa2308f999d54ed78f075`
  - arm64 `busybox64a.exe` — `e67f873d19d58c535cc9f0c4965ffd622e19b7bab87e3da89cb2185fb54464d7`
  - GPLv2-only, spawned as a separate process (no effect on nub's own license). The GPLv2 `LICENSE` and the corresponding source tarball are vendored alongside for §3 compliance.
- **Packaging** (`.github/workflows/release.yml`): the arch-appropriate binary is staged into each win32 package's `bin/busybox.exe`, behind a SHA-256 gate that fails the release if the vendored bytes drift. `files: ["bin"]` already ships the directory, so no package.json change is needed.
- **Tests**: `ci.yml`'s Windows test leg points an internal resolver override at the vendored binary so the integration suite exercises the busybox path; `tests/busybox-run-probe/` is a branch-scoped windows-latest probe that validates the real sidecar layout end-to-end (`.bin/*.cmd` shim resolution, POSIX-isms, `/tmp` redirection, and the `--script-shell` override reaching a native shell).
- **Docs**: `site/content/docs/runner/run.mdx` documents the cross-platform POSIX-`sh` script shell.

## Scope

This bundles busybox on **Windows only** (Option A): Unix keeps the system `/bin/sh` at no cost. Bundling busybox on Unix too, for byte-identical reproducibility, is a possible follow-up (Option B). Install-lifecycle scripts (run by the embedded aube engine) are unchanged and out of scope here.

Marked `breaking`: Windows script-execution semantics change from cmd.exe to POSIX `sh`.

Closes #399


colinhacks:

PR #571, opened 2026-07-27
feat(upgrade): add mise install channel

afonsojramos:
`nub upgrade` misrouted mise installs: the npm-backend layout contains `node_modules`, so it hit the npm channel and `npm install -g` planted a second global copy that shadows the mise shim.

- Detect both mise layouts (`installs/nub/…`, `installs/npm-nubjs-nub/…`) before the `node_modules` rung.
- Delegate: `mise ls --json <tool>` finds the pinning config, `mise use --path <config> <tool>@<resolved>` installs + rewrites the pin.
- Canary is not served via mise → standalone-installer hint, like Homebrew/winget.

Unit tests cover both layouts and `mise ls` parsing; dry-run + real upgrade from a mise install verified manually.

### DIFF
--- a/crates/nub-cli/src/cli.rs
+++ b/crates/nub-cli/src/cli.rs
@@ -5642,6 +5642,15 @@ enum UpgradeChannel {
     /// discussion #3304) and the next real `winget upgrade` would silently
     /// clobber it. So this channel only prints the command to run.
     Winget,
+    /// mise-managed install (registry/plugin tarball layout or the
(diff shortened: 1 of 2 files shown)


afonsojramos:
Addressed: the FAQ now names mise alongside Homebrew and winget as a stable-only channel, and the upgrade answer notes that a mise-managed install upgrades in place via `nub upgrade` (delegating to mise and rewriting the pin). Also rebased onto main — the previous musl Docker smoke failure was a cancelled image build at the 35-minute timeout, unrelated to this change; the fresh run should clear it.

PR #573, opened 2026-07-27
fix(watch): gate auto-discovered --env-file on Node 20.6, and stop a .env NODE_ENV leaking

colinhacks:
`nub watch` was unusable on Node 18.19–20.5 in any project with a `.env` file: the watched Node aborted with `bad option: --env-file=.env` before executing a line.

## The bug

To make `.env` edits live across restarts (#207), the watch path hands the watched `node --watch` its env files as raw `--env-file=<path>` args, so Node re-reads them on every restart. Node's `--env-file` landed in **20.6.0**, and nub supports Node from **18.19**.

The explicit `--env-file` flags were already gated on that floor by #479. The auto-discovered `.env*` cascade never was — those args were pushed unconditionally. Node rejects an unknown option before running anything, so the result was a dead watcher with no output at all.

Reproduced with only the Node version varying:

| Node | `nub watch probe.cjs` |
| --- | --- |
| v18.19.1 | exit 9, zero stdout, `bad option: --env-file=.env` |
| v24.14.0 | runs and watches normally |

The flag floors were pinned empirically across the installed Nodes rather than read off a comment: v19.3.0 rejects `--env-file` and v20.10.0 accepts it; v22.7.0 rejects `--env-file-if-exists` and v22.11.0 accepts it.

## The fix

Apply the same 20.6.0 floor to the auto cascade. Below it the values fall back to whole-map injection — live-reload of `.env` edits is lost there, but that is the only delivery Node offers, and it is the behavior that predates #207.

The care is in one variable. `raw_env` served double duty: the expansion base for the injected map, and the set of values Node already delivers that the inject loop must skip (injecting one would freeze it at startup). Only the second half may go empty below the floor, so it is now split — `auto_raw_env` stays loaded for the base, `forwarded_raw_env` carries what Node actually receives. On Node 20.6+ the behavior is unchanged.

## Also: a `.env`-set `NODE_ENV` leaked through watch

Forwarding raw files to Node bypasses the load-time drop in `load_env_files_raw_reporting`, so `nub watch` set `NODE_ENV` from `.env` while `nub <file>` on the same project did not — the #263 failure mode (`next build`'s prerender forks inheriting a mode the project never asked for), reachable again through watch.

Re-applied at the process boundary with the guard that already exists for the runtime-control denylist: a canonical placeholder in the preload-free supervisor stops Node's `--env-file` from setting the key, and the child-only `--require` removes it before user preloads run. An **ambient** `NODE_ENV` is the user's own and passes through untouched, exactly as `NODE_OPTIONS` does today.

Guarded only when the auto cascade is the live family. The explicit `--env-file` map does not drop `NODE_ENV` on the direct path either, and the two families never co-occur, so keying on which is live reproduces each one's direct-runner behavior rather than imposing a new policy on the explicit flag.

The drop is also now explained. Watch loaded the cascade through the report-discarding `load_env_files_raw`, so it changed `NODE_ENV` under the user with nothing to point at, while the docs promise a warning on every surface. Both notices are `Once`-guarded, so a long-lived watcher emits each at most once.

## Verification

Every watch run prints, including the version that used to die. `leak` is whether the `.env`-set `NODE_OPTIONS` reached the child.

| run | `GREETING` | expanded | `NODE_ENV` | leak |
| watch v18.19.1 | hello | hello_world | null | false |
| watch v20.10.0 | hello | hello_world | null | false |
| watch v24.14.0 | hello | hello_world | null | false |
| direct v24.14.0 | hello | hello_world | null | false |
| watch, ambient `NODE_ENV=test` | hello | hello_world | test | false |

Gates: `cargo fmt --check` clean, `cargo clippy --all-targets --all-features -- -D warnings` clean, unit tests pass, and both new integration tests pass on v18.19.1 and v24.14.0.

One pre-existing failure is unrelated to this change: `node_compat_env_forces_vanilla_under_watch` fails identically on `origin/main` with the same Node 24.14.0, in the `NODE_COMPAT=1` branch that returns before any code this PR touches.

## Why it shipped, and the test that stops it recurring

Every watch env-file test skips below Node 20.6, so no CI leg exercised the auto cascade on the compat tier — and CI does run `ubuntu-latest` on Node 18.19. The two new integration tests are deliberately ungated and assert only that the values arrive, not which mechanism delivered them, so they cover both sides of the floor.

Refs #207, #479, #263.

### DIFF
--- a/crates/nub-cli/src/cli.rs
+++ b/crates/nub-cli/src/cli.rs
@@ -273,20 +273,32 @@ fn merge_child_env(
 /// injecting a var Node already delivers identically would FREEZE it at the `nub
 /// watch` startup value (Node's `--env-file` never overrides an already-present
 /// env var). Inject a key iff nub's `${VAR}` expansion changed it from the raw
-/// value Node's `--env-file` delivers (`raw_env` is the unexpanded `.env*` merge);
-/// every plain var is left to Node's `--env-file` and live-reloads on restart. A
-/// key absent from `raw_env` is injected — that is the explicit `--env-file` case
-/// (`raw_env` is empty there: auto-discovery is suppressed and Node gets no
-/// `--env-file` args, so injection is the only delivery channel). Pure over its
-/// inputs so the selection can be unit-tested without spawning Node.
+/// value Node actually delivers — `forwarded_raw` is the unexpanded merge of the
+/// files that really reach Node as `--env-file` args; every plain var is left to
+/// Node's `--env-file` and live-reloads on restart. A key absent from
+/// `forwarded_raw` is injected, because injection is then its only delivery
+/// channel: that covers the explicit `--env-file` case (auto-discovery is
+/// suppressed) and every source below Node's 20.6.0 `--env-file` floor, where
+/// nothing is forwarded at all. Pure over its inputs so the selection can be
+/// unit-tested without spawning Node.
 fn watch_inject_vars<'a>(
     env_vars: &'a HashMap<String, String>,
-    raw_env: &HashMap<String, String>,
+    forwarded_raw: &HashMap<String, String>,
 ) -> Vec<(&'a String, &'a String)> {
     env_vars
         .iter()
-        .filter(|(k, v)| raw_env.get(*k) != Some(*v))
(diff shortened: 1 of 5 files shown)


colinhacks:

PR #574, opened 2026-07-27
feat(runtime): polyfill Promise.allKeyed / allSettledKeyed (TC39 await dictionary)

colinhacks:
Adds `Promise.allKeyed` / `Promise.allSettledKeyed` — TC39 [await dictionary](https://github.com/tc39/proposal-await-dictionary), Stage 3 as of the week of 2026-07-20 — as a spec-faithful polyfill in `runtime/polyfills.cjs`. No engine and no Node ships them, so the whole supported range is polyfilled and the feature-detect is the step-aside for whenever V8 lands it, the same posture as Temporal.

```js
const { me, feed } = await Promise.allKeyed({ me: getMe(), feed: getFeed() });
```

Ported deliberately, since each is observable and easy to get wrong: the result object's null prototype, own-enumerable keys including symbols (the proposal's divergence from `combine-promises` / `p-props`, which follow `Object.keys`), reject-vs-sync-throw on a bad argument or `this`, and subclass support through `this`.

Verified by a 69-assertion battery derived from the spec text, green on 18.19.0 / 20.19.0 / 22.14.0 / 22.15.0 / 24.17.0 / 25.9.0 / 26.5.0. Three controls make that meaningful: `typeof Promise.allKeyed` is `undefined` on every version before install, so nothing passes against a native implementation; the battery dies on the first call with the installer removed; and a pre-installed sentinel survives untouched. The `@nubjs/types` declarations carry their own control confirming the mapped type infers concrete properties rather than widening to `any`.

Also lands feature-matrix rows for both globals, the `@nubjs/types` declaration plus positive fixture, the docs Modern APIs table row, and an integration test through the real preload chain.

---

Second commit is **unrelated to the polyfill** — `worker_unhandled_error_is_fatal` failed once on the ubuntu node-24 leg (every other leg green), and it is racy by construction: the fatality is a deferred `process.nextTick` throw, while the fixture's swallow oracle calls `process.exit(0)` straight out of the worker's `exit` event, and `process.exit` does not drain the tick queue. Moving the oracle into `setImmediate` makes the ordering fixed. Positive control: with the `nextTick` throw neutered the hardened fixture still reports the swallow and fails 10/10, so the test keeps its teeth.

---

Third commit adds **`Promise.withResolvers`** (Stage 4 / ES2024, native from Node 22.0), which was absent with no polyfill across the whole 18.19–21.x compat tier. The gap existed because `wiki/research/v0.1-additional-candidates.md` marked it "No action — native on Nub's floor" back when the floor was 22.15; the floor later moved to 18.19 and the verdict was never revisited. Three sibling rows rest on the same premise and are also absent on 18.19–21.x — `Array.fromAsync`, the seven Set methods, and `URL.parse` — left open here rather than bundled, with the survey and its changelog corrected so the premise stops misleading.

Bands were measured empirically across eight Node versions rather than read off release notes, which is also how the 21.x hole in `URL.parse` surfaced. The 15-assertion battery doubles as a differential oracle: it reports whether the method was native before installing, so on Node 22+ it passes against the **engine's** implementation, pinning the polyfill's observable shape to native.

### DIFF
--- a/crates/nub-cli/tests/integration.rs
+++ b/crates/nub-cli/tests/integration.rs
@@ -1287,9 +1287,21 @@ try {
 } catch (e) {
   rejected = e.message;
 }
+// Promise.withResolvers: native on Node 22+, polyfilled on the 18.19-21.x compat
+// tier where it was previously missing entirely.
+const wr = Promise.withResolvers();
+wr.resolve("wr-ok");
+const withResolvers =
+  Object.getPrototypeOf(wr) === Object.prototype &&
+  Object.keys(wr).join(",") === "promise,resolve,reject" &&
+  (await wr.promise) === "wr-ok";
 // The additive contract: nub's additions stay invisible to enumeration.
-const hidden = Object.keys(Promise).filter((k) => k.endsWith("Keyed")).length;
-console.log(JSON.stringify({ keyed, scoped, nullProto, statuses, rejected, hidden }));
+const hidden = Object.keys(Promise).filter(
+  (k) => k.endsWith("Keyed") || k === "withResolvers",
+).length;
+console.log(
+  JSON.stringify({ keyed, scoped, nullProto, statuses, rejected, withResolvers, hidden }),
+);
 "#,
     )
     .unwrap();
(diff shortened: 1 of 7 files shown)


colinhacks:
Superseded by #584, merged as 74eca575a — these three commits are its first three.

Issue #581, opened 2026-07-28
minimumReleaseAge: missing publish-time data fails open, even in strict mode


afonsojramos:
## Context

While root-causing the `nub update` age-gate bypass (see the linked PR), a design question surfaced that is independent of that fix.

`pick_version` treats a candidate version with no `time:` entry as eligible — the cutoff is simply ignored for that version (documented in `semver_util.rs`: *"When the packument has no time entry for a version (e.g. abbreviated corgi payload in Highest mode), the cutoff is ignored and the version stays eligible"*). That behavior mirrors pnpm and keeps registries that never serve time data (some private mirrors, older Verdaccio) usable.

But it means the gate's guarantee is only as strong as the metadata pipeline that feeds it: any code path (or registry) that yields a time-less packument silently disables `minimumReleaseAge` — including `minimumReleaseAgeStrict=true`, whose whole point is to fail closed. The update-path bug proved this class of failure is real and invisible to the user: no warning, no error, just a fresh publish sailing through a strict 24-hour floor.

## Question

Should strict mode treat missing publish-time data as a **hard failure** (or at least a loud warning) instead of failing open?

Options, roughly in increasing strictness:

1. **Status quo**: missing time ⇒ eligible, silent (pnpm parity).
2. **Warn**: missing time ⇒ eligible, but emit a per-package warning when a cutoff is active, so a disabled gate is at least visible.
3. **Strict fails closed**: under `minimumReleaseAgeStrict=true`, a candidate with no time entry is treated like an age-gated version (excluded; hard error if nothing else satisfies). Loose mode keeps the status quo. Registries that never serve time data would need `minimumReleaseAgeExclude` (or a per-registry opt-out like `registrySupportsTimeField`'s sibling) to keep working.

Option 3 matches the promise nub's defaults make (*"a real 24-hour trust floor, not an advisory one"* — `nub_setting_defaults`), with option 2 as a cheap interim. Happy to implement whichever direction maintainers prefer.

colinhacks:
Fixed in #602. A missing publish time now resolves against the packument's `modified` timestamp, an upper bound that proves the whole document mature; when that can't settle it, strict mode blocks instead of admitting the version. Lenient mode is unchanged, so standalone behaviour is preserved.

The vulnerability re-pick shares the same wall, but ranks rather than vetoes: an undated safe version still beats keeping a known-vulnerable one.

On `main`, not yet on a published release.

colinhacks:
[...]

PR #582, opened 2026-07-28
fix(resolver): enforce minimumReleaseAge without the full-packument cache

afonsojramos:
## Problem

`nub update` crosses `minimumReleaseAge` onto publishes minutes old — versions that `nub install` (correctly) refuses — and rewrites `package.json` onto them. Nub's own strict 1440-minute default is silently disabled on the whole update path; the same hole covers `add`/`dedupe`/`audit` via `build_resolver`.

Live repro (while `@types/node@26.1.2` was ~20h old):

```
$ nub install          # range ^26.1.1
+ @types/node@26.1.1  latest 26.1.2     # gate holds
$ nub update
+ @types/node@26.1.2                     # gate gone; manifest rewritten to ^26.1.2
```

The resulting lockfile is then rejected by pnpm 11 (`ERR_PNPM_MINIMUM_RELEASE_AGE_VIOLATION`) in pnpm-compat projects, which is how this was found.

## Root cause

`update`/`add`/`dedupe`/`audit` deliberately resolve without the full-packument disk cache (`cache_full_packuments: false` — the dist-tag freshness rule). When `needs_time` is on and that cache dir is `None`, the initial fetch (`fetch.rs`) and all three pick-site heal arms (`driver.rs`) fall back to `client.fetch_packument()` — the **abbreviated corgi document, which npmjs serves without a `time` map**. A version with no publish time bypasses the cutoff inside `pick_version`, so the gate self-destructs exactly on the paths that rewrite manifests.

Instrumented pick-site trace, same project, same cutoff:

```
install: PICK @types/node ^26.1.1 cutoff=… strict=true time_len=2357 -> Found 26.1.1
update:  PICK @types/node ^26.1.1 cutoff=… strict=true time_len=100  -> Found 26.1.1   (primer)
         PICK @types/node ^26.1.1 cutoff=… strict=true time_len=0    -> Found 26.1.2   (post-refetch corgi)
```

## Fix

Add `RegistryClient::fetch_packument_with_time` — the uncached sibling of `fetch_packument_with_time_cached` (full Accept header via `fetch_packument_json_fresh`, `time` included, no disk cache, so the dist-tag freshness property is preserved) — and use it in all four `needs_time` fallbacks.

## Test

The regression test mocks npmjs's Accept-dependent shape (corgi without `time`, full document with it) and resolves with `minimumReleaseAge` active but no full cache configured — red before the fix (picks the immature version), green after.

A Verdaccio bats test can't catch this: Verdaccio serves the `time` map even for the corgi Accept header, unlike npmjs, so the unit test pins the registry shape explicitly.

Verified end-to-end against npmjs: `nub update` now stays on the mature version ("Already up to date") and leaves the manifest untouched.

Related design question (independent of this fix): #581 — whether strict mode should fail closed on missing time data at the pick site.

### DIFF
--- a/vendor/aube/crates/aube-registry/src/client/packument.rs
+++ b/vendor/aube/crates/aube-registry/src/client/packument.rs
@@ -349,6 +349,22 @@ impl RegistryClient {
         Ok(packument)
     }
 
+    /// Fetch the full (non-corgi) packument fresh — no disk cache — and
+    /// parse it into [`Packument`], `time` map included. The uncached
+    /// sibling of [`Self::fetch_packument_with_time_cached`], for callers
+    /// that need publish times (`minimumReleaseAge`, time-based mode,
+    /// `trustPolicy=no-downgrade`) while deliberately running without the
+    /// full-packument disk cache (`aube update`'s dist-tag freshness
+    /// rule). Falling back to [`Self::fetch_packument`] there instead
+    /// returns the abbreviated corgi document, whose missing `time` map
+    /// silently disables the age gate at the pick site.
+    pub async fn fetch_packument_with_time(&self, name: &str) -> Result<Packument, Error> {
+        let value = self.fetch_packument_json_fresh(name).await?;
+        let packument: Packument = serde_json::from_value(value)
+            .map_err(|e| Error::Io(std::io::Error::new(std::io::ErrorKind::InvalidData, e)))?;
+        Ok(packument)
+    }
+
     pub async fn fetch_packument_with_time_cached_after_lookup(
         &self,
         name: &str,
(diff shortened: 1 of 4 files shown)


colinhacks:
Merged as part of #607, with your commit and authorship intact — thanks for tracking this down.

#602 landed the pick-site half first, which made the two halves inseparable: with an undateable version failing closed, the corgi fallback this fixes turned from a silent bypass into a hard `AgeGate` abort on `update`/`add`/`dedupe`/`audit`. Rebasing this branch onto that would have meant rewriting your commit, so the combined PR carried it forward instead.

Two changes on top: the regression test now runs in strict mode (Nub's default — in lenient mode the lowest-satisfying fallback picks the mature version regardless, so it passed for the wrong reason once #602 landed), and the corgi fixture keeps `modified`, which npmjs does serve.

afonsojramos:
Thanks!

PR #583, opened 2026-07-28
fix(manifest): preserve package.json indent, line endings, and EOF newline

afonsojramos:
## Problem

Any package.json rewrite through the aube command layer — `update`/`add`/`audit` dep-section syncs and the namespace settings editors — reformats the whole file: a tab- or 4-space-indented manifest is reindented to two spaces, CRLF becomes LF, and a missing EOF newline gets appended. A one-key change diffs as the entire file:

```
$ cat package.json   # tab-indented
{
	"name": "x",
	"devDependencies": { … }
}
$ nub update         # bumps one specifier…
$ git diff --stat package.json
 1 file changed, 44 insertions(+), 44 deletions(-)   # …and reindents everything
```

npm and pnpm both detect and reproduce the file's own style (`detect-indent` in `@npmcli/package-json` / `@pnpm/read-project-manifest`). nub-core's pin writer (`edit_manifest` in `pm/resolve.rs`) already does this for the `devEngines` stamp — the command-layer writers just never got the same treatment.

## Root cause

Six writers serialize with `serde_json::to_string_pretty` (hardwired two-space indent, LF) and append `\n` unconditionally:

- `aube/src/commands/manifest_io.rs`: `write_manifest_json`, `update_manifest_json_object`
- `aube-manifest/src/workspace/edits.rs`: `remove_setting_entry`, `edit_setting_map`, `add_to_pnpm_only_built_dependencies`, `set_pnpm_allow_builds_entries`

## Fix

Add `detect_json_style` / `serialize_json_with_style` to `aube-manifest` (first indented line's whitespace run as the indent unit, CRLF reproduction, trailing-newline state — the same rules as nub-core's `detect_format`) and route all six writers through them. Files with no indentation to imitate keep the previous two-space/LF/trailing-newline output, so freshly created manifests are byte-identical to before.

## Tests

- Style helpers: tab/4-space/CRLF/EOF-newline detection, style-faithful serialization, and a byte-for-byte round-trip of a tabbed document.
- `write_manifest_dep_sections` preserves tab indentation (the `nub update` repro) and never two-space-reindents.
- `write_manifest_json` preserves CRLF + absent trailing newline.
- `edit_setting_map` preserves tab indentation through a settings write.

Verified end-to-end: `nub update` on a tab-indented manifest bumps the one specifier and leaves every other byte alone (including the devEngines stamp interplay).

### DIFF
--- a/vendor/aube/crates/aube-manifest/src/lib.rs
+++ b/vendor/aube/crates/aube-manifest/src/lib.rs
@@ -1590,6 +1590,62 @@ pub fn serialize_json_with_indent<T: serde::Serialize>(
         .map_err(|e| serde_json::Error::io(std::io::Error::new(std::io::ErrorKind::InvalidData, e)))
 }
 
+/// The surface style of an existing JSON manifest: indent unit, line-ending
+/// flavor, and trailing-newline state. Reproducing all three keeps an
+/// `update`/`add`/settings edit diffing as the changed keys rather than as a
+/// whole-file reformat. Deliberately the UNION of what the reference PMs
+/// preserve, since each drops one: npm reproduces the indent and line ending
+/// but always appends a final newline, pnpm reproduces the indent and the
+/// trailing-newline state but never CRLF.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct JsonStyle {
+    pub indent: String,
+    pub crlf: bool,
+    pub trailing_newline: bool,
+}
+
+impl Default for JsonStyle {
+    /// The style used when there is no original to imitate (a manifest created
+    /// from scratch): two-space indent, LF, trailing newline.
+    fn default() -> Self {
+        JsonStyle {
+            indent: "  ".to_string(),
+            crlf: false,
+            trailing_newline: true,
+        }
+    }
+}
+
+/// Detect the [`JsonStyle`] of an existing JSON document. Indent detection is
+/// [`detect_json_indent`]; a raw `\r\n` can only be a line terminator (JSON
+/// escapes a carriage return inside a string), so its presence marks the file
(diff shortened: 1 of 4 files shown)


afonsojramos:
Good catch — folded `patch-commit`/`patch-remove` into the PR (same detect-before-parse + style-serialize pattern, plus a tab-preservation test). Left `deploy/rewrite.rs` on `to_string_pretty` deliberately since it writes a generated deploy artifact rather than the user's source manifest. Also rebased onto main.

afonsojramos:
Right again — the fixture had no lockfile, so the write routed through `config_write_target` → `edit_setting_map` and the test duplicated the `edits.rs` coverage without touching the changed writer. Replaced it with `bun_patch_writers_preserve_tab_indentation`: a `bun.lock` in the fixture takes the `LockfileKind::Bun` arm into `upsert_manifest_patched_dependency`, the assertion on the top-level `patchedDependencies` field proves the routing reached that writer, and the removal leg covers `remove_bun_patched_dependency` the same way. Verified red with the two writer substitutions reverted, green with them in place.

colinhacks:
Rebased onto main past the aube 2.2.9 sync. Upstream had added indent detection since, so the branch now carries only what main still lacks, CRLF and trailing-newline preservation plus indent at the three writers that were still on `to_string_pretty`, built on upstream's helpers. Two small changes on the way: kept main's read-error handling in the manifest writer instead of `unwrap_or_default`, and a new manifest gets the default style so it still ends with a newline. Pushed to your branch.

colinhacks:
Shipped in v0.9.0: https://github.com/nubjs/nub/releases/tag/v0.9.0 — thanks for the contribution.

PR #584, opened 2026-07-28
feat(runtime): polyfill every Stage 3+ library surface missing under nub

colinhacks:
Rebased onto `main`; **supersedes #574**, whose three commits are the first three here. #574 can be closed.

Everything Stage 3 or later that was missing under nub, enumerated from tc39/proposals + finished-proposals.md and then **measured** across eight Node versions rather than read off release notes.

| Surface | Was |
|---|---|
| `Promise.allKeyed` / `allSettledKeyed` | no engine ships these |
| `Promise.withResolvers` | native 22+, absent below |
| Set `union`/`intersection`/`difference`/`symmetricDifference`/`isSubsetOf`/`isSupersetOf`/`isDisjointFrom` | native 22+, absent below |
| `Array.fromAsync`, `Iterator` + `Iterator.from` + 11 helpers | native 22+, absent below |
| `Iterator.concat`, `Map`/`WeakMap` `getOrInsert`/`getOrInsertComputed` | native 26+, absent below |
| `Iterator.zip` / `zipKeyed`, `Iterator.chunks`/`windows`/`includes`/`join` | no engine ships these |
| `Math.sumPrecise`, `Symbol.metadata` | no Node ships them |
| `URL.parse`, `String.isWellFormed`/`toWellFormed`, `Object`/`Map.groupBy`, Array+TypedArray copy methods, `ArrayBuffer.transfer`/`detached`, `Atomics.pause` | native above their line, absent below |

**No-clobber, enforced mechanically.** Every installer guards per *method*. `polyfills_never_clobber_native` plants a sentinel on ~57 surfaces, runs the installer, and fails if one was replaced. Zero clobbered on 18.19/20.10/21.0/22.15/26.5, and it runs under the host node so the CI matrix covers it across versions.

**Native as the oracle.** Each battery runs identical assertions against the polyfill and against a runtime that ships the surface natively; bun is the oracle for `Math.sumPrecise`. That caught 13 real fidelity bugs before merge — seven pre-review (including `Math.sumPrecise` returning `NaN` on overflow inputs, and helpers not forwarding early closure to the source iterator) and the six from review.

**Review findings all fixed and covered**, each confirmed against an oracle first: `toSpliced(start)` truncation, `Array.fromAsync` negative array-like length, `transfer` vs `transferToFixedLength` resizability, `Math.sumPrecise([-0])` sign, and three `Iterator.zip` divergences (string primitive, padding over-consumption, strict-mode throw point). The `isWellFormed` receiver report had a deeper cause: sloppy-mode `this` coercion made every `RequireObjectCoercible` check silently pass, so the file is now `"use strict"` — which fixes the whole class rather than one method.

**Deliberately excluded:** `JSON.rawJSON`/`isRawJSON` would need native `JSON.stringify` replaced to serialize markers as raw text, the one change here that would overwrite a native builtin. `Error.prototype.stack` accessor would modify existing semantics. RegExp `v`/`\A\z`/dup-captures and WeakMap symbol keys are engine-level.

**Known limit:** helper objects are generators, so their prototype is not `%IteratorHelperPrototype%` and `Symbol.toStringTag` reads `"Generator"`. Iteration, laziness, validation and source-closing are correct.

### DIFF
--- a/crates/nub-cli/tests/integration.rs
+++ b/crates/nub-cli/tests/integration.rs
@@ -1287,9 +1287,21 @@ try {
 } catch (e) {
   rejected = e.message;
 }
+// Promise.withResolvers: native on Node 22+, polyfilled on the 18.19-21.x compat
+// tier where it was previously missing entirely.
+const wr = Promise.withResolvers();
+wr.resolve("wr-ok");
+const withResolvers =
+  Object.getPrototypeOf(wr) === Object.prototype &&
+  Object.keys(wr).join(",") === "promise,resolve,reject" &&
+  (await wr.promise) === "wr-ok";
 // The additive contract: nub's additions stay invisible to enumeration.
-const hidden = Object.keys(Promise).filter((k) => k.endsWith("Keyed")).length;
-console.log(JSON.stringify({ keyed, scoped, nullProto, statuses, rejected, hidden }));
+const hidden = Object.keys(Promise).filter(
+  (k) => k.endsWith("Keyed") || k === "withResolvers",
+).length;
+console.log(
+  JSON.stringify({ keyed, scoped, nullProto, statuses, rejected, withResolvers, hidden }),
+);
 "#,
     )
     .unwrap();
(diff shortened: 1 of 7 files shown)


colinhacks:

PR #589, opened 2026-07-28
build: seed isolated target dirs from the shared cache; content-key the buckets

colinhacks:
The wrapper claimed a relocated target dir gets 0% reuse, so isolation meant a cold build. Measured on the REAL 153-crate workspace with controls both ways: rebuilding in place rebuilt 0, and the same source+command against a CoW clone at a new path also rebuilt **0** — relocation is free. A cold dir rebuilt all 153. The 0%-reuse claim holds for sccache (it keys on the rustc command line, embedding absolute paths), not for a dir copy.

22 of 28 worktrees are isolated at any time, each paying that 153-crate rebuild. Cloning 6.4 GB of warm artifacts took 14s (CoW).

- Seed a private dir from a bucket, staged under a claim dir and published by rename so cargo never sees a half-copied tree.
- Content-key buckets: a merge-base proved only that a worktree matched its *own* base (4 contents measured among 6 "shared" worktrees).
- Seeding falls back to any bucket, since a key hashes the index and buckets only exist for base content — without it, committed divergence never seeded.
- `touch` the target: verified `mkdir -p` and nested writes do not refresh mtime, so the GC could collect a live bucket.
- Exclude two leaves nothing links: `nub-native`, `nub-phantom`.

A cross-worktree seed with real source drift saved 77 of 153 crates; a matching seed saves all 153.


colinhacks:

PR #592, opened 2026-07-28
build: run the analysis gates under --profile fast, matching the dev loop

colinhacks:
The dev loop uses `--profile fast`; `make verify` and CI's check/clippy used `dev`, sharing nothing — crossing between them rebuilt 153/153 units, and the host held 12 GB of `target/fast` beside 14 GB of `target/debug`. `fast` inherits `dev`, differing only in debuginfo.

- check + clippy (root and `nub-native`) move to `--profile fast`, in CI and `make verify`.
- Test jobs and the `nub-phantom` check stay on `dev`: the former reuse debug artifacts by path, the latter defines no `fast` profile.
- Fixes Makefile `PROFILE`: `CARGO_FLAGS` was empty for non-release profiles, so `PROFILE=fast` built into `target/debug` while the copy read `target/fast`.


colinhacks:

PR #594, opened 2026-07-28
scripts: run the CI gates on ephemeral GCE spot VMs

colinhacks:
Adds `scripts/remote-build.ts` — dispatch a Rust build, clippy gate, or test suite to an ephemeral GCE spot VM instead of the dev Mac, and for a darwin build cross-compile `aarch64-apple-darwin` on Linux and pull the signed binary back.

## Why

Measured on the dev host under normal agent load: ~30% of CPU sits **idle** while the load average reads 155, sys time is ~25%, and the disk sustains 3000–4000 tps at 5–6 KB/transfer. The bottleneck is cargo fingerprint churn across a dozen multi-GB target dirs on one APFS volume, not compute — so each remote builder helps mainly by bringing its own disk.

## What goes remote, and what does not

| Job | Remote | Mac | Verdict |
|---|---|---|---|
| warm incremental | 8.1s | ~5s | **stays local** |
| cold `release` | 7m 00s | ~15m | remote, 2× |
| `clippy --all-targets --all-features` | 35.3s | — | remote |
| `cargo test` | 39.4s warm | — | remote (718 passed / 0 failed on Linux) |

The inner loop is deliberately not a job type.

## Verified end to end

- `--fanout 10 --job clippy` → **10/10 succeeded, 322s wall**, all VMs auto-reaped. All ten hit `no c3-standard-8 capacity in us-central1-a` simultaneously and fell back to `us-central1-b`, which is why placement walks a zone × machine grid.
- `--job build --profile release` → binary + addon pulled and **run on macOS**: `--version`, a TypeScript file with a non-erasable `enum` (real oxc path through the addon), and `nub install is-odd@3.0.1` over TLS, which exercises `rustls-platform-verifier` into Security.framework and proves the stubs are runtime-correct, not just link-correct.

## Cross-compiling needs no Apple SDK

`cargo-zigbuild` + zig **0.16.0** (pinned — 0.14.1/0.15.2 SIGSEGV in the Mach-O linker with an empty `= note:` and a zero-byte output). zig ships its own `libSystem.tbd`, and `scripts/darwin-stubs/` supplies symbol-only TBD stubs for the Apple frameworks nub links transitively. No Apple SDK is installed on the builder, so the SDK licence never applies.

## Orphan-proofing

Three layers, because a local `finally` dies with a SIGKILL: signal handlers; a server-side `--max-run-duration` + termination action on **every** instance (builders DELETE at 45m, the bake STOPs at 90m since it images its own disk); and a `nub-builder=1` label that `--reap` sweeps. `--reap` only touches VMs older than 90m, so it cannot destroy a sibling agent's live build.

## Known gap

The golden image is registry-warm but **not artifact-warm** — the bake's warm step stops after `cargo fetch` for reasons not yet root-caused (tracing is on). Builders therefore cold-compile, so every timing above is pessimistic rather than wrong. Correctness is unaffected.

Also unresolved from the underlying research: cross-compiled artifacts come out with a **macOS 13.0** deployment target rather than 11.0, so this must not go near the release path until that is solved.

Refs #573


colinhacks:

Issue #341, opened 2026-07-06
`enableGlobalVirtualStore=false`: relinking an existing git-url virtual store entry fails with "File exists (os error 17)"


adamhl8:
With a project-local virtual store (`enableGlobalVirtualStore=false`), any install that performs a full link pass over a virtual store that already contains a git/url-sourced package entry fails while linking that entry's dependencies. Registry-sourced entries are unaffected, and the same flow works fine with the default global virtual store.

### Reproduction

`package.json` with a dependency that (transitively) pulls in a git-url package. In my case `obsidian-dataview`, which declares `"@codemirror/language": "https://github.com/lishid/cm-language"`:

```json
{
  "devDependencies": {
    "obsidian-dataview": "^0.5.68"
  }
}
```

```sh
nub install                 # fresh install, succeeds
nub i --no-frozen-lockfile  # fails
```

The second command fails with:

```
  × failed to link node_modules
  ╰─▶ I/O error at <project>/node_modules/.nub/@codemirror+language@url+7b3dff266b55084c/node_modules/@codemirror/
      state: File exists (os error 17)
```

colinhacks:
Shipped in v0.3.0: https://github.com/nubjs/nub/releases/tag/v0.3.0

colinhacks:
Shipped in v0.3.1: https://github.com/nubjs/nub/releases/tag/v0.3.1

PR #595, opened 2026-07-29
fix(linker): reclaim a per-project virtual store when relinking on Windows

### LINKED ISSUE #566: [Bug Report] Fail to remove dependencies
console stderr:
```Updated package.json
nub 0.6.0
███████████░░░░  131/185 pkgs · ETA 1.1s
  × failed to link node_modules
  ╰─▶ I/O error at \\?\C:\Users\Administrator\TestEnvironment\meta-
      magic_mount-zig\webui\node_modules\.store\@babel+helper-string-
      parser@7.29.7: Cannot create a file when that file already exists. (os error 183)
```

### LINKED ISSUE #576: nub add fails on Windows with "Cannot create a file when that file already exists (os error 183)"
- OS: Windows 11
- Nub: v0.6.0
- Node: v24.11.0

Steps to reproduce

1. Run `nub init -y`
2. Run `nub install`
3. Run `nub add -D oxlint oxfmt`

Expected

The packages are added to package.json and linked into node_modules.

Actual

package.json is updated, but the install fails with:

failed to link node_modules
I/O error at ... Cannot create a file when that file already exists. (os error 183)

The new packages are not linked into node_modules.

Workaround

Delete node_modules and run:

nub install

After that, the packages are installed correctly.

This happens consistently whenever I add new packages to an existing project.

colinhacks:
`nub add` / `nub remove` aborted on Windows with `os error 183`, leaving the tree unlinked.

`read_link` failing with `InvalidInput` was the "is a real directory" test. That errno is Unix-only, so `detect_aube_dir_gvs_mode` never returned `Some(false)` on Windows and `reset_on_mode_change` never wiped a per-project tree. The surviving directories were cleared with a non-recursive `remove_dir`, so junction creation aborted.

`is_real_dir` now keys on whether the entry is a link; cleanup uses `try_remove_entry`, as the git/url path already did (#341).

aube's suite ran ubuntu-only and the covering test was `unix`-gated. Both fixed, plus a PR-gating windows-latest leg.

Closes #566
Closes #576

### DIFF
--- a/.github/workflows/aube-parity.yml
+++ b/.github/workflows/aube-parity.yml
@@ -40,10 +40,10 @@ on:
       - "tests/aube-conformance/**"
   schedule:
     # Nightly at 06:30 UTC. The Windows conformance leg is gated off
-    # pull_request (it is the only Windows-runner contributor in this workflow);
-    # the nightly schedule keeps that Windows coverage running on a cadence even
-    # between vendor/aube bumps. Scheduled runs ignore the path filters above and
-    # exercise the full workflow, Windows leg included.
+    # pull_request; the nightly schedule keeps that coverage running on a
+    # cadence even between vendor/aube bumps. (The Windows `cargo test` leg
+    # does gate PRs — see the rationale on that job.) Scheduled runs ignore
+    # the path filters above and exercise the full workflow.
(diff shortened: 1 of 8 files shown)


colinhacks:

PR #597, opened 2026-07-29
fix(install): link bins of hoisted transitive deps

### LINKED ISSUE #570: Hoisted layout does not link bins of hoisted transitive dependencies, breaking lifecycle scripts
With `--node-linker hoisted`, nub links bins only for the root project's direct dependencies. Bins that belong to hoisted transitive dependencies are never linked into `node_modules/.bin`, so any lifecycle script that invokes one exits 127. `--node-linker isolated` is unaffected — each package gets its own `node_modules/.bin` — and pnpm's hoisted layout links all of them.

The common case is `node-gyp-build`, which most prebuild-based native packages call from their `install` script.

## Reproduction

```jsonc
// package.json
{
  "name": "bin-repro",
  "private": true,
  "dependencies": { "bufferutil": "4.1.0" }
}
```

```yaml
# pnpm-workspace.yaml
allowBuilds:
  bufferutil: true
```

```bash
nub install --node-linker hoisted
```

```
sh: 1: node-gyp-build: not found
  × lifecycle script install failed for bufferutil@4.1.0: script `install`
  │ exited with code 127
```

`bufferutil` declares `node-gyp-build` as a dependency and its `install` script is `node-gyp-build`. The package is present at `node_modules/node-gyp-build` and its manifest declares three bins, but none of them are linked.

## Result

| Install                          | `.bin/node-gyp-build` | exit |
| -------------------------------- | --------------------- | ---- |
| **nub 0.6.0, hoisted**           | **missing**           | **1** |
| nub 0.6.0, isolated              | in `.store/bufferutil@4.1.0/node_modules/.bin` | 0 |
| pnpm 11.17.0, hoisted (control)  | present               | 0    |

At repo scale the same failure shows up on the first native package nub reaches. Installing a 22-workspace monorepo (2819 packages) from a `pnpm-lock.yaml` fails on `bufferutil` with the identical message; the same install with `--node-linker isolated` completes cleanly in 3.4s.

## Expected

The hoisted layout should link bins for hoisted transitive dependencies into `node_modules/.bin`, the way pnpm's hoisted layout does, so lifecycle scripts can resolve them from `PATH`.

## Environment

- nub 0.6.0
- Node v24.18.0
- Linux 6.12.94 x86_64
- Control: pnpm 11.17.0

colinhacks:
Under the hoisted linker, bins were linked only for the root importer's direct deps and the per-dep pass skipped hoisted, so a hoisted transitive's bins landed nowhere and any lifecycle script invoking one exited 127.

Mirrors pnpm's hoisted layout: every package in a `node_modules/` gets its bins in that directory's `.bin`, driven off `HoistedPlacements`. The per-dep pass now runs under hoisted too, since a script gets only its containing directory's `.bin` on PATH. Passes run lowest-precedence first (dep-lifecycle, placements, direct deps, self-bin), so a direct dep still wins a name collision.

Verified on the issue's `bufferutil@4.1.0` fixture: hoisted goes 1 → 0.

Closes #570

### DIFF
--- a/vendor/aube/crates/aube-scripts/src/lib.rs
+++ b/vendor/aube/crates/aube-scripts/src/lib.rs
@@ -1434,14 +1434,12 @@ fn unshare_one_file(path: &Path, mode: u32) -> std::io::Result<()> {
 /// `node_modules/.aube/<dep_path>/node_modules/<name>`). The manifest
 /// is the dependency's own `package.json`, *not* the project root's.
 ///
-/// `dep_modules_dir` is the dep's sibling `node_modules/` — i.e.
-/// `package_dir`'s parent for unscoped packages, or `package_dir`'s
-/// grandparent for scoped (`@scope/name`). `<dep_modules_dir>/.bin`
-/// is prepended to `PATH` so the dep's postinstall can spawn tools
-/// declared in its own `dependencies` (the transitive-bin case —
-/// `prebuild-install`, `node-gyp`, `napi-postinstall`). The install
-/// driver writes shims there via `link_dep_bins`; `rebuild` mirrors
-/// the same pass.
+/// Every `<modules_dir>/.bin` from the package's own out to the project
+/// root is prepended to `PATH`, closest first, so the dep's postinstall
+/// can spawn tools declared in its own `dependencies` (the transitive-bin
+/// case — `prebuild-install`, `node-gyp`, `napi-postinstall`). The install
+/// driver writes those shims via `link_dep_bins` (isolated) and
+/// `link_hoisted_placement_bins` (hoisted).
 ///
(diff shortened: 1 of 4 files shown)


clemens:
Should this be in the latest canary already?

I'm asking because I have a similar case (different package, but same failure) and it's not working for me either (neither with 0.6.0 nor with 0.6.0-canary.20260804.220). Repo here, if you need it: https://github.com/clemens/nub-bug-repro.

Command I'm using for the canary: `XDG_DATA_HOME=$(mktemp -d) XDG_CACHE_HOME=$(mktemp -d) npx -y @nubjs/nub@0.6.0-canary.20260804.220 install --frozen-lockfile`

colinhacks:

PR #599, opened 2026-07-29
fix(resolve): probe extensions for bare package subpaths

### LINKED ISSUE #562: Extensionless package subpath imports fail with ERR_MODULE_NOT_FOUND (tsx resolves them)
## Summary

`nub` fails to resolve **extensionless subpath imports into packages** (`pkg/subpath`), raising `ERR_MODULE_NOT_FOUND`, in cases where `tsx` resolves them fine. This affects both workspace packages whose entry point is TS source and ordinary third-party CJS packages.

This bites hard in a TS monorepo, because it is **invisible to `tsc --noEmit`** — TypeScript resolves these specifiers happily (they're legal under `moduleResolution: bundler`/`node`), so a full type-check passes and the failure only appears at runtime, per-script.

## Environment

- nub `v0.1.14`
- Node `v24.12.0`
- macOS (arm64)
- Compared against tsx `4.20.3`

## Reproduction

```sh
rm -rf /tmp/nub-subpath && mkdir -p /tmp/nub-subpath/packages/lib \
  /tmp/nub-subpath/node_modules/@repro /tmp/nub-subpath/node_modules/thirdparty/components
cd /tmp/nub-subpath
echo '{"name":"root","private":true,"workspaces":["packages/*"]}' > package.json
echo '{"compilerOptions":{"moduleResolution":"bundler","module":"esnext","target":"esnext"}}' > tsconfig.json

# (A) workspace package whose entry is TS source, no "exports" map
echo '{"name":"@repro/lib","main":"./index.ts"}' > packages/lib/package.json
echo 'export const root = 1;'                    > packages/lib/index.ts
echo 'export const createPrompt = () => "hi";'   > packages/lib/prompt.ts
ln -s ../../packages/lib node_modules/@repro/lib

# (B) third-party CJS package, no "exports" map, subpath file
echo '{"name":"thirdparty","main":"index.js"}'   > node_modules/thirdparty/package.json
echo 'module.exports={};'                        > node_modules/thirdparty/index.js
echo 'module.exports={lang:"python"};'           > node_modules/thirdparty/components/prism-python.js

echo 'import { createPrompt } from "@repro/lib/prompt";' > a-workspace.ts
echo 'console.log("A ok:", createPrompt());'            >> a-workspace.ts
echo 'import "thirdparty/components/prism-python";'      > b-thirdparty.ts
echo 'console.log("B ok");'                             >> b-thirdparty.ts

for f in a-workspace b-thirdparty; do
  printf 'tsx: '; npx tsx $f.ts 2>&1 | grep -E "ok|Cannot find module" | head -1
  printf 'nub: '; nub     $f.ts 2>&1 | grep -E "ok|Cannot find module" | head -1
done
```

## Actual

```
tsx: A ok: hi
nub: Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/private/tmp/nub-subpath/node_modules/@repro/lib/prompt' imported from /private/tmp/nub-subpath/a-workspace.ts
tsx: B ok
nub: Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/private/tmp/nub-subpath/node_modules/thirdparty/components/prism-python' imported from /private/tmp/nub-subpath/b-thirdparty.ts
```

## Expected

Both resolve, as they do under `tsx` (and as `tsc` accepts them):

- (A) `@repro/lib/prompt` → `packages/lib/prompt.ts`
- (B) `thirdparty/components/prism-python` → `node_modules/thirdparty/components/prism-python.js`

## Notes

The docs/skill state that "Extensionless `.ts` imports and `.js`→`.ts` rewrites resolve like `tsc`". That holds for **relative** specifiers and for `compilerOptions.paths`, but not for **bare package subpaths** like the two above — which is the gap here. Case (B) additionally shows it isn't TS-specific: an extensionless subpath into a plain CJS dependency (the shape `prismjs/components/prism-*` uses) also fails.

## Real-world impact

Hit while evaluating nub as a `tsx` replacement in a TypeScript monorepo. The pattern in (A) — workspace packages with `"main": "./index.ts"` and no `exports` map, imported by subpath — is used by 38 packages there, so a large share of scripts fail at runtime under nub while `tsc --noEmit` stays green across the whole repo. Case (B) broke a script via `prismjs/components/prism-python`.

Working around it would mean either adding `exports` maps to every package or rewriting import specifiers to carry extensions, both of which are substantial changes to shared package surfaces — so we've reverted to `tsx` for now. Happy to test a patch.

colinhacks:
Extensionless subpaths (`pkg/sub`) got no probing, failing `ERR_MODULE_NOT_FOUND` where `tsc` and `require()` both work — `resolve_ts` probed relative specifiers only.

Probing now covers bare subpaths. A dependency declaring `exports` is never probed: key presence defers to Node, so a withheld subpath stays an error (`"exports": null` included). Order `.js .json .ts .tsx`, importer-independent. Result is realpathed, so a symlinked workspace package lands outside `node_modules` and transpiles.

Not gated on a TypeScript importer, matching tsx. Additive.

7 tests; docs updated. Parity corpus reports 3 divergences, identical on unmodified `main`.

Closes #562

### DIFF
--- a/crates/nub-cli/tests/bare_subpath_resolution.rs
+++ b/crates/nub-cli/tests/bare_subpath_resolution.rs
@@ -222,6 +222,30 @@ console.log("LOADED");
     );
 }
 
+/// A traversing subpath must never resolve outside the package root. The guard
+/// splits on both separators, since lexical normalization honors `\` on Windows;
+/// this pins the separator-independent half on every platform.
+#[test]
+fn traversing_subpath_never_escapes_the_package_root() {
+    let dir = fixture("traversal");
+    write(&dir.join("outside-secret.js"), r#"module.exports={};"#);
+    write(
+        &dir.join("entry.ts"),
+        r#"import "thirdparty/../../outside-secret";
+console.log("ESCAPED");
+"#,
+    );
+    let (stdout, stderr) = run(&dir, "entry.ts");
+    assert!(
+        !stdout.contains("ESCAPED"),
+        "a traversing subpath escaped the package root: {stdout}"
+    );
+    assert!(
+        stderr.contains("ERR_"),
+        "expected the traversal to stay unresolved, got: {stderr}"
+    );
+}
+
 /// THE ENCAPSULATION GUARD. A package that declares `exports` owns its subpath
 /// map; probing must never reach a path it withheld. Without this, #562's fix
 /// would be a sandbox escape out of every `exports` map on disk.
--- a/crates/nub-cli/tests/bare_subpath_resolution.rs
+++ b/crates/nub-cli/tests/bare_subpath_resolution.rs
@@ -65,6 +65,23 @@ fn fixture(name: &str) -> PathBuf {
     dir
 }
 
+/// `(major, minor)` of the `node` first on PATH. 22.15 is the fast-tier floor,
+/// and one test below is a fast-tier-only contract.
+fn path_node_version() -> Option<(u32, u32)> {
+    let out = Command::new("node").arg("--version").output().ok()?;
+    if !out.status.success() {
+        return None;
+    }
+    let v = String::from_utf8_lossy(&out.stdout);
+    let v = v.trim().trim_start_matches('v');
+    let mut parts = v.split('.');
+    Some((parts.next()?.parse().ok()?, parts.next()?.parse().ok()?))
+}
+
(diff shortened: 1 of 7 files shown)


colinhacks:

PR #600, opened 2026-07-29
fix(pm): stop nub writing aube-branded filesystem paths

colinhacks:
Nub created engine-branded paths in users' caches, stores, and `node_modules`. Two causes:

1. `embedder()` falls back to the aube profile when identity is unregistered, and only the PM preflight registered it — so `nub run` wrote `<cache>/aube/tools/node-gyp/lazy-bin/` and exported it as `npm_config_node_gyp`. `main` now registers identity first.
2. ~16 hardcoded on-disk names now compose from `aube_util::prog()`. Real tree: 31 `.aube-side-effects-cache`, 3 `pm/git/aube-git-*`.

`prog()` is `"aube"` by default, so standalone aube is byte-identical (aube-scripts 87/87).

Guards: `nub_never_writes_an_aube_branded_path`, `check-path-literals.sh`. fmt, clippy, lints, pm_identity 12/12 pass.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -218,6 +218,12 @@ jobs:
       # grep, no build — runs alongside clippy on the same run_rust gate.
       - name: Brand lint (no AUBE_* env reads)
         run: tests/brand-lint/check-env-reads.sh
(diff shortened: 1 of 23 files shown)


colinhacks:

PR #601, opened 2026-07-29
fix(linker): drop com.apple.quarantine from materialized native binaries

### LINKED ISSUE #575: CAS store should strip com.apple.quarantine after integrity verification
On macOS, `com.apple.quarantine` survives a copy. If a store entry acquires it, every `node_modules` linked from that entry inherits it, and `rm -rf node_modules` does not clear it because the store copy is the source.

Unnotarised `.node` addons then become permanently unloadable; a quarantined ad-hoc-signed binary is SIGKILLed on exec.

pnpm hit this: https://github.com/pnpm/pnpm/issues/11056

Fix: clear the attribute on store entries once integrity verification passes.

colinhacks:
Closes #575

On macOS, `com.apple.quarantine` reaches a materialized package two ways. The issue describes the first; the second is why the fix is not where the title says.

1. **Source propagation** — `clonefile(2)`/`copyfile(3)` carry the source's xattrs, so a quarantined CAS blob poisons every package materialized from it.
2. **Creation-time stamping** — a process that inherited quarantine flags (a terminal inside a Gatekeeper-enabled app) has the kernel stamp *every inode it creates*, whatever the source was.

Measured against a verifiably clean store blob under a quarantine-enabled process: the `clonefile` and `copyfile` targets still come out quarantined; only a hardlink escapes, having created no new inode. So stripping the **store entry** would fix only the hardlink strategy, and macOS defaults to `ReflinkAuto` (`clonefile`). Opting the process out isn't possible either — the flags are monotonic, and the path-exclusion SPI returns `EPERM`.

The strip therefore lands on the **materialized target**, and on cache hits as well as fresh materializations: the global virtual store is on by default and outlives `rm -rf node_modules`, so an entry written by a pre-fix build would otherwise stay quarantined forever — the exact state the issue reports.

### Notes

- **Filter is the union of the executable bit and `.node`/`.dylib`/`.so`.** Neither alone suffices: npm ships loadable modules mode 0644, while native CLI binaries (`bin/esbuild`, `biome`) are 0755 with no extension. Both halves catch a real file below.
- Uses `removexattr(2)` + `XATTR_NOFOLLOW`, not a spawned `xattr(1)`. Driven off the package index, so it also covers the whole-dir `clonefile` branch that fills a package without visiting files.
- Warm-path cost is one `removexattr` per guarded file: 19 files on a 20-package project, ~0.7ms against a ~229ms warm install.
- macOS-only. The filter and its test sit outside the macOS module so aube's ubuntu CI leg exercises them; the behavioural xattr tests are macOS-scoped and were verified locally.

### Not covered (documented in the module)

An install that relinks nothing (the frozen fast path) strips nothing; removing `node_modules` invalidates it. Files written into a package *after* linking — `node-gyp` output and side-effects-cache restores — are invisible to an index-driven pass. The side-effects cache is a second persistent cache with the same propagation shape and is worth its own issue.

### Verification

Differential against v0.2.10 on an isolated store.

| Scenario | pre-fix | fixed |
|---|---|---|
| A — quarantined store, clean installer | 3/4 quarantined | 0/3 |
| B — clean store, quarantine-enabled installer | 4/4 | 0/3 |
| C — poisoned virtual store, `rm -rf node_modules`, reinstall | 3/4 | 0/3 |

Per-file in A: `@esbuild/darwin-arm64/bin/esbuild` and `esbuild/bin/esbuild` (Mach-O executables) and `lightningcss.darwin-arm64.node` (Mach-O dylib) all go quarantined → clean. Plus 5 unit tests and `clippy --all-targets -D warnings` clean on both the macOS and non-macOS `cfg` paths.

### DIFF
--- a/vendor/aube/crates/aube-codes/src/warnings.rs
+++ b/vendor/aube/crates/aube-codes/src/warnings.rs
@@ -28,6 +28,7 @@ pub const WARN_AUBE_LTHASH_MISMATCH: &str = "WARN_AUBE_LTHASH_MISMATCH";
 pub const WARN_AUBE_DELTA_INVALIDATE_FAILED: &str = "WARN_AUBE_DELTA_INVALIDATE_FAILED";
 pub const WARN_AUBE_GVS_INCOMPATIBLE: &str = "WARN_AUBE_GVS_INCOMPATIBLE";
 pub const WARN_AUBE_GVS_MODE_CHANGED: &str = "WARN_AUBE_GVS_MODE_CHANGED";
+#[rustfmt::skip] pub const WARN_AUBE_QUARANTINE_STRIP_FAILED: &str = "WARN_AUBE_QUARANTINE_STRIP_FAILED";
 
 // ── settings / config validation ────────────────────────────────────
 pub const WARN_AUBE_INVALID_CONCURRENCY: &str = "WARN_AUBE_INVALID_CONCURRENCY";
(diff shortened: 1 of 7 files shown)


colinhacks:

PR #602, opened 2026-07-29
fix(resolver): fail closed when publish time can't prove a version's age

### LINKED ISSUE #581: minimumReleaseAge: missing publish-time data fails open, even in strict mode

colinhacks:
`minimumReleaseAge` treated a version with no `time:` entry as eligible, so a packument without publish times silently disabled the gate, even under `minimumReleaseAgeStrict=true`.

Repro: with a 10-year cutoff and strict on, adding `registry-supports-time-field=true` to `.npmrc` flipped `nub add strip-ansi-cjs` from `ERR_NUB_NO_MATURE_MATCHING_VERSION` to a clean install.

A missing time is now resolved against the packument's `modified` — an upper bound proving every version in the document is mature, which keeps abbreviated metadata usable without a full fetch. When even that can't settle it, `strict` decides: strict blocks, lenient stays eligible as before, so standalone aube's default is preserved. The `modified` bound is pnpm 11's mechanism (`pickPackageFromMeta`); the terminal default differs, since pnpm warns-and-skips where nub blocks.

The vulnerability re-pick ranks in two tiers rather than vetoing: an undated safe version still beats returning a known-vulnerable fallback, but a dated-and-too-new one stays excluded.

Also gives `registrySupportsTimeField` a `workspaceYaml` source and field — pnpm reads it only from `pnpm-workspace.yaml`.

BEHAVIOR CHANGE: under strict, a registry serving no `time` map blocks packages whose `modified` is newer than the cutoff. Escapes: `minimumReleaseAgeExclude`, `minimumReleaseAge=0`.

Overlaps #587, which carries the same `strict` gate without the `modified` bound.

Closes #581

### DIFF
--- a/vendor/aube/crates/aube-manifest/src/workspace/config.rs
+++ b/vendor/aube/crates/aube-manifest/src/workspace/config.rs
@@ -544,6 +544,14 @@ pub struct WorkspaceConfig {
     #[serde(default)]
     pub minimum_release_age_strict: Option<bool>,
 
+    /// Whether the configured registry returns per-version `time:` data in
+    /// its abbreviated (corgi) packument, letting the resolver skip the
+    /// extra full-packument fetch when a publish-age cutoff is active.
+    /// pnpm reads this from `pnpm-workspace.yaml` only, so the yaml is the
+    /// source that matters for mirroring a pnpm incumbent.
+    #[serde(default)]
+    pub registry_supports_time_field: Option<bool>,
+
     /// OSV `MAL-*` advisory check policy for `aube add`. One of
     /// `"on"` (default, fail open on fetch error), `"required"` (fail
     /// closed on fetch error), or `"off"`.
--- a/vendor/aube/crates/aube-resolver/src/resolve/vulnerable.rs
+++ b/vendor/aube/crates/aube-resolver/src/resolve/vulnerable.rs
@@ -41,18 +41,18 @@ pub(super) fn prefer_non_vulnerable_pick<'a>(
     // waves a version past the minimumReleaseAge gate here too (still
     // subject to `exempt_cutoff`, the time-based wall), otherwise the
     // re-pick could discard the exempt safe version and keep the
-    // vulnerable one.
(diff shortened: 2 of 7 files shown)


colinhacks:

PR #603, opened 2026-07-29
fix(linker): converge directory-link creation on a re-link

### LINKED ISSUE #566: [Bug Report] Fail to remove dependencies

### LINKED ISSUE #576: nub add fails on Windows with "Cannot create a file when that file already exists (os error 183)"

colinhacks:
Closes #576
Closes #566

`create_dir_link` failed on any occupied link path, so re-linking a live `node_modules` aborted the install. On Windows `junction::create` begins with `fs::create_dir`, so even an identical junction returns 183. It now no-ops when the entry already points at the target, else clears the slot and retries once. The clear is non-recursive: a populated real directory stays a loud error, since that shape means the linker is about to overwrite content it does not own.

Guards that should have caught it were also wrong on Windows: `read_link` on a real dir gives raw 4390, not `InvalidInput`, and two removals used `remove_dir().or_else(remove_file())`, which cannot evict a populated dir. The second also caused `os error 5` installing over an npm tree. Classification is now one `lstat` — a junction's reparse tag already makes std report it as a symlink, not a directory.

Behavior change: `detect_aube_dir_gvs_mode` can now return `Some(false)` on Windows, arming the `node_modules` wipe on a store-locality flip. Intended, matches Unix, warns.

8 tests. Differentials: reverting `create_dir_link` fails the convergence cases; a recursive clear fails the refusal case. CI now runs the vendored engine tests — `vendor/aube` is excluded from the root workspace, so they never ran anywhere.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -780,13 +780,21 @@ jobs:
       # linker included. That gap is why the Windows-only `read_link` /
       # `junction::create` divergences behind #576 and #566 went unnoticed: the
       # semantics they depend on differ per-OS and nothing exercised them off Unix.
-      # Run the linker's platform-semantics suite from inside the aube workspace on
-      # every OS leg. Scoped to `-p aube-linker --lib` deliberately: a light
-      # dependency subtree, no network or registry fixture, and it covers the
-      # link-creation and entry-classification primitives every install rides on.
+      # Run the engine's platform-semantics suites from inside the aube workspace.
+      # Both crates are needed: the link-creation and entry-classification
+      # primitives live in `aube-linker`, but `detect_aube_dir_gvs_mode`'s tests
+      # live in `aube` — and those are the ones that were `#[cfg(all(test, unix))]`.
+      # `--lib` keeps it to unit tests: no network, no registry fixture.
+      #
+      # Linux is excluded on purpose. These are Node-independent Rust unit tests,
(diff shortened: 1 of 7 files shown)


colinhacks:
Superseded by #595, which fixed #576 and #566 with the same diagnosis. The one part not covered there — `nub install` aborting on Windows over an incumbent npm/yarn `node_modules` — is extracted to #613.

PR #604, opened 2026-07-29
fix(resolver): skip optional deps whose version pick finds no match

### LINKED ISSUE #580: Attempts to resolve @rolldown/binding-darwin-arm64@1.0.3 on Linux
OS: Linux x64
Node: 22.23.1
nub: 0.6.0

package.json:
```json
{
  "devDependencies": {
    "rolldown": "1.0.3"
  }
}
```

Expected:
- install @rolldown/binding-linux-x64-gnu@1.0.3

Actual:
- tries to resolve @rolldown/binding-darwin-arm64@1.0.3
- fails with ERR_NUB_NO_MATCHING_VERSION

colinhacks:
`rolldown` declares 15 `@rolldown/binding-*` packages as `optionalDependencies`. The version pick runs before the `os`/`cpu` filter, so a binding the host will never install could fail the entire install when the registry did not carry it. An optional dependency should never be fatal — the fetch-failure and platform-mismatch paths already skip, but the version-pick path did not.

Optional deps whose range matches no version are now skipped and recorded, with a warning. The age-gate path stays fatal: a version withheld by `minimumReleaseAge` is a supply-chain signal the user opted into, not a package the registry lacks.

## Verification

Differential against a registry with `1.0.3` removed from one wrong-platform binding, fresh cache per leg:

| | released 0.6.0 | this branch |
| --- | --- | --- |
| registry unmodified | exit 0 | exit 0 |
| version removed | `ERR_NUB_NO_MATCHING_VERSION`, exit 20 | exit 0 |

`require('rolldown')` loads after the install, so the tree is usable rather than silently incomplete. npm and pnpm both install cleanly in the same scenario.

New tests cover both directions — an optional dep skips, a required dep still fails. The optional test was confirmed to fail without this change.

Closes #580

### DIFF
--- a/vendor/aube/crates/aube-codes/src/warnings.rs
+++ b/vendor/aube/crates/aube-codes/src/warnings.rs
@@ -86,6 +86,7 @@ pub const WARN_AUBE_INVALID_CLIENT_CERT: &str = "WARN_AUBE_INVALID_CLIENT_CERT";
 
 // ── resolver ────────────────────────────────────────────────────────
 pub const WARN_AUBE_UNSUPPORTED_PLATFORM_INSTALL: &str = "WARN_AUBE_UNSUPPORTED_PLATFORM_INSTALL";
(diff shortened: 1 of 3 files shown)


colinhacks:

PR #607, opened 2026-07-29
fix(resolver): fetch publish times on the paths that resolve without the full-packument cache

colinhacks:
Completes the `minimumReleaseAge` fix. #602 made an undeterminable publish age fail closed at the pick site; this lands the other half — making sure the publish times are actually fetched — plus the diagnostic and docs the new fail-closed path needs.

## Why both halves are required

`update` / `add` / `dedupe` / `audit` resolve without the full-packument disk cache (the dist-tag freshness rule, `settings_context.rs:452`), so the `needs_time` fallback fetched npmjs's abbreviated document, which carries no `time` map. Before #602 that silently disabled the gate. **After #602 it hard-fails instead**: no version can be dated, npmjs's corgi `modified` is recent for any actively published package, so every candidate is blocked and strict mode aborts.

Verified on the merged tree — the regression test in strict mode, which is Nub's shipped default:

```
AgeGate(AgeGateDetails { name: "foo", range: "^1.0.0", minutes: 1440,
                         gated: ["1.1.0", "1.0.0"] })
```

Both versions gated, including one published in 2024. Fetching the full document fixes it.

## Contents

- **The fetch fix**, authored by @afonsojramos — `RegistryClient::fetch_packument_with_time`, the uncached full-packument sibling, wired into all four `needs_time` fallbacks. Carried over from #582 with authorship intact; that PR predates #602 and can't merge as-is against it.
- **The regression test, in strict mode.** It ran lenient, where the lowest-satisfying fallback lands on the mature version regardless of the wall — so it would have passed with the gate disabled. The corgi fixture also cleared `modified`, which npmjs does not (verified against registry.npmjs.org: abbreviated metadata carries `modified`, never `time`), and for an active package that timestamp is recent — so it must not stand in as the maturity proof a time-less document otherwise gets.
- **An honest error.** A registry serving no publish dates now blocks the whole range, and the age-gate wording reported that as an ordinary cutoff miss — listing years-old versions as "blocked by age gate" and offering a wider window as the remedy, which would never have helped. Split the headline and help on whether any satisfying version carried a date at all. One dated version keeps it an ordinary age gate.
- **Docs.** The resolution-level window had no page of its own — only the build-script floor that consumes it — so dlx and the runner pointed at the floor for a posture it doesn't describe. Also corrects the `trustPolicyExclude[]=` example: the npmrc reader compares keys verbatim against the settings registry, which carries `trustPolicyExclude` and `trust-policy-exclude`, so the bracket form was silently ignored.

## Verification

290 resolver tests pass. Root `cargo fmt --check` clean and `cargo clippy --all-targets --all-features -- -D warnings` clean. Vendored aube carries 36 pre-existing rustfmt diffs on `main`; this branch has 35 — it fixes one and adds none.

Refs #581. Supersedes #582.

### DIFF
--- a/site/content/docs/install/index.mdx
+++ b/site/content/docs/install/index.mdx
@@ -412,7 +412,7 @@ Which manifest field grants permission tracks the inferred incumbent: pnpm proje
 
 A registry-resolved version must be older than `minimumReleaseAge` — 24 hours by default — before Nub will install it. The window is what keeps a compromised publish out of your tree during the hours between it going up and being caught.
 
-Publish dates come from the registry's `time` metadata, so the gate is only as strong as what the registry serves. An age Nub cannot establish counts as a failure, not a pass:
+Publish dates come from the registry's `time` metadata, so the gate is only as strong as what the registry serves. Under the default `minimumReleaseAgeStrict=true`, an age Nub cannot establish counts as a failure rather than a pass:
 
 | Registry metadata | Outcome |
 |---|---|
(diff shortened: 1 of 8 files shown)


colinhacks:

PR #608, opened 2026-07-29
fix(install): clear quarantine on side-effects-cache restores and self-upgrades

colinhacks:
Refs #575

Five macOS surfaces the linker fix in #601 could not reach. All write files that are then loaded or executed, and none goes through a package index.

### Side-effects cache — both directions

`restore_if_available` copies a cached build tree into an already-linked package. A hardlink shares the cache entry's inode and therefore its xattrs; the `fs::copy` fallback is `fcopyfile` with `COPYFILE_ALL`, which carries xattrs *and* mints a new inode that a quarantine-enabled process gets stamped regardless of the source. The payload is locally compiled `node-gyp` output — ad-hoc signed at best, so Gatekeeper refuses it. One poisoned cache entry re-poisons every future restore.

The strip runs after the restore, not when the entry is saved: the new-inode stamping happens however clean the source was. Same reason the linker fix landed on the target rather than the store.

The **miss** path needs it too, and matters more: `sideEffectsCache` defaults on, so that is the first install of any package with a native build. `node-gyp` writes `build/Release/*.node` in place, in a child that inherited this process's quarantine flags. Stripping only on restore would leave the install that *built* the addon shipping it unusable. The strip runs after the dep-hook loop, off the async worker via `spawn_blocking` like every other filesystem step there.

Note this does **not** make the saved cache entry clean: `save` copies with `CopyMode::Copy`, so `fs::copy` mints inodes the kernel stamps again regardless of the now-clean source. The restore-side strip stays load-bearing — the two cover different halves rather than one making the other redundant.

### nub's own embedded runtime addon

`addons/nub-native.node` is ad-hoc signed (`codesign`: `adhoc, linker-signed`, no Team ID), inflated in-process from the in-binary blob, and `dlopen`ed by `transform-core.mjs` — which swallows a load failure into a null handle and then calls it unguarded. A quarantined addon is therefore a hard failure for every TypeScript transpile, not a degraded path.

This one strips on the warm path as well as a fresh extract. `VERIFIED_ENTRYPOINTS` compares BLAKE3 content hashes, and an xattr does not change a file's bytes, so a poisoned dir verifies clean forever and the self-heal re-extract never fires. Costs one syscall per process, behind the existing `OnceLock`.

### `nub upgrade`

Released nub binaries are ad-hoc signed — `codesign` reports `adhoc, linker-signed`, no Team ID — and macOS kills a quarantined ad-hoc-signed executable on exec. `curl` and `tar` run as children of nub, so an upgrade from a terminal that inherited quarantine flags installs a nub that cannot start. The same seam also covers the self-shim's provision-then-exec path, whose binary has identical provenance. Both archives are SHA-256 verified before use.

### Not included

Node dist provisioning. Official Node is Developer ID signed with a hardened runtime, so Gatekeeper verifies it rather than blocking, and the other extractor in that crate handles JS bundles. No failure was demonstrable, so it stays unchanged rather than shipping speculative hardening.

### Notes

- Adds walk-driven `strip_quarantine_from_tree` for callers with no index; uses `symlink_metadata` and never descends a symlink.
- The shared filter now takes an `executable` bool so the index- and walk-driven paths cannot drift apart on the union rule.
- `ensure_bin_executable`'s doc updated — it now has two jobs, both prerequisites for running a freshly downloaded binary.
- The syscall wrapper lives in `nub-core::quarantine`, shared by the upgrade path and the runtime cache rather than copied; its tests use the crate's existing temp-dir convention instead of adding a dev-dependency.
- The upgrade warning said `nub upgrade:` on a seam that also serves the self-shim, so it now says `nub:`.

### Verification

Local (re-run after the follow-up commit): aube clippy (`aube-linker` + `aube`, `--all-targets -D warnings`) clean on both the macOS and non-macOS `cfg` paths; full linker suite 141 passed; nub workspace clippy `--all-targets --all-features` clean; new `nub-core::quarantine` tests pass; fmt clean. Two new tests, each with a verified negative control — making the walk follow symlinks fails only the escape test, and dropping the filter's executable arm fails only the tests that depend on it. The #601 install e2e re-run against this build shows no regression (0/3 cold, 0/3 quarantine-enabled installer, and the poisoned-store heal still 3/3 → 0/3).

Not verified end-to-end: a live `nub upgrade` (it would download a real release and swap `~/.nub`), and the Gatekeeper kill itself (that needs executing a quarantined binary). The signing status was verified directly with `codesign`.

### DIFF
--- a/crates/nub-cli/src/cli.rs
+++ b/crates/nub-cli/src/cli.rs
@@ -6352,55 +6352,31 @@ fn ensure_bin_executable(path: &Path) -> Result<()> {
     Ok(())
 }
 
-/// Drop `com.apple.quarantine` from the freshly swapped-in binary.
+/// Drop `com.apple.quarantine` from a binary nub just downloaded.
 ///
 /// Released nub binaries are ad-hoc signed — `codesign` reports
 /// `adhoc, linker-signed` with no Team ID — and macOS kills a quarantined
 /// ad-hoc-signed executable on exec rather than merely warning about it.
-/// `curl` and `tar` above run as children of this process, so when nub is
-/// invoked from a terminal that inherited quarantine flags (one embedded
-/// in a Gatekeeper-enabled app), everything they write is stamped and the
-/// upgrade would install a nub that cannot start. The archive's SHA-256 is
-/// checked against the published checksum before the swap, so clearing the
-/// attribute afterwards is the same trade Homebrew makes for a verified
-/// download.
+/// `curl` and `tar` run as children of this process, so when nub is invoked
+/// from a terminal that inherited quarantine flags (one embedded in a
+/// Gatekeeper-enabled app), everything they write is stamped: the upgrade
+/// would install a nub that cannot start, and the self-shim would exec one.
+/// Both archives are checksum-verified before use, so clearing the attribute
+/// afterwards is the trade Homebrew makes for a verified download.
 ///
 /// Best-effort by the resilience contract in `perform_selfowned_upgrade`:
-/// the binary is already swapped and executable, so a failure here warns
-/// with the manual fix rather than failing the upgrade.
-#[cfg(target_os = "macos")]
+/// the binary is already swapped and executable, so a failure warns with the
+/// manual fix rather than aborting.
+#[cfg(not(windows))]
 fn drop_quarantine(path: &Path) {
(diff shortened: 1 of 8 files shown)


colinhacks:

PR #609, opened 2026-07-29
fix(lockfile): drop the redundant npm-alias target left by the pnpm reader

### LINKED ISSUE #578: dedupe repeatedly reports npm alias targets removed without changing pnpm-lock.yaml
## Summary

   In pnpm compatibility mode, an `npm:` package alias causes `nub dedupe` to repeatedly report the
 alias target as removed even though `pnpm-lock.yaml` remains byte-identical.

   As a result, `nub dedupe --check` permanently exits with code 1.

   ## Environment

   - Nub: `v0.6.0`
   - Also reproduced with: `v0.6.0-canary.20260727.102`
   - Node: `v24.18.0`
   - Platform: macOS 26.5.2 arm64
   - Lockfile: `pnpm-lock.yaml`, version 9

   ## Minimal reproduction

   Create `package.json`:

   ```json
   {
     "name": "nub-dedupe-alias-repro",
     "private": true,
     "packageManager": "pnpm@11.17.0",
     "dependencies": {
       "number-alias": "npm:is-number@7.0.0"
     }
   }
 ```

 Then run:

 ```sh
nub install --reporter=append-only

shasum -a 256 pnpm-lock.yaml
nub dedupe --reporter=append-only
shasum -a 256 pnpm-lock.yaml

nub dedupe --check --reporter=append-only
echo $?
 ```

 The lockfile in this reproduction is generated by Nub; the pnpm CLI is not involved.

 Actual result

 nub dedupe reports:

 ```text
- is-number@7.0.0
Dedupe: 1 removed, 0 added (net -1 packages)
Wrote pnpm-lock.yaml
 ```

 However, the SHA-256 is identical before and after:

 ```text
d64cf23037a59104db36be9d2786b5561da832a5eedafc51f8febb1fb93208bd
 ```

 A subsequent check reports the same removal and exits with code 1:

 ```text
- is-number@7.0.0
Dedupe: 1 removed, 0 added (net -1 packages)
× dedupe --check: lockfile is not deduped
 ```

 Expected result

 Because is-number@7.0.0 is the required target of the number-alias dependency:

 - It should not be reported as removed.
 - nub dedupe should report that the lockfile is already deduped.
 - nub dedupe --check should exit with code 0.

 Additional observations

 - Replacing the alias with a direct dependency makes the check pass.
 - The issue appears specific to pnpm-lock.yaml; Nub's native package.lock mode does not reproduce it.
 - This looks similar to #494, but for serialized registry alias targets rather than link: packages.

 The parsed lockfile graph appears to contain the target key (is-number@7.0.0), while the freshly
 resolved graph used by dedupe does not. diff_graphs therefore reports a phantom removal, but
 finalization/install preserves the required target and leaves the lockfile unchanged.

colinhacks:
Closes #578

The pnpm reader kept the canonical real-name entry alongside the alias-keyed clone it synthesizes for an `npm:` alias. A fresh resolve emits only the clone, so consumers walking `packages` raw disagreed with a re-resolved graph: `dedupe` reported a phantom removal every run (`--check` red forever on a byte-identical lockfile), `fetch` over-counted, `rebuild <real-name>` matched a non-installed dep. Usually hit transitively, via `glob@10` reaching `@isaacs/cliui`.

The reader now drops an alias-clone source nothing else reaches; shared targets survive, `file:` targets are excluded.

Also fixed, each blind to aliased packages: `ignored-builds`/`approve-builds` (a prerequisite here), `audit`, `outdated`.

Conformance green against real pnpm 10.15.1 and npm 11.17.0.

### DIFF
--- a/crates/nub-cli/tests/pm_verbs.rs
+++ b/crates/nub-cli/tests/pm_verbs.rs
@@ -645,6 +645,67 @@ fn dedupe_ignores_workspace_links_and_check_passes() {
     );
 }
 
+/// An `npm:` alias must not read as a dedupe change (#578). The pnpm
+/// reader synthesizes an alias-keyed package and used to keep the
+/// canonical real-name entry too, while a fresh resolve emits only the
+/// clone — so `dedupe` reported the target as removed on every run and
+/// `--check` failed forever on a byte-identical lockfile. Needs the
+/// registry: dedupe's whole job is a fresh resolve.
+#[test]
+#[ignore = "network: resolves + fetches is-number@7.0.0 from the npm registry"]
+fn dedupe_ignores_npm_alias_targets_and_check_passes() {
+    if !registry_reachable() {
+        eprintln!("skipping: registry.npmjs.org unreachable");
+        return;
+    }
+    let dir = pm_tmpdir("dedupe-alias");
+    // `packageManager` selects pnpm-lock.yaml; the bug is specific to that
+    // format, since only its writer re-keys an alias under the real name.
+    std::fs::write(
+        dir.join("package.json"),
+        r#"{"name":"fixture","private":true,"packageManager":"pnpm@11.17.0","dependencies":{"number-alias":"npm:is-number@7.0.0"}}"#,
+    )
+    .unwrap();
+    let (data, cache) = (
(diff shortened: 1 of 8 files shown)


colinhacks:

PR #612, opened 2026-07-29
fix(resolver): drop primer validators on the abbreviated packument cache

colinhacks:
Two independent fixes, both found while investigating #580.

## Primer validators on the abbreviated packument cache

The bundled primer is a truncated slice of a packument (newest `version_cap` versions) but carries the registry's real validators for the *complete* document. The full cache tier deliberately drops them so a later range-miss refetch is an honest unconditional GET. The abbreviated tier still wrote them, so a revalidation of that entry is conditional, the registry answers `304`, and the truncated body is restored as authoritative — and because a disk-cache hit is not flagged primer-seeded, the driver's range-miss heal never fires. A version the registry plainly serves then fails with `ERR_AUBE_NO_MATCHING_VERSION`.

The abbreviated tier is in play whenever `needs_time` is false, which is where `registry-supports-time-field=true` lands. The regression test pins that configuration, drives two resolves against one cache dir, and fails without the fix — the failure reports the conditional request and the 304 that caused it.

## Stale `nub-dev` symlink

`make install-dev` hardcoded `$(CURDIR)/target` for the `nub-dev` / `nubx-dev` symlinks, but every documented rebuild path goes through `scripts/rust-build.sh`, which normally writes to the shared target dir. Nothing refreshed `./target/fast/nub`, so `nub-dev` kept serving whatever was built the last time `make` ran — days stale, and it made a merged fix look broken during #580.

`rust-build.sh` gains `--print-target`, which re-resolves the same shared/isolated decision a real build makes, and `install-dev` symlinks to that. Verified in both modes: shared for a clean tree, isolated for a worktree that diverges a depended-on crate.

## Verification

- New test fails without the cache fix, passes with it.
- `aube-resolver`: 291 passed, 0 failed.
- `cargo fmt --check` and `cargo clippy --all-targets --all-features -- -D warnings`: clean.

Refs #580

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,12 +51,19 @@ addon:
 # The fast profile rebuilds it in a fraction of that. addon-fast builds the
 # native addon under the same profile so a single `cargo build --profile fast`
 # pass serves both.
+# Builds through scripts/rust-build.sh and symlinks to the target dir IT picked.
+# Hardcoding $(CURDIR)/target left nub-dev pointing at a stale binary: every
+# documented rebuild path (the post-merge refresh, the dev-loop) goes through the
+# wrapper, which normally writes to the SHARED dir, so nothing updated
+# ./target/fast/nub and `nub-dev` silently served whatever was built last time
+# make ran. `--print-target` re-resolves the same shared/isolated decision.
 install-dev: addon-fast qos-global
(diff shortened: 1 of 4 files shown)


colinhacks:

PR #613, opened 2026-07-29
fix(linker): reclaim an incumbent node_modules tree on Windows

colinhacks:
`nub install` aborted on Windows in any project whose `node_modules` npm or yarn had already written:

```
failed to link node_modules
I/O error at D:\proj\node_modules\express: Access is denied. (os error 5)
```

`reconcile_top_level_link` clears a stale top-level entry with `remove_dir` falling back to `remove_file`. Those cover a junction and a plain file, but an incumbent package manager leaves a populated real directory, which is neither — `remove_dir` answers `ERROR_DIR_NOT_EMPTY`, `remove_file` then answers `ERROR_ACCESS_DENIED`. That is os 5, which `is_transient_fs_error` counts as transient, so it also burns the full ~10s retry ladder first.

Reclaiming that tree is the linker's job, and the Unix branch has always recursed for it — which is why this is Windows-only. The recursion is gated on `is_real_dir` so it reaches that shape and never a junction, whose target must survive, and it sits inside the existing ladder rather than adding a second.

Reproduced on 0.6.0 and canary on a windows-latest runner.

Note the incumbent-tree test exercises the `cfg(not(windows))` branch on Unix, where it passes either way; the windows leg of `aube-parity` is what makes it meaningful. The test says so in place.

### DIFF
--- a/vendor/aube/crates/aube-linker/src/link.rs
+++ b/vendor/aube/crates/aube-linker/src/link.rs
@@ -1862,7 +1862,10 @@ impl Linker {
 /// symlink, Windows normalizes to an absolute path before calling
 /// `junction::create`. A plain `read_link == expected` check that
 /// works on Unix would miss every warm run on Windows.
-fn reconcile_top_level_link(link_path: &Path, expected_target: &Path) -> Result<bool, Error> {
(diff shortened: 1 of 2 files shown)


colinhacks:

PR #614, opened 2026-07-29
fix(build): resolve CARGO_MANIFEST_DIR at run time, not compile time

colinhacks:
Three build scripts resolved their inputs through `env!("CARGO_MANIFEST_DIR")`, which bakes an absolute path into the compiled build-script binary. Cargo caches those per target dir, and this repo shares one target dir across worktrees — so the path stays pinned to whichever worktree compiled it first.

While both worktrees exist, a build silently reads the other tree's files. Once that worktree is deleted, every tree sharing the cache fails on a path absent from the current checkout — a build error on correct source, naming a directory the developer has never seen.

Both modes were live in the shared cache: an `aube-settings` script pinned to a deleted worktree, and two `nub-cli` scripts pinned to different live worktrees. `nub-cli` is the worst — it bakes `site/content/docs`, which feeds `nub agent skill`. `nub-native` stamps the git SHA into `NUB_NATIVE_BUILD_ID`, the transpile-cache key.

Cargo sets `CARGO_MANIFEST_DIR` in the build script's environment at run time, so the runtime lookup resolves per invocation. `cargo publish --verify` is unaffected — it still resolves inside `target/package/<crate>-<ver>/`. `nub-core` and `aube-resolver` already did this.

Recovery for an already-poisoned cache: `rm -rf <target>/*/build/{aube-settings,nub-cli,nub-native}-*`. Note `cargo clean -p aube-settings` from the repo root is a silent no-op — it reports "Removed 0 files" and exits 0, because the aube crates are not root workspace members.

Verified: `rustfmt` clean; `nub-cli/build.rs` type-checks standalone under rustc; the `String`-to-path conversions type-check. A full `cargo build` was OOM-killed four times by concurrent load on the dev host, so compilation of all three is left to CI.


colinhacks:

PR #615, opened 2026-07-29
test(aube): make the manifest-root namespace tests hermetic under parallel execution

colinhacks:
The three `set_allow_builds_*` tests fail when the crate's tests run in parallel, and pass under `--test-threads=1`.

Cause: the surface gates (`read_branded_pnpm_config`, `read_manifest_root_config`) live on the process-global `EngineContext`, and the code under test re-reads them mid-call. A sibling's `update_engine_context` lands in between, so the wrong branch is taken.

Each posture-sensitive test now claims the gates through a guard that serializes, sets both gates explicitly, and restores the default on drop — the shape `pm_engine::ENGINE_GLOBAL_LOCK` already uses. Test file only.

Running the test binary across `--test-threads` 1..8: 83/100 failures before, 0/300 after.

Refs #602

### DIFF
--- a/vendor/aube/crates/aube-manifest/tests/root_namespace_write.rs
+++ b/vendor/aube/crates/aube-manifest/tests/root_namespace_write.rs
@@ -12,10 +12,12 @@
 //! root and whose read side gates the `pnpm` namespace off — so a nested
 //! `pnpm.*` write would be orphaned.
 
+use std::sync::{Mutex, MutexGuard, PoisonError};
+
 use aube_manifest::{
     AllowBuildRaw, PackageJson, workspace::edit_setting_map, workspace::set_allow_builds,
 };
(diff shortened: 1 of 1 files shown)


colinhacks:
Closing unmerged. The flake this fixes cannot reach CI: `vendor/aube/.cargo/config.toml` pins `RUST_TEST_THREADS=1` workspace-wide, and the Aube parity legs run with `working-directory: vendor/aube`, so they inherit it. That config's own comment records choosing the pin *over* the per-test mutex machinery this PR adds.

The real defect was the invocation. Cargo discovers config from the CWD, not from `--manifest-path`, so running the suite from the repo root silently drops the pin and runs those tests in parallel — which is how this surfaced. Documented in the `dev-loop` and `rust-build` skills (f966b973ad) instead.

PR #616, opened 2026-07-29
fix(linker): stop wiping unchanged packages on every hoisted relink

### LINKED ISSUE #610: Hoisted linker wipes unchanged package directories on every relink
Under `node-linker=hoisted`, every relink wipes and refills each package directory, including packages whose content is unchanged (`hoisted.rs:781` — `try_remove_entry`, then refill from the store index, which holds tarball contents only).

Confirmed: a file written into `node_modules/accepts/build/` does not survive an unrelated `nub add`.

Not confirmed: whether a package with postinstall build output is then rebuilt. `lifecycle_delta_filter` selects only packages whose content hash changed and skips before the side-effects-cache restore, which suggests not — but I could not reproduce that end to end, since build scripts are deny-by-default and my fixture's postinstall never ran. Settling it needs a native-dep fixture with builds approved.

Isolated mode is unaffected.

colinhacks:
Hoisted relinks wiped and refilled every placed package, and the refill restores published tarball contents — so an unrelated `nub add` deleted whatever a postinstall produced (`build/Release/*.node`, a downloaded binary) and never put it back. The delta filter then skipped those packages because their fingerprint was unchanged, so they were neither rebuilt nor restored.

npm and pnpm — including pnpm's hoisted layout — leave an unchanged package directory alone. This does the same, vouching for a package only when the previous link COMPLETED, the layout matches, and the fingerprint is unchanged. Completion is tracked by a sentinel, not the install state: state written by a successful install outlives a later one that died mid-link, so state alone would vouch for a half-written tree.

Faster than what it replaces: ~720ms vs ~920ms per incremental add. Isolated is untouched.

Closes #610

### DIFF
--- a/vendor/aube/crates/aube-linker/src/builder.rs
+++ b/vendor/aube/crates/aube-linker/src/builder.rs
@@ -44,7 +44,22 @@ impl Linker {
             no_integrity_read_keys: std::collections::BTreeMap::new(),
             link_progress: None,
             disk_materialize: std::collections::HashSet::new(),
+            reusable_hoisted: std::collections::HashSet::new(),
         }
+    }
+
+    /// Vouch that these dep paths are already correctly and completely placed
+    /// on disk, letting the hoisted pass skip its wipe-and-refill for them.
+    ///
+    /// The caller owns this judgement (see the field docs on [`Linker`]): it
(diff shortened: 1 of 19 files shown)


colinhacks:

PR #617, opened 2026-07-29
test(init): close the mock registry connection gracefully

colinhacks:
The age-floor install test failed on Windows CI with a transport error, not an HTTP status, fetching the tarball — the largest body served.

Cause: the fixture read one 4096-byte chunk of the request, wrote the response, and dropped the socket. Closing a socket that still holds unread bytes is an abortive close: it discards the send buffer, so a client mid-read never gets the rest.

The handler now reads the request to its end, half-closes, drains to EOF, and forces blocking mode on the accepted socket.

A probe on branch `probe/win-socket-close` (4 MiB body, 15 requests, run 30473314496) lost every response on Windows, Ubuntu and macOS with the old close, none with the new.

Refs #602

### DIFF
--- a/crates/nub-cli/tests/init_cmd.rs
+++ b/crates/nub-cli/tests/init_cmd.rs
@@ -205,18 +205,21 @@ impl RegistryFixture {
 
 /// Serve one request, then close the connection *gracefully*.
 ///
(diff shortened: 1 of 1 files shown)


colinhacks:

PR #618, opened 2026-07-29
fix(linker): clear a stale workspace store entry recursively

colinhacks:
`link_all` and `link_workspace` carry near-duplicate step-1 GVS-populate loops. #595 fixed the `EntryState::Stale` removal in the first — a non-recursive `remove_dir` cannot clear a populated real directory, and the swallowed error resurfaces as EEXIST / os-183 from the link creation right below — and left the identical arm in the workspace loop untouched. A workspace therefore wedged permanently where a single-package project self-healed:

```
× failed to link workspace node_modules
╰─▶ I/O error at C:\ws\node_modules/.store\express@4.18.2: Cannot
    create a file when that file already exists. (os error 183)
```

`nub install`, `nub install --force`, and member-level `nub add` / `nub remove` all failed the same way; only deleting `node_modules` recovered. Confirmed on Windows Server 2022 against `0.6.0-canary.20260729.129`, with a single-package control that recovered under the same perturbation.

Not Windows-specific despite where it surfaced — with this arm reverted the new test fails on macOS too, with EEXIST (os 17).

The trigger is a mixed store (some entries links, one a real directory). `detect_aube_dir_gvs_mode` returns `Some(true)` on the first link it sees, so the mode-change wipe does not fire and the survivor reaches this arm. That wipe is currently the only thing pre-empting the failure and it is a different layer, so this is latent rather than unreachable: a partial cache restore, a crash, or external tooling produce the same shape.

### DIFF
--- a/vendor/aube/crates/aube-linker/src/link.rs
+++ b/vendor/aube/crates/aube-linker/src/link.rs
@@ -1131,9 +1131,15 @@ impl Linker {
                                 nested_link_targets.as_ref(),
                             )?;
 
+                            // Same `Stale` shapes as `link_all`'s step 1, and
+                            // the same reason a non-recursive `remove_dir`
+                            // cannot clear them — see the comment there. This
+                            // arm is the workspace twin of that code and was
+                            // left on the old removal, so a workspace wedged
+                            // permanently on os-183 where a single-package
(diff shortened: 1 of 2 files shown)


colinhacks:

PR #619, opened 2026-07-29
test(watch): close the watcher-registration race in the env reload waits

colinhacks:
`watch_reloads_explicit_env_file_across_restarts` failed 7 times in 200 runs on main, always the same way: only the startup snapshot appeared and the 15s wait ran out without a restart.

The race is in the test. Node's `--watch` supervisor builds its watch set asynchronously, and an `--env-file` is consumed before user code runs — so observing the first snapshot proves the child started, not that the env file is being watched yet. The single rewrite can land in that gap, be missed, and then nothing ever restarts.

Both restart tests now re-apply the write while polling, via a `wait_for_watch_reload` helper. That drops the dependence on winning the registration race without weakening anything: the restarted child must still report the new values for the callers' assertions to hold. Waits for a *first* snapshot have no such race and keep using `wait_for_watch_snapshot`.

Same binary, same protocol: **7/200 before, 0/200 after**.

Note for anyone reproducing this by running the test binary directly rather than through cargo — export `CARGO_MANIFEST_DIR=<worktree>/crates/nub-cli` first, or every run aborts in a helper before reaching the watch logic and looks like a 100% failure.

### DIFF
--- a/crates/nub-cli/tests/integration.rs
+++ b/crates/nub-cli/tests/integration.rs
@@ -3988,6 +3988,55 @@ fn finish_watch_probe(child: &mut std::process::Child) {
     let _ = child.wait();
 }
 
+/// Wait for a RESTART to produce `needle`, re-applying `poke` as it polls.
+///
+/// Node's `--watch` supervisor builds its watch set asynchronously, and an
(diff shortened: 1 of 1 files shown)


colinhacks:

PR #620, opened 2026-07-29
fix(watch): stop nub watch leaking an immortal node --watch supervisor

colinhacks:
`run_watch` was the only long-lived spawn in nub that called `Command::status()` directly. Every other one — `spawn.rs:672`, `spawn.rs:1182`, `cli.rs:4647` — goes through a process-group + reaper path.

That single gap is why orphan sweeps on the dev host kept finding stranded processes:

```
rustc:  13 total,   0 orphaned
cargo:   3 total,   0 orphaned
node:  212 total,  99 running --watch, ALL with PPID=1   (oldest: 17 hours)
```

Exclusively `node --watch`. Never rustc, never cargo — because those paths are already guarded.

## Why this path matters most

`node --watch` is a supervisor that **by design never exits**, and it spawns a watched grandchild of its own. With no process group and no reaper, the death of the nub leader — a killed agent session, a test harness that gives up, a closed terminal — reparents both to `launchd`, where they run forever holding an fsevents watch on a source tree. They accumulate monotonically, so the only remedy was clearing them by hand once the machine slowed to a crawl.

## The fix

Both spawn sites (`cli.rs:4832` NODE_COMPAT, `cli.rs:~5084` augmented) now use `status_forwarding_signals()` — the existing helper documented as *"the signal-faithful, subtree-reaching equivalent of `cmd.status()`"*: own process group, the #480 SIGKILL-on-leader reaper held across the wait, and terminating signals forwarded to the whole subtree.

No new machinery. This is #480's reaper applied to the one path that was missing it.

## The regression test kills the LEADER, not the group — and that is the test

A first version killed the process group and **passed against unfixed code**. On the unfixed path the watched node *inherits* nub's group, so a group signal reaches it directly and no reaper is exercised. It would have shipped green forever while protecting nothing.

Killing the leader alone is also the shape of the real failure: a session dies, nub goes with it, `node --watch` does not.

Verified both directions:

| | Result |
|---|---|
| Unfixed + leader-kill test | **FAILED** — `watched node (pid 19633) survived` |
| Fixed + leader-kill test | **5 passed, 0 failed** |

## Gates

`cargo check -p nub-cli` · `cargo fmt --check` · `cargo clippy --all-targets --all-features -- -D warnings` · `cargo test -p nub-cli --test pdeath_watch` — all clean locally.

## Scope

This stops *new* orphans. It does not address the separate contention on that host (~20 concurrent builds on 10 cores, 74 live worktrees / 262 GB of target dirs) — those are structural and tracked separately.

Refs #480

### DIFF
--- a/crates/nub-cli/src/cli.rs
+++ b/crates/nub-cli/src/cli.rs
@@ -4829,12 +4829,15 @@ fn run_watch(file: &str, args: &[String]) -> Result<i32> {
         let mut node_args = vec!["--watch".to_string(), "--watch-preserve-output".to_string()];
         node_args.push(file.to_string());
         node_args.extend(args.iter().cloned());
-        let status = std::process::Command::new(node.path.as_str())
-            .args(&node_args)
+        let mut cmd = std::process::Command::new(node.path.as_str());
+        cmd.args(&node_args)
             .stdin(std::process::Stdio::inherit())
             .stdout(std::process::Stdio::inherit())
-            .stderr(std::process::Stdio::inherit())
-            .status()?;
+            .stderr(std::process::Stdio::inherit());
+        // Not `.status()` — see the note on the augmented path below. `node --watch`
+        // never exits on its own, so a bare status() leaks the supervisor and its
+        // watched child on leader death.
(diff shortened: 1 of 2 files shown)


colinhacks:

PR #621, opened 2026-07-29
aube: sync upstream v1.35.0

colinhacks:
Merges jdx/aube v1.32.0..v1.35.0 (79 commits) into `vendor/aube`.

Behavior: `add` gains `minimumPackageAge` (30d default, fail-closed; TTY prompt, else `ERR_AUBE_NEW_PACKAGE_NAME`) plus similar-name and host-prompt gates. Scoped to `add` — install/dlx ungated, public npm only, lockfile entries exempt. node-gyp bootstrap runs in-process. `update` skips deprecated versions in range.

Second commit retires nub #394: upstream `BinLaunch::Direct` covers it off the same signal, honors `preferSymlinkedExecutables: false`, and keeps `NODE_PATH`. Upstream test for the case restored, passes.

Gates green both workspaces: check, clippy `-D warnings`, fmt; aube 2771 tests.

Overlaps #612 #609 #603 #587 #583 #577 — in-tree wins, second merger resolves.


colinhacks:

PR #622, opened 2026-07-29
fix(resolver): give an undeterminable publish age its own error

colinhacks:
Refs #581

An undeterminable publish age blocked in strict mode but reported as an ordinary age gate: it listed versions never shown to be too new, and suggested `minimumReleaseAgeStrict=false`, which yields the lenient fallback, not a mature pick.

`pick_version` now returns the cause. With nothing datable, the resolver and `add` raise `ERR_AUBE_RELEASE_AGE_MISSING_TIME` (exit 28), mirroring `ERR_AUBE_TRUST_MISSING_TIME`; its help names the registry, `minimumReleaseAgeExclude` and `minimumReleaseAge=0`, and omits strictness. A dated too-new candidate outranks an undated one, so a mixed range stays an age gate.

Default-preserving: standalone aube defaults strict to false, so neither error is reachable there.


colinhacks:

PR #629, opened 2026-07-30
sandbox(win): stream a confined spawn's stdio, and give it a real IPC channel

colinhacks:
A confined piped `spawn` used to spin forever, so the shipped shim rewrote every `'pipe'` slot to a scratch file and handed the bytes back at child exit. That fixed the hang and cost three things: incremental delivery, `maxBuffer`, and `fork()` — which failed fast because a file cannot emulate a duplex pipe.

Both are now real. The pipe is created here, in the AppContainer-private namespace the same jail permits (`\\.\pipe\LOCAL\…`), and the child is handed an already-connected end as a raw fd — libuv's `UV_INHERIT_FD`, which duplicates a handle it never had to name. The parent reads the accepted end. An `'ipc'` slot is removed before `uv_spawn` can name a global pipe, and the channel is re-implemented over the same namespace as newline-delimited JSON, the framing Node's own `json` mode uses.

Both live in the ONE `ChildProcess.prototype.spawn` override the shim already had. Two stacked overrides would have an install order, and the wrong order would silently restore the refusal.

## Measured

Same driver, three arms. Node 20.19.0 / 22.15.0 / 22.23.1 / 26.5.0.

| | shipped shim | this branch | real Node |
|---|---|---|---|
| two writes 700ms apart | both at exit, gap 0ms | gap 902ms | gap 902ms |
| `maxBuffer` on a runaway child | callback never fires, exit 0 | `ERR_CHILD_PROCESS_STDIO_MAXBUFFER`, child killed | same |
| `fork()` round trip | `ERR_NUB_SANDBOX_NO_IPC` | `{"pong":"p1","connected":true}` | same |
| nested `fork()` two levels down | — | `{"relayed":"p2"}` | same |
| async spawn with no writable directory | no repair installed | streams | n/a |

The `maxBuffer` row is a defect this found rather than a property it adds: `execFile`'s maxBuffer path calls `child.stdout.destroy()`, which under the file-backed path stranded the close accounting and lost the callback entirely — silently, exit 0.

## The payload had to shrink

The shim's source percent-encodes to 56215 characters. A `CreateProcessW` environment block holds 32767 in total, so shipping it raw would have failed every jailed launch on Windows for a reason that looks nothing like its cause. Whole-line comments and indentation are dropped before encoding, bringing the delivered `NODE_OPTIONS` to 23995 characters against the 22111 the previous shim already used. A unit test pins the budget, and the semantics suite drives the delivered payload rather than the source so the strip is measured on the shipping path.

The margin is thinner than the payload alone suggests. The probe reports `env-block-chars=27557` for the whole block, so about 5.2k characters are spare for `PATH` and the `npm_config_*` family. `node-options-full-launches` gates it on a real runner.

## Scope

`spawnSync` keeps the file-backed path and must: `ParseStdioOption` has no `'wrap'` case and a synchronously-blocked loop cannot drain a pipe. That path is now also the fallback for a namespace that refuses, and it is the only thing that still needs a writable directory — the async path needs no filesystem grant at all.

Handle passing and `serialization: 'advanced'` still throw `ERR_NUB_SANDBOX_NO_IPC` naming the per-package opt-out. Neither can ride a userland stream.

`NODE_OPTIONS` re-synthesis preserves the parent's own value rather than re-deriving from `import.meta.url`, because in production that variable carries the egress gate as well as this shim, and re-deriving would drop the gate from every grandchild while the channel kept working. `a_descendant_keeps_the_whole_preload_chain` pins it with a second preload standing in for the gate, forked twice through a rebuilt `env` and an emptied `execArgv`.

## Verification

- `stdio_shim_semantics`: 8 tests, differential — every case with an unshimmed twin runs both ways and the transcripts must match verbatim. Against the shipped shim 6 of the 8 fail; the 2 that pass are the behaviours that were already correct.
- `cargo test -p nub-sandbox` green (181 lib + all integration), `cargo clippy --all-targets --all-features` clean, `cargo fmt --check` clean.
- Real AppContainer, `windows_jail_repairs` on windows-latest — [run 30528704494](https://github.com/nubjs/nub/actions/runs/30528704494), green:

```
prop:shim-spawn-stream-returns=PASS  inner=CLOSED code=0 out=pong
prop:shim-fork-roundtrip=PASS  inner=OK direct={"pong":"p1","connected":true} nested={"relayed":"p2"}
prop:shim-fork-nested=PASS   prop:node-options-full-launches=PASS  ok:written
prop:unshimmed-piped-hangs=PASS  outer=HUNG killed_after_ms=15051 cpu_ms=14984
```

  The last line is the control: with the stamp withheld the same spawn still spins, cpu 14984 of 15051 wall, so the shim is what changed the outcome. `shim-fork-nested` is the one that could only be measured here — a confined descendant creating a second private pipe of its own.

  That run is on `sandbox/win-stdio-stream-verify`, this branch's commits cherry-picked onto `sandbox/win-ace-kernel`. On this PR's own base the probe aborts in the ancestor-repair arm before reaching any stdio arm, and it does so identically at the unmodified base commit ([run 30502067382](https://github.com/nubjs/nub/actions/runs/30502067382) at `96f8de366f`, where the old `shim-fork-fails-fast` is also never measured) — pre-existing, fixed by `SetKernelObjectSecurity` on `win-ace-kernel`.

Base is `sandbox/win-stdio-stream-base`, a ref at the shared ancestor of the six in-flight sandbox branches, so the diff here is only this change. `crates/nub-sandbox` is not on `main` yet.

### DIFF
--- a/.github/workflows/win-jail-repairs-probe.yml
+++ b/.github/workflows/win-jail-repairs-probe.yml
@@ -162,10 +162,12 @@ jobs:
             need_pass "$p"
           done
 
-          # The env-block ceiling is a HARD one and the payload sits inside it by ~8.7k characters:
-          # a CreateProcessW environment block holds 32767 in total, and the stamped NODE_OPTIONS is
-          # 23995 of that. `tiny` carries a flag-sized value and `full` the real one, so a launch
-          # failure caused by SIZE separates itself from anything the shim does.
+          # The env-block ceiling is a HARD one and the margin is thinner than the payload alone
+          # suggests: a CreateProcessW environment block holds 32767 characters IN TOTAL, the
+          # stamped NODE_OPTIONS is 23995 of them, and the whole block measured 27557 on this runner
+          # — about 5.2k spare, which PATH and npm_config_* have to fit inside. `tiny` carries a
+          # flag-sized value and `full` the real one, so a launch failure caused by SIZE separates
+          # itself from anything the shim does.
           echo "---- NODE_OPTIONS delivery size ----"
           for p in node-options-tiny-launches \
                    node-options-full-launches; do
--- a/.github/workflows/win-jail-repairs-probe.yml
+++ b/.github/workflows/win-jail-repairs-probe.yml
@@ -6,8 +6,16 @@
 # WHAT IS MEASURED. Repair 1 makes every ancestor of a granted path reachable, so an absolute
 # `require()` no longer dies on `EPERM: lstat 'C:\'` — by writing a non-inherited traverse ACE
 # where the user can, and requesting the capability SIDs already on those DACLs where it
-# cannot. Repair 2 stamps a NODE_OPTIONS preload that rewrites piped stdio into scratch files,
-# so a confined `child_process` spawn returns instead of spinning in libuv's retry loop.
+# cannot. Repair 2 stamps a NODE_OPTIONS preload that creates each pipe ITSELF in the
+# AppContainer-private namespace (`\\.\pipe\LOCAL\…`) and hands the child an already-connected end
+# as a raw fd, so a confined `child_process` spawn STREAMS instead of spinning in libuv's retry
+# loop — and `fork()` gets a real channel over that same namespace rather than failing fast.
+#
+# THE TWO ARMS THAT NEED A REAL LOWBOX TOKEN, and cannot be measured anywhere else: whether
+# `fs.openSync` on a `LOCAL\` pipe name is permitted (the fd handed to the child), and whether a
+# CONFINED DESCENDANT can create a second private pipe of its own (`shim-fork-nested`). Everything
+# about how faithfully the shim reproduces `child_process` is measured cross-platform by
+# `stdio_shim_semantics`; only the namespace question lives here.
 #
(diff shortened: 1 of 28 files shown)


colinhacks:
zizmor fails with 345 findings (52 high) across the sandbox lineage's ~14 probe workflows — unpinned action refs and one impostor-commit pin. None are from this diff: the 7 naming `win-jail-repairs-probe.yml` are all on `uses:` lines this branch does not touch, and my added comment lines only shifted their line numbers. It passes on `main` because none of these workflows exist there, so this is the first PR on the lineage to audit them. Unrelated pre-existing debt.

`win-jail-repairs-probe` is running — that is the arm that matters here.

colinhacks:
Verified on a real AppContainer: [run 30528704494](https://github.com/nubjs/nub/actions/runs/30528704494) green, all stdio arms PASS. Both things that could only be measured under a live LowBox token hold — `fs.openSync` on a `LOCAL\` pipe name is permitted, and a confined descendant can create a second private pipe of its own (`shim-fork-nested`).

Control in the same run: `unshimmed-piped-hangs=PASS outer=HUNG killed_after_ms=15051 cpu_ms=14984`. With the stamp withheld the same spawn still spins, so the shim is what changed the outcome.

Measured env block: `env-block-chars=27557` of 32767, ~5.2k spare. PR body updated.

colinhacks:

PR #630, opened 2026-07-30
pm: run Windows dependency lifecycle scripts through bundled busybox sh

colinhacks:
Dependency lifecycle scripts (`preinstall` / `install` / `postinstall`) ran through `cmd.exe` on Windows, while `nub run` already defaulted to the bundled busybox-w32 `sh`. One POSIX script body therefore behaved differently depending on which Nub surface ran it. This moves the lifecycle path onto the same shell.

## Why

- **One shell.** Only aube's lifecycle spawn still hardcoded `cmd.exe`; `nub run` has defaulted to busybox since it landed. This removes a divergence rather than creating one.
- **No compatibility cost.** A sweep of the 344 widely-used packages that run install scripts found zero of 363 script bodies using cmd-only syntax — no `%VAR%`, `%~dp0`, `if exist`, `copy`/`del`/`rd`, `>nul`, `call`, `set VAR=`, caret escapes, or `.cmd`/`.bat` invocation. POSIX `sh` parses all 363. 144 use forward-slash paths and 6 chain with `&&`.
- **It fixes two packages.** `detox-recorder` and `svf-lib` invoke a shipped `./*.sh` from their postinstall, which `cmd.exe` cannot execute.
- **Fewer moving parts under the sandbox.** Confined in an AppContainer with no filesystem-ancestor repair at all, busybox's `cd`, globbing, redirection, reads, and whole spawn battery are byte-identical to unconfined. `cmd.exe` depends on the sandbox's ancestor repair to report the filesystem correctly, and that repair is best-effort by design — a refused ACE write is skipped rather than fatal, and an ancestor whose DACL names no harvestable capability yields nothing to request.

## Mechanism

A new embedder seam, `EngineContext::default_script_shell`, copied into `ScriptSettings::default_shell` by the settings pass and read by the spawn **only when the user set no `script-shell`**. Precedence is user `script-shell` → embedder default → platform default, so `None` reproduces aube's own behavior exactly.

Nub fills it on Windows with `busybox.exe sh -c`. The applet name has to lead: busybox is a multi-call binary that dispatches on `argv[0]` or a leading applet name, so `busybox.exe -c <body>` selects no shell at all. This is the same form `nub run` passes, and both now resolve the sidecar through the same `resolve_bundled_busybox`. A missing sidecar warns and leaves the engine on `cmd.exe` rather than failing read-only verbs like `nub list` that never spawn a script.

## The quoting audit

Shell selection was duplicated across the spawn and `shell_quote_arg`, which picked its dialect by `cfg(windows)`. Replacing the Windows default with a POSIX shell would have cmd-quoted args that a POSIX shell then reparses: `%` doubles to `%%`, and `$X` / backticks stay live inside the double quotes cmd-quoting emits. Both now derive from one `resolve_shell`, and the dialect follows the resolved program rather than the target platform.

The mismatch was not reachable under Nub today — every live caller passes empty args (`dlx --shell-mode` joins raw, `clean`/`purge` delegate with `&[]`, and aube's own `run`/`exec` are not registered verbs) — but it was one argument away from a silently-wrong command.

## Verification

The Windows differential is `tests/busybox-lifecycle-probe/` — a branch-scoped workflow, no PR needed. One binary installs each case twice and the only variable is whether the `busybox.exe` sidecar is present; its absence reproduces the pre-change `cmd.exe` path exactly. From [run 30527237372](https://github.com/nubjs/nub/actions/runs/30527237372) on `windows-latest`:

| case | control (no sidecar → `cmd.exe`) | with sidecar (`busybox sh -c`) |
| --- | --- | --- |
| `echo "MARK=${LIFECYCLE_PROBE:-posix}" > out.txt && test -d . && …` | **exit 0**, wrote `"MARK=${LIFECYCLE_PROBE:-posix}"  \ndirok` | exit 0, `MARK=posix\ndirok` |
| `./scripts/build.sh` | exit 1, no marker, `'.' is not recognized as an internal or external command` | exit 0, `SH_RAN` |
| `for f in src/*.txt; do cat $f >> out.txt; done` | exit 1, no marker, `f was unexpected at this time.` | exit 0, `A\nB` |

The first row is the one worth noting: `cmd.exe` exited **0** while writing the wrong bytes.

Three broken controls were found and fixed while building that probe, each of which had made a control agree with the expected answer for the wrong reason:

- A shared fixture let one case's lifecycle failure abort the `JoinSet` driving the parallel pass, killing sibling scripts mid-write — one marker came back truncated and another never appeared.
- The side-effects cache is on by default and keys on `(name, version, engine, input hash)`; the shell is not in the key, so the second arm hardlinked the first arm's built tree back and never ran the script.
- A shared package name let the content-addressed store serve one case's built directory to all three, so every case reported the first case's marker.

The probe now also asserts its own setup (`approve-builds` actually approved the build in each arm) and requires the split in **both** directions, so a case that behaves identically under both shells is reported as a failure rather than a pass.

Promoted into the committed suite:

- `install_runs_lifecycle_scripts_under_a_posix_shell` (`crates/nub-cli/tests/pm_augment.rs`) — a root postinstall whose body is POSIX-only, so a regression fails the install instead of passing quietly. Runs on every platform; on Windows CI it is the regression guard.
- `embedder_default_leading_args_survive_a_real_spawn` (`aube-scripts`) — spawns through the `<program> <applet> -c <body>` form using `/usr/bin/env` as a same-shaped stand-in, so the form is exercised on a platform this suite can run. Negative control: replacing `spec.args` with `vec!["-c"]` fails this test and the structural one, and leaves the other three green.
- `busybox_lifecycle_shell_leads_with_the_sh_applet_name` (`pm_engine`) — pins the applet form, a Windows-only breakage no other test here would see.

Gates, all from the worktree:

```
cargo fmt --check                                                  clean (nub + the four aube files)
cargo clippy --all-targets --all-features --profile fast -- -D warnings   clean
cargo test                                                         green, full nub suite
cargo test (vendor/aube)                                           green, full aube suite
```

Four `integration.rs` transpile tests failed on the first run and were **not** related to this change — the worktree had no root `node_modules`, so `@oxc-project/runtime` was unresolvable. Confirmed by running `npm ci` and re-running: 213 passed, 0 failed.

## Not verified

- Whether `cmd.exe`'s AppContainer behavior is repaired by the sandbox's production ancestor grant. That is being re-measured separately and this change does not depend on the answer.
- The win32-arm64 sidecar, `busybox64a.exe`. The probe runs on `windows-latest`, which is x64, so the arm64 sidecar is exercised only by the release matrix.

### DIFF
--- a/.github/workflows/busybox-lifecycle-probe.yml
+++ b/.github/workflows/busybox-lifecycle-probe.yml
@@ -1,0 +1,57 @@
+# Branch-scoped Windows probe (see .claude/skills/ci-adhoc-test) — NO PR needed.
+# Pushing to the busybox-lifecycle-shell branch runs this on a real
+# windows-latest runner. It builds one nub and drives the same dependency
+# postinstall fixtures twice, with `busybox.exe` staged next to nub.exe and
+# without it — the sidecar's presence is the only variable, and its absence
+# reproduces the pre-change cmd.exe behavior exactly.
+name: busybox-lifecycle-probe
+on:
+  push:
+    branches: [busybox-lifecycle-shell]
+    paths:
+      - 'tests/busybox-lifecycle-probe/**'
+      - '.github/workflows/busybox-lifecycle-probe.yml'
+      - 'crates/nub-cli/src/pm_engine/mod.rs'
+      - 'crates/nub-cli/src/cli.rs'
+      - 'vendor/aube/crates/aube-scripts/**'
+      - 'vendor/aube/crates/aube-util/src/engine_context.rs'
+      - 'vendor/busybox-w32/**'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  probe:
+    name: busybox dependency lifecycle (windows-latest)
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd  # v5.0.1
+        with:
+          persist-credentials: false
+      - uses: dtolnay/rust-toolchain@4be7066ada62dd38de10e7b70166bc74ed198c30  # stable # zizmor: ignore[impostor-commit,ref-version-mismatch] dtolnay force-pushes its version tags, so pinned SHAs go unreachable; commit verified genuine upstream
+      - uses: Swatinem/rust-cache@e18b497796c12c097a38f9edb9d0641fb99eee32  # v2
+        with:
+          shared-key: busybox-lifecycle-probe
+      - uses: actions/setup-node@a0853c24544627f65ddf259abe73b1d18a591444  # v5
+        with:
+          node-version: 24
+          cache: 'npm'
+      # Root deps + the nub-native addon, mirroring ci.yml's Windows test leg —
+      # the install path resolves the augmentation preload through them.
+      - run: npm ci
+      - name: Build nub-native addon
+        shell: bash
+        run: |
+          (cd crates/nub-native && cargo build)
+          mkdir -p runtime/addons
+          cp target/debug/nub_native.dll runtime/addons/nub-native.node
+      - name: Build nub
+        run: cargo build -p nub-cli
+      # The probe stages and un-stages target/debug/busybox.exe itself, so it is
+      # deliberately NOT copied here.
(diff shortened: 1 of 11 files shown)


colinhacks:

PR #631, opened 2026-07-30
scripts: add re-runnable npm install-script census

colinhacks:
Adds a re-runnable census of npm packages above a weekly-download threshold, recording per version whether they run install lifecycle scripts. This is the corpus the build jail's compatibility claims rest on. The previous one held 344 packages, 293 of them copied from bun's trusted-dependencies list — a curation axis, not a download ranking.

## What it measures

Three sources, each for the job it is actually good at:

```
ecosyste.ms ?sort=downloads    complete registry enumeration (not search-seeded)
api.npmjs.org                  authoritative weekly counts, package and per-version
registry.npmjs.org packument   the scripts object, per version
```

A package counts as install-relevant when any of its top 3 versions by download count carries a preinstall, install, or postinstall script. Those three are the only keys npm runs for a registry dependency; prepare and prepublish are recorded separately and never counted. N is a parameter, and every version of every candidate is cached with its own download count, so changing it re-derives from cache without refetching.

## Why per version

Classifying from `latest` understates the corpus, because lockfiles pin old versions and installs pile up there.

| package | top-downloaded version | weekly | install script | `latest` |
|---|---|---|---|---|
| sharp | 0.34.5 | 47,808,857 | `install: node install/check.js \|\| npm run build` | 0.35.3, clean |
| protobufjs | 7.6.5 | 20,953,325 | `postinstall: node scripts/postinstall` | 8.7.1, clean |
| better-sqlite3 | 12.11.1 | 2,431,794 | `install: prebuild-install \|\| node-gyp rebuild --release` | 13.0.2, clean |

A latest-only census scores all three as clean.

## Results

Ranked by ecosyste.ms monthly downloads, classified from cached packuments:

| definition | top 5,500 | top 10,000 |
|---|---|---|
| latest only | 37 | 81 |
| top 3 versions | 50 | 110 |
| any version ever | 201 | 384 |

The 100,000 weekly bar falls around eco rank 24,000, while eco's top 10,000 bottoms out near 287,000 weekly — roughly 3x above the bar. Both rank cutoffs in prior use therefore cover well under half the population the threshold describes.

An earlier 38/82 figure reproduces here as 37/81. A competing ~60/~130 figure resolves to a ranking artifact: re-ranking the same packages by npm weekly instead of ecosyste.ms monthly moves the latest-only count from 37 to 51 at top-5,500. Both were internally consistent for their own ranking choice, and both understate by roughly 35% because both classified from `latest`.

## Two traps a future maintainer must not reintroduce

Both silently corrupted results, and both are documented in comments at their sites.

- **A failed lookup must never be cached as zero.** It makes the package read as below-threshold permanently, because a re-run reads the cached zero. This put 431 zeros in the cache and dropped 219 packages from a band whose lowest member has 54,622,466 monthly downloads. Failed lookups are now left uncached so a later run retries them.
- **Unmeasured is not the same as below-threshold.** An unresolved candidate has no weekly figure at all, and treating that absence as zero omitted it from the output as though it had been measured and found small. Coverage is now reported in the summary, so a partial run states its own incompleteness.

## Rate limits

The download API sustains about 76 lookups per minute and rejects scoped names in bulk lookups, so each scoped package needs its own request. Requests are paced per host with multiplicative backoff on 429 and a per-host in-flight cap, both calibrated against burst tests rather than guessed. An earlier cut escalated retry waits and split a throttled 128-name batch into 128 singles, amplifying load 128x at the moment the host asked for less; that pinned throughput at 43 per minute.

Ecosyste.ms reports npm's last-month figure but only for entries it synced recently, and its error tracks `last_synced_at` — up to 134x on entries months stale. That bound is what lets a cheap deep sweep run without querying npm about every candidate; the margin over the observed maxima is validated by sampling excluded entries, and the residual leak is reported rather than assumed away.

Datasets are written to gitignored `.fray/` and are not part of this change.

### DIFF
--- a/scripts/npm-install-script-census.ts
+++ b/scripts/npm-install-script-census.ts
@@ -539,7 +539,8 @@ const stageWeekly = async (names: string[], label: string): Promise<Map<string, 
       // attempt. A throttled batch is left uncached and picked up by the next run.
       for (const n of batch) {
         const s = await getJSON(POINT + enc(n));
-        record(n, s.ok ? Number((s.body as { downloads?: number }).downloads ?? 0) : 0);
+        if (s.ok) record(n, Number((s.body as { downloads?: number }).downloads ?? 0));
+        else unresolved.push(n);
       }
     } else {
       unresolved.push(...batch);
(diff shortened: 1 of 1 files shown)


colinhacks:

PR #632, opened 2026-07-30
sandbox(win): write the ancestor traverse ace with SetKernelObjectSecurity

colinhacks:
The ancestor repair writes a traverse ace on every ancestor of a granted path, once per lifecycle spawn. It went through `SetSecurityInfo`, which runs advapi32's inheritance-propagation pass before returning — so the write walked the directory's existing subtree, and the chain includes `%TEMP%`. That is what wedged a 20-minute CI step and what killed the last two runs of this probe: both died on the first repair-ON launch, `watchdog-stalled-at=run_jailed mode=fsprobe`, exit 97, with no group after `ancestor_repair` ever running.

`SetKernelObjectSecurity` goes straight to `NtSetSecurityObject`: it writes the object's own descriptor and there is no propagation pass to skip. It wants a whole descriptor rather than a bare ACL, so the descriptor is hand-built (`InitializeSecurityDescriptor` + `SetSecurityDescriptorDacl`). `SetEntriesInAclW` still does the merge — it only assembles an ACL in memory and never propagated anything. `SE_DACL_AUTO_INHERITED` and `SE_DACL_PROTECTED` are carried across so a fresh descriptor cannot clear them on a directory nub does not own.

Second change, same root cause on the other side. A leaf grant is inheritable by design, so it must propagate — the lever there is not to write it when the target already publishes it. Measured: `%ProgramFiles%` carries `ALL APPLICATION PACKAGES: ReadAndExecute` inheritably, so an all-users python or node needs no ace at all; only per-user layouts pay. The teardown list is driven from `grant_leaf_ace`'s single return value, so a skipped path is never recorded and the revoke cannot strip an ace the launch did not create.

## Measured, windows-latest, run 30528232196

Same trustee, same path, same run. The propagating writer is kept as a local control in the probe rather than in the product.

| grant on | new (`SetKernelObjectSecurity`) | old (`SetSecurityInfo`), control |
|---|---|---|
| empty directory | 209 µs | 370 µs |
| 4,000-entry tree | **131 µs** | **534,378 µs** |
| re-grant, ace already present | 140 µs | — |
| real `%TEMP%` on the runner | **630 µs** | previously minutes |

The control reproduced: the old writer went 370 µs → 534 ms with tree size, a 1,444× blow-up. The new one is flat, and a re-grant costs the same as a fresh one, which is a descriptor write rather than a tree walk.

Effect and scope, asserted rather than assumed — the ace lands carrying exactly `0x001000a1` with flags `0x00`, is absent on a child directory, is gone after revoke, and the DACL control word is `0x8404` before and after. Seen from inside the jail, `repair-on-lstat-chain-permitted` and `repair-on-node-absolute-require-resolves` pass while `repair-on-ancestor-sibling-read-refused` and `-listing-refused` still refuse, so the grant reaches the confined child and stops at the directory object.

For the skip: `aap-readable-programfiles=true`, `aap-readable-programfiles-nodejs=false` (the known outlier), a protected user-only tree is correctly declined, and a seeded inheritable AAP ace survives a launch that granted the path.

This is the workflow's first green run. Four `shape-plain-*-unrepaired` cells are FAIL, byte-identical to both baseline runs — they are the repair-OFF blast-radius measurement and are un-gated by design.

## Base

Branched off `sandbox/win-jail-bin`, the newest tip carrying both the ancestor repair and the probe harnesses. `set_ace_on_object` is byte-identical across all seven branches that have it, so this applies to any of them.

### DIFF
--- a/.github/workflows/win-jail-repairs-probe.yml
+++ b/.github/workflows/win-jail-repairs-probe.yml
@@ -17,13 +17,19 @@
 #
 # The icacls dump is a FACT, not a gate: it names the capability SIDs on `C:\` and `C:\Users`
 # that the backend harvests, which is what makes the ancestors reachable without elevation.
+#
+# THE COST GROUP. Repair 1's ace is written per lifecycle SPAWN, and the propagating writer it
+# used to go through wedged a 20-minute step and stalled three runs of this probe. It now writes
+# through `SetKernelObjectSecurity`. `ace-cost-named-writer-scales-with-tree` is that group's
+# control — a run where the propagating writer did NOT slow down on the populated tree measured
+# nothing, so it is gated as a PASS alongside the kernel cells rather than reported as a fact.
 name: win-jail-repairs-probe
 
 on:
   push:
     # Path-filtered so a change to a SIBLING probe does not spend ~25 minutes of Windows runner
     # on this one. Shared capacity: every push here costs the full matrix.
-    branches: [sandbox/win-jail-caps]
+    branches: [sandbox/win-jail-caps, sandbox/win-ace-kernel]
     paths:
       - 'crates/nub-sandbox/**'
       - '.github/workflows/win-jail-repairs-probe.yml'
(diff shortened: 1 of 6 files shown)


colinhacks:
Reproduced on a second independent run, 30528865105.

| grant on | new (r1 / r2) | old, control (r1 / r2) |
|---|---|---|
| empty directory | 209 / 205 µs | 370 / 400 µs |
| 4,000-entry tree | 131 / 130 µs | 534,378 / 616,863 µs |
| real `%TEMP%` | 630 / 654 µs | minutes |

All 13 `ace-*`/`aap-*` properties PASS in both, watchdog fired in neither, both reached `STEP done`.

`zizmor` is red on every branch in this lineage, including ones touching no workflow; its six findings against `win-jail-repairs-probe.yml` are unchanged by this diff.

colinhacks:

PR #633, opened 2026-07-30
sandbox(win): run lifecycle scripts on a nub-owned copy of the project's Node

colinhacks:
The Windows build jail could not start a single dependency lifecycle script for a standard user whose Node came from the stock MSI. Two independent reasons, both now closed by running the script on a nub-owned copy of the project's own Node.

## The defect

A leaf read grant is an ACE, and writing one needs `WRITE_DAC` on the target. The stock MSI installs to `%ProgramFiles%\nodejs`, whose DACL — measured on this PR's own run — grants a standard user read but not that:

```
C:\Program Files\nodejs  BUILTIN\Users:(OI)(CI)(RX)
                         NT AUTHORITY\Authenticated Users:(OI)(CI)(RX)
                         BUILTIN\Administrators:(OI)(CI)(F)
                         NT AUTHORITY\SYSTEM:(OI)(CI)(F)
```

No `ALL APPLICATION PACKAGES` entry, so the already-published skip cannot apply either — that skip needs an *inheritable* AppContainer ace, which no file can carry. `C:\hostedtoolcache` is the same shape. De-elevated, the grant failed `Access is denied. (os error 5)` and `?` aborted the launch.

Widening the DACL would not have been the fix even where nub can write it. `CreateProcessW` opens the image in the CALLER's context, so once the caller is itself inside the AppContainer, opening that image is a confined open and is refused — measured previously against the identical command line unconfined. Everything a lifecycle script spawns is such a nested spawn.

## The fix

The interpreter is now a copy of the same distribution under `<cache>/jail-bin/<version>-<arch>`, keyed by the version the interpreter reported. That key is load-bearing rather than incidental: `prebuild-install` and `node-gyp-build` both default the ABI they fetch for to `process.versions.modules` of the running Node, so staging a newer Node would make that whole family download a prebuild that then fails to load. A copy of the project's own Node cannot diverge.

The ace is written on the EMPTY staging directory and inherited at creation. That is both the cheap direction — 24 ms against 426 ms re-granting a populated 2,435-entry tree — and what makes the backend's own per-spawn leaf grant skip entirely, so a lifecycle script pays nothing. Population is by copy, never hard link: an NTFS hard link shares one MFT record and therefore one security descriptor, so an ace written on the link lands on the original.

Two smaller changes ride along.

A refused **read** grant is now skipped rather than fatal, matching the ancestor-chain repair two dozen lines below it. A read grant may legitimately name a toolchain the user holds no `WRITE_DAC` on, and aborting every lifecycle script over one unreachable toolchain is the loudest possible failure for the mildest cause. A grant is a reduction from the unconfined spawn's complete access, so skipping one leaves the child with less. Write grants stay fatal — every one is nub's own tmp or the package dir being built.

The extra-read grant's doc described only the POSIX layout, whose rationale is false on Windows: that archive ships no `include/` and no `lib/`, so `<node-root>/lib/node_modules` named a path that does not exist and `include/node` described headers the distribution does not carry. Both halves now state the shape they belong to, with node-gyp 13's actual behaviour cited (`lib/configure.js` `getNodeDir`, `lib/install.js`) — it reads `<nodedir>/include/node` and `<nodedir>/$(Configuration)/node.lib` when nodedir is set, and downloads into its own `%LOCALAPPDATA%\node-gyp\Cache` devDir when it is not, which the jail's deny-all egress refuses. The pinning test covered only the POSIX shape, which is how the Windows spelling stayed inert; it now covers both, with a sibling refusal so a widened grant is caught too.

## Verified

Run [30545352967](https://github.com/nubjs/nub/actions/runs/30545352967): 16/16 properties PASS in both the elevated and de-elevated arm, against a real Node 24.18.1 MSI installed to `%ProgramFiles%\nodejs`, de-elevated on a restricted token (`admin=0`, integrity 8192, `write_dac=false` on the ambient interpreter).

| de-elevated property | what it establishes |
| --- | --- |
| `interp-staging-needs-no-write-dac-anywhere-privileged` | mkdir + ace + 2,435-entry copy, no elevation |
| `interp-staged-copy-runs-a-lifecycle-script-without-write-dac` | a real `.cmd` postinstall through `cmd.exe` completes |
| `interp-staged-child-execpath-is-the-nub-owned-copy` | the child's own `process.execPath` is the copy, not the ambient install |
| `interp-staged-abi-matches-the-source` | `process.versions.modules` equal on both sides (137) |
| `interp-staged-dir-publishes-read-to-appcontainers` | the per-spawn leaf grant skips |
| `interp-staged-deep-entry-inherited-the-ace-at-creation` | the ace preceded the copy |
| `interp-staged-ungranted-secret-still-refused` | confinement survives |
| `interp-ambient-install-is-usable-only-where-nub-can-write-its-dacl` | the dependence, stated so it holds in both arms |

The execPath cell is the attribution control, and it matters now that a refused read grant is skipped: a launch can succeed while the child quietly runs the ambient interpreter, and only the child reporting its own binary tells those apart. The probe drives the same engine entry the product does rather than a re-implementation of the staging.

Getting there took four rounds, all four defects in the probe rather than the fix, and each is written up in its commit. The one worth repeating: the first execPath control compared path *spellings*, so a `/`-containing `join` literal failed against Windows's own separators while the fix worked underneath — a control that varied more than the thing under test.

## It also settles a sibling lane's cmd.exe verdict

A sibling lane measured `cmd.exe` de-elevated as 19-of-19 cells absent — exit 1 at 100 ms, empty transcript — and read it as cmd being unusable under confinement. That reading was an artifact of the fail-closed behaviour this PR replaces. Run [30547014209](https://github.com/nubjs/nub/actions/runs/30547014209), de-elevated, one variable:

| | fail-soft | fail-closed |
| --- | --- | --- |
| cmd cells | 7/9 | **0/9** |
| rc | 0 | -101 |

```
sandbox: installing read grant ACE on C:\Windows\system32\cmd.exe failed: Access is denied. (os error 5)
```

That path is the program file `resolve_program` auto-grants, and that string is verbatim what the sibling lane saw. The seam (`NUB_SANDBOX_WIN_FAIL_CLOSED_READ_GRANTS`) exists so both directions are measurable in one run; it can only make the jail stricter. The sibling's own busybox row is the same mechanism from the other side — it is staged into its fixture, so its grant reads `writable` where `cmd-exe` reads `refused:Some(5)`. A nub-owned program is grantable, which is what this PR does for the interpreter.

Not overstated: pwsh ran confined de-elevated and `powershell.exe` started, both with equally-refused grants, so "fail-closed aborted everything refused" is not a universal account. The cmd case is what is established.

Two de-elevated residuals remain, both belonging to the ancestor/traverse work rather than the interpreter: `dir /b` is refused in the package dir while it works elevated, and `where.exe node` misses when PATH points at the staged tree. The cell tally is therefore a fact rather than a gate; what is gated is the part read as zero — cmd starts, opens its script, resolves `node`, and reaches its verdict line.

## Limits

- **The bundled-npm read is verified elevated only.** Reading `node_modules/npm` out of the distribution is recorded as a fact, not gated, because a CJS `require` of an absolute path realpath's every prefix from the volume root and dies de-elevated with `EPERM … lstat 'C:\'` in *both* interpreter arms identically. That is the open ancestor-reachability repair, not this one, and no interpreter work will change it.
- **The de-elevated arm did not independently reproduce the ambient failure**, because the elevated arm runs first and leaves `%ProgramFiles%\nodejs` AppContainer-readable — measured directly this run (`publishes-aap` false → `publishes-aap-after` true). The mechanism is unexplained: the arm's own teardown revokes a different sid than that read-back matches. `tests/win-msi-volume` already owns that question, and this PR's fix means nub no longer grants the ambient install at all. The de-elevated evidence for the original failure is the MSI DACL above plus `write_dac=false`.
- Neither the piped-`child_process` hang nor the ancestor repair is touched here.
- **`can_write_dacl` is reported as a fact and is not sound above medium integrity.** It returned false on `System32\cmd.exe` in the elevated arm where the real grant plainly succeeded — an elevated token's DACL-write authority does not come from the file's DACL, so an access-checked open cannot see it. The gated property keys on the launcher's own refusal instead. It is labelled in the log so nobody keys a verdict on it again.

## Gates

`cargo fmt --check`, `cargo clippy --all-targets --all-features -- -D warnings`, and `cargo test -p nub-sandbox` are green. Four `nub-cli` transpilation tests fail in a fresh worktree for want of `node_modules/@oxc-project/runtime`, and `init_cmd`'s registry-fixture test flaked once under a full parallel run (its fixture is a busy-spin non-blocking listener against a client configured `fetch-retries=0`); both are pre-existing and unrelated.

Refs #561

### DIFF
--- a/.github/workflows/win-jail-interp-probe.yml
+++ b/.github/workflows/win-jail-interp-probe.yml
@@ -188,9 +188,15 @@ jobs:
           # a System32 program's own grant is attempted and refused de-elevated — under `?` that
           # aborted the launch, which looks exactly like cmd misbehaving.
           echo "---- cmd.exe: fail-closed vs fail-soft, one variable ----"
-          need_pass interp-cmd-under-fail-closed-aborts-only-when-its-own-grant-is-unwritable
-          need_pass interp-cmd-under-fail-soft-runs-the-whole-battery
-          grep -E '^  fact:interp-cmd-program-grant-writable=' probe.log || { echo "MISS the program-grant writability fact"; fail=1; }
+          need_pass interp-cmd-under-fail-closed-aborts-iff-its-own-program-grant-is-refused
+          need_pass interp-cmd-under-fail-soft-reaches-step-cmd-opened-the-script
+          need_pass interp-cmd-under-fail-soft-reaches-step-bare-node-resolved-and-ran
+          need_pass interp-cmd-under-fail-soft-reaches-lifecycle-ok
+          grep -E '^  fact:interp-cmd-program-grant-write-dac-open=' probe.log || { echo "MISS the program-grant fact"; fail=1; }
+          # The cell TALLIES are facts, not gates. De-elevated `dir /b` is refused in the package dir
+          # while it works elevated — a real residual that belongs to the ancestor/traverse work, not
+          # to the interpreter — so a full 9/9 is not the bar here. What IS gated is the part the
+          # sibling lane read as zero.
           grep -E '^  fact:interp-.*-cmd-cells=' probe.log || { echo "MISS the cmd cell tallies"; fail=1; }
 
           echo "---- the pre-existing groups must not have regressed ----"
--- a/.github/workflows/win-jail-interp-probe.yml
+++ b/.github/workflows/win-jail-interp-probe.yml
@@ -1,0 +1,185 @@
+# Branch-scoped ad-hoc probe: can the Windows build jail start a lifecycle script for a STANDARD
+# USER whose Node is an all-users MSI install? No pull request required
+# (.claude/skills/ci-adhoc-test/SKILL.md). CI is the only venue — AppContainer cannot be launched
+# over SSH (session 0 has no window station, so every launch returns 0xC0000142), which also rules
+# out the standing nub-win VM.
+#
+# WHAT IS MEASURED. A leaf read grant is an ACE, which needs WRITE_DAC on the target; the stock
+# Node MSI installs to %ProgramFiles%\nodejs, where a standard user does not hold it. And even
+# where nub can write it, CreateProcessW opens the image in the CALLER's context, so once the
+# caller is itself in the AppContainer the un-ACE'd image is refused anyway. The fix is that the
+# interpreter is a nub-owned COPY of the same distribution, published AppContainer-readable by
+# granting the EMPTY directory and letting the copy inherit. This runs the same lifecycle script
+# against both interpreters, under the same real compile_build_jail policy, on both an ELEVATED and
+# a DE-ELEVATED token.
+#
+# THE MSI STEP IS THE POINT, and it needs admin — which is not a finding about nub, because a user
+# installs Node with the same consent. What matters is the DACL the installer leaves behind, and
+# that the de-elevated arm afterwards holds no authority to change it. Without this step the probe
+# would measure the runner image's unzipped hostedtoolcache Node, which is a different tree with a
+# different (also un-ACE'd) DACL — usable, but not the configuration a real user has.
+#
+# THE CONTROLS. `interp-ambient-install-cannot-run-a-lifecycle-script` is the arm where the defect
+# still reproduces: without it a green staged arm measures nothing.
+# `interp-staged-child-execpath-is-the-nub-owned-copy` is the attribution control — a refused read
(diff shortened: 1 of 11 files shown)


colinhacks:

PR #634, opened 2026-07-30
sandbox(win): pre-resolve Visual Studio for the build jail

colinhacks:
node-gyp finds MSVC through the VS SetupConfiguration COM server, which an AppContainer cannot activate (`80070005`, run 30522944978). Not a DACL, so no grant reaches it.

Resolve it in the unconfined parent — `vswhere`, then that installation's `vcvarsall.bat` — and stamp `VCINSTALLDIR`/`VSCMD_VER`/`WindowsSDKVersion` into the jailed spawn, plus the allowlist entries that pass them through.

All three or none: `VCINSTALLDIR` alone makes `checkConfigVersion` reject every other installation while supplying no version or SDK. `accept` re-runs node-gyp's predicates.

Read grants on the VS and SDK roots, redundant for a default install. Probe: two arms differing in one seam.

### DIFF
--- a/.github/workflows/win-msvc-inject-probe.yml
+++ b/.github/workflows/win-msvc-inject-probe.yml
@@ -159,10 +159,22 @@ jobs:
             need "$f" "prop:$S/jail-enforced=PASS" "control/$f"
           done
 
-          # THE FINDING.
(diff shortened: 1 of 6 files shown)


colinhacks:

Issue #473, opened 2026-07-14
Provision nub in Cloudflare Workers Builds / Pages build image


colinhacks:
Track provisioning nub in Cloudflare's build image (Workers Builds / Pages), so package scripts that call `nub`/`nubx` resolve without a per-project workaround.

Cloudflare's builder ships a fixed, pre-installed toolset and runs the install itself — there's no step to inject [`nubjs/setup-nub`](https://github.com/nubjs/setup-nub), and no user-facing hook to add a tool. Its build-image docs pin versions of the existing tools via `.node-version`/`.nvmrc`/env vars — not `.tool-versions`, and there's no plugin mechanism. A repo whose scripts call `nub` then fails with `sh: nub: not found`.

**Available install paths (done — for asdf/mise users generally):**
- Official asdf plugin: https://github.com/nubjs/asdf-nub — CI-green on Linux + macOS, SHA-256-verified release binaries.
- Registry submission: asdf-vm/asdf-plugins#1170 — makes `asdf plugin add nub` resolve by short name.
- mise already supports nub via its registry (`mise use nub@latest`).

**Upstream:** cloudflare/pages-build-image#13. None of the above unblocks Cloudflare on its own — the image has no user-addable tools, so nub lands only once Cloudflare adds it to the pre-installed set themselves. Downstream, not nub-side work; the asdf plugin at most makes it trivial for them if their image pipeline uses asdf internally.

**Workaround today (the only path for Cloudflare users):** pin `@nubjs/nub` as a devDependency; the pnpm install the builder runs puts `nub`/`nubx` on `node_modules/.bin`. See the [Deployment docs](https://nubjs.com/docs/deployment).

ericclemmons:
> 2. **Cloudflare adds nub to the build image's plugin set** ([cloudflare/pages-build-image#13](https://github.com/cloudflare/pages-build-image/issues/13)) — Cloudflare. The actual fix; everything above is the enabling groundwork.

This will be live in ~hour!

> 3. **Verify + document once it lands** — nub side, small: confirm `.tool-versions` provisioning works end-to-end on a real Workers Builds / Pages build, then add a short "provisioned natively via `.tool-versions`" note to `deployment/index.mdx` alongside the existing mise line, and close [#473](https://github.com/nubjs/nub/issues/473).

Let me know on X if this doesn't work as expected 🙏

colinhacks:
@ericclemmons Woohoo! 😍

colinhacks:
[...]

PR #639, opened 2026-07-31
site: add Cloudflare and Vercel deployment pages

colinhacks:
Two pages under Deployment, plus index routing.

**Cloudflare** — nub landed in the Workers Builds image 2026-07-30 (cloudflare/pages-build-image#13). The image already ships Node 22.23.2/24.18.0, so only the binary is new. Pages is not covered.

Not independently verified: Cloudflare's docs and changelog don't mention nub, and there is no public image to inspect. Probe at colinhacks/nub-cf-probe answers it from a build log. Hold until green.

**Vercel** — verified across three deployments. Calling the binary by absolute path half-works: nub runs but is off PATH, so a script calling nub or nubx fails. Both commands export PATH.

devDependency example: ^0.1.0 to ^0.6.0.

Refs #473


colinhacks:

PR #640, opened 2026-07-31
fix(install): detect cgroup-v1 PID limits inside containers; repair remote-build

colinhacks:
**cgroup-v1 PID limits go undetected in containers.** `cgroup_v1_pids_max` joined the pids mount point with the host-absolute path from `/proc/self/cgroup`, but that mount is already rooted at our cgroup, so the read failed open and `--pids-limit 64` got a 128-thread pool. Now strips the mountinfo root, longest-root first. v2 unaffected: Docker uses `cgroupns=private` there.

Verified on Ubuntu 20.04/Docker and Rocky 8/podman vs the shipped release and a merge-base build; 32 cgroup-v2 cells byte-identical.

**remote-build never completed a job:** it set `AUBE_REQUIRE_PRIMER=1` (ci.yml does not; the primer is gitignored) and ran clippy without `--profile fast`. Adds `--detach`/`--attach`.


colinhacks:

PR #644, opened 2026-07-31
fix(pm): move the phantom sidecar tier with a store-dir override

### LINKED ISSUE #643: store-dir override leaves the phantom sidecar tier at the default store
A `store-dir` override (`npm_config_store_dir` / `.npmrc` `store-dir` / workspace `storeDir`) moves the CAS tiers (`v1/files`, `v1/index`, `v1/trees`) but not the phantom sidecar tier: `dynamic_phantom::store_v1_dir()` derives from the default data dir unconditionally. An overridden install still writes `store/v1/phantom/` sidecars into the machine's real store, and the link-time consumer opens the default store handle, so under an override every package degrades to "not flagged" and phantom ejection silently stops working. Repro: `npm_config_store_dir=/tmp/scratch nub add ./fixture.tgz` → the CAS lands under `/tmp/scratch/v1`, the sidecar lands in `~/.local/share/nub/store/v1/phantom/`.

jdalton:
Closes #643.

Nub can be told to keep its package store somewhere other than the default — a CI runner, a scratch volume, a test sandbox that must not touch the machine's real store. When you do that today, **nub's phantom-dependency protection silently stops working**: every package comes back "not flagged," so a dependency your code imports without declaring resolves by accident instead of failing. Nothing warns you.

The same run also still writes into the real store it was told to avoid, so the sandbox is not a sandbox. This PR makes the store override move every tier, so relocating the store keeps the protection intact and leaves the machine's real store untouched.

<details>
<summary><b>What went wrong</b> — the content store moved, the phantom sidecar tier did not</summary>

A `store-dir` override — per-invocation `npm_config_store_dir`, or `store-dir` via `nub config set` / `.npmrc` / workspace `storeDir` — relocated the content-addressable store (the CAS, where package files actually live) but left the phantom sidecar tier at the default data directory. Sidecars are the small per-package records that link time reads to decide whether a package is phantom.

That split produced two failures at once:

| Failure | Consequence |
| --- | --- |
| The writer kept the default path | an overridden install still wrote `store/v1/phantom/` into the machine's real store |
| The reader opened the default store handle | it found nothing for the relocated packages, so phantom detection degraded to "not flagged" for **every** package |

The second one is the quiet one. Detection did not error — it returned a clean result.

</details>

<details>
<summary><b>The fix</b> — one resolution path, so all four store tiers move together</summary>

`dynamic_phantom::store_v1_dir()` now routes through the engine's own `storeDir` resolution (`aube::commands::resolved_project_store_dir`, newly `pub`) instead of deriving from the default data directory. There is now a single answer to "where is the store," and the sidecar tier gets the same answer as the CAS tiers.

The anchor is the load-bearing half. Neither `.npmrc` nor `pnpm-workspace.yaml` discovery walks up, while the install pipeline anchors at `workspace_or_project_root()` — so resolving against the raw process cwd fixes a single-package project but still misses the override for every command run from inside a workspace member, leaving the sidecar tier back at the default store. `resolved_project_store_dir` picks the anchor itself, and `resolved_store_dir` stays crate-internal so an embedder cannot reintroduce the bug by passing a wrong one.

`nub store path` carried the same wrong anchor: from a workspace member it reported the default store while installs from that member used the root's override. Real pnpm reports the override from both, so it moves to the workspace-aware anchor too, keeping its outside-a-package-tree fallback.

One timing detail: the extract hook resolves its directory on first fire rather than at registration, because registration happens before the session's `--dir` chdir. Memoizing that answer is safe only under the walked-up anchor — the recursive verbs retarget the cwd per member mid-process, and every member of one workspace walks up to the same root.

</details>

<details>
<summary><b>The new docs section</b> — the supported way to keep CI and probes off the real store</summary>

`install/index` gains a **Relocating the store** section. Until now the only way to keep a probe or a CI run away from the machine's real store was to override `XDG_DATA_HOME` process-wide, which moves far more than intended. The section documents `store-dir` as the supported knob — the same one pnpm exposes under that name — and spells out the precedence order and where the caches live separately.

</details>

<details>
<summary><b>What I ran</b> — seven offline tests, plus a before/after repro on both project shapes</summary>

**Ran** — new tests in `pm_publish_store_config`, all offline:

| Test | What it pins |
| --- | --- |
| env override | `npm_config_store_dir` relocates the store |
| project-`.npmrc` override | the file form works too |
| relative path | resolves against the project root, not the cwd — invoked from a SUBDIRECTORY, since at the root the two anchors are the same directory and the test cannot tell them apart |
| env-over-npmrc precedence | the environment wins |
| `store_dir_override_relocates_every_store_tier` | pack, then add a local tarball under an override — CAS **and** sidecar tiers land under it, and the default store location is never created |
| `store_dir_override_relocates_every_tier_from_a_workspace_member` | the same invariant with the override declared at the workspace root and the install run from a member — the shape a monorepo actually installs in, and the one an anchor bug hides in |
| `store_path_reports_the_workspace_store_from_a_member` | `store path` reports the store the workspace root configures, matching what installs from that member use |

**Ran** — `pm_publish_store_config` 38/38; `nub-launcher` `--locked` clippy + build + test 73/73; `cargo fmt --check` and `cargo clippy --all-targets --all-features --profile fast -- -D warnings` clean on the root workspace, `nub-native` and `nub-launcher`; all three workspace locks satisfiable under `--locked --offline`.

Note for anyone reproducing the lint gate locally: `--all-features` enables `embed-runtime`, whose `nub-core` build script hashes `runtime/addons/nub-native.node` and fails loudly when it is absent. Stage the placeholder CI stages, or the gate fails for a reason unrelated to the change.

**Ran** — manual repro on macOS, both shapes. Before the fix, an overridden `nub add ./fixture.tgz` wrote `…/nub/store/v1/phantom/s4/<fp>.json` into the default store. After it, the sidecar lands under `<override>/v1/phantom/` and the default store directory is never created — from a single-package project AND from a workspace member.

**Ran** — the consumer half, which is the silent one. With the CAS warm, wiping the sidecar tier and `node_modules` and re-installing offline rewrote **0** sidecars before the fix (link-time detection never reached the relocated CAS, so every package came back "not flagged") and **11** after.

</details>


jdalton:
Added a black-box repro of #643 against the released `@nubjs/nub@0.6.0` from npm, in case it is useful alongside the local-build repro here: https://github.com/nubjs/nub/issues/643#issuecomment-5150143109

Two bits that bear on this change:

- A run with `npm_config_store_dir` and `XDG_DATA_HOME` pointed at *different* directories puts the CAS in the first with no `phantom/` subdirectory and the sidecar in the second. So the sidecar path is a function of the XDG data base specifically — which is what routing through `resolved_store_dir` fixes, and it means `store_dir_override_relocates_every_store_tier` is testing the right invariant.
- `phantom-check=off` (both the env and `.npmrc` forms) does **not** suppress the sidecar write, so there is no workaround short of overriding `XDG_DATA_HOME` process-wide. The new "Relocating the store" page might be worth a sentence on that, since it is the question someone hits right before finding this page.

colinhacks:
Pushed to this branch.

Resolution was anchored on the process cwd, but `.npmrc` / workspace-yaml discovery does not walk up and install anchors at `workspace_or_project_root`. From inside a workspace member the override was therefore missed: the CAS relocated, the sidecar tier stayed at the default store. Confirmed against a build of e9d2b8f29.

- Both halves route through `resolved_project_store_dir`, anchored where install anchors. `resolved_store_dir` is crate-internal again, so an embedder can only reach the anchored form.
- The `nub store path` command carried the same anchor: from a member it reported the default store while installs from that member used the root's override. Real pnpm reports the override from both.
- The Windows failure was the store-path assertion canonicalizing one side only.
- The relative-path test ran at the project root, where cwd and project root are the same directory, so it could not tell the two anchors apart. It runs from a subdirectory now.
- Added workspace-member coverage for the sidecar tier and for `store path`.
- Ran `cargo fmt`, corrected `v3/files` to `v1/files`, and merged main (149 behind).

The `phantom-check` setting gates the warn-only scan of the project's own source, not the dependency sidecars, so it is a separate feature — left undocumented rather than surface an internal store tier.

colinhacks:

PR #645, opened 2026-07-31
pm: add per-invocation minimumReleaseAge CLI flags

colinhacks:
Clearing the age gate previously required an `.npmrc` or env-var edit — global and persistent, to get past one command.

Adds `--minimum-release-age <DURATION>` and `--minimum-release-age-exclude <PKG>` to install/ci, the engine verbs, and nubx.

Each mirrors both surfaces the setting already has: pnpm 11.9.0's spellings, and the values the matching `nub.jsonc` field accepts. The duration calls the same parser as `install.minimumReleaseAge`, so file and flag cannot drift to two grammars. A bare number means minutes (pnpm's contract; `nub.jsonc` rejects the bare form on purpose). Sub-minute values round up, so tightening the gate can never silently disable it.

No strictness flag, and none in `nub.jsonc`: under Nub the gate is strict, and `--minimum-release-age=0` is the opt-out — turn the window off rather than keep one and quietly install versions that fail it. Verified 0 disables completely, including alongside `minimumReleaseAgeStrict=true` and `paranoid=true`.

Also fixes: `nub add`'s age-gate error carried no help text at all, and `install --minimum-release-age 0` misrouted `0` as a package.

### DIFF
--- a/crates/nub-cli/src/cli.rs
+++ b/crates/nub-cli/src/cli.rs
@@ -938,6 +938,12 @@ pub enum Command {
         #[arg(long = "no-check")]
         no_check: bool,
 
+        /// Per-invocation `minimumReleaseAge` overrides for the fetch path.
+        /// `nubx <just-published-tool>` is the common way to hit the age gate,
+        /// so the escape hatch belongs on this surface.
+        #[command(flatten)]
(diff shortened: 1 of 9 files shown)


colinhacks:

PR #646, opened 2026-08-01
fix(resolver): tolerate a non-integer dist.unpackedSize

colinhacks:
A registry serving `"unpackedSize": {...}` aborted the whole install with `ERR_NUB_REGISTRY_ERROR: invalid type: map, expected u64`. Packument fetches run concurrently, so the name in the message is whichever fetch lost the race — it reads as a different random package each run.

`unpacked_size` only feeds the progress bar's size estimate, so strictness cost a dead install for a wrong pixel. Now parsed through a tolerant visitor: integers kept, every other shape dropped to `None`. `integrity`/`shasum` stay strict on purpose; a note at the field records why.

Against a local registry serving that shape: v0.6.0 dies at the parse; this branch and pnpm 10 both resolve and reach the tarball fetch.


colinhacks:

PR #647, opened 2026-08-01
fix(lockfile): a declaration resolves ambiguity, whatever it names

colinhacks:
Two changes to the same error.

**Behavior.** `detect.rs`'s module docs have always promised a declaration naming the running tool "never makes a multi-lockfile project ambiguous". The code contradicted them, so a declared project with two lockfiles refused to install. That was inescapable, not just strict: `nub install` writes `devEngines.packageManager: nub` itself, so any stray second lockfile broke every later install and the suggested remedy was already applied. A self-name declaration now resolves like any other — canonical lockfile wins, else precedence. Undeclared with two lockfiles is still a hard error, unchanged.

**Message.** The ambiguity remedy no longer advises setting a declaration; declaration-mismatch keeps its wording, where the advice is correct.

Found on Cloudflare Workers Builds, which manufactures the state by running its own install (bun) beside a committed `nub.lock`.

Docs: corrects two claims that a declaration resolves ambiguity, replaces an invented mockup line with captured output, drops a false `.tool-versions` warning on the Cloudflare page, and documents `SKIP_DEPENDENCY_INSTALL`.

Caveat: resolving past a stray exposes it to parsing, so a malformed stray now fails on its own merits rather than as an identity error.

Refs #473

### DIFF
--- a/crates/nub-cli/src/pm_engine/mod.rs
+++ b/crates/nub-cli/src/pm_engine/mod.rs
@@ -1901,9 +1901,21 @@ fn resolve_identity_walk_up(
 /// `nub pm use` is the one-command fix for both states). Exit code is the
 /// generic 1 (the session-build path has no per-code exit channel); the
 /// stable code string in the output is the contract scripts can branch on.
+///
+/// The two states get DIFFERENT remedies because "set the declaration" is a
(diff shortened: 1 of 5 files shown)


colinhacks:

PR #648, opened 2026-08-01
fix(registry): attribute packument decode failures; three more cosmetic fields stop aborting installs

colinhacks:
Stacked on #646 — review the top commit only.

**Decode errors name what failed.** `I/O error: invalid type: map, expected u64` named no package and no field, and packument fetches are concurrent, so it got blamed on whichever fetch surfaced first. Now:

```
failed to decode packument @img/colour: invalid type: map, expected u64
  — near: …"shasum":"0000…","unpackedSize":{"bytes":1}},"depend…
```

The excerpt supplies the field name; serde's message never carries the key.

**`modified`, `peerDependenciesMeta.optional`, `hasInstallScript`** each still aborted a whole install; each now degrades to the conservative reading. `integrity`/`shasum` stay strict.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -352,6 +352,7 @@ dependencies = [
  "reqwest 0.13.4",
  "serde",
  "serde_json",
+ "serde_path_to_error",
  "sha2 0.11.0",
  "sonic-rs",
  "thiserror 2.0.18",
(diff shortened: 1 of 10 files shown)


colinhacks:

PR #649, opened 2026-08-01
fix(launcher): self-heal never fired under pnpm 11

colinhacks:
`leadsToUs()` scanned a cmd-shim's quoted paths with `/"([^"]+)"/g`. That class cannot match an empty pair, so pnpm 11's `exe=""` / `msys=""` header lines each consumed one quote and re-paired every following quote off-by-one. The target token was never produced, so `healPathEntry` skipped the entry.

Every `nub` call under pnpm 11 kept paying the ~50ms node hop instead of the ~1-4ms sh trampoline. pnpm 10 has no empty assignment in its template and was unaffected — which is why the matrix stayed green.

Silent perf regression, not a crash: nothing failed, the fast path just never engaged.

## Changes

- widen the class to `[^"]*`
- prefer pnpm >=11's own `# cmd-shim-target=<abs>` trailer when present — the shim declaring its own target, so it can't drift with template churn the way quote-scraping does
- add a `pnpm11` fixture style carrying both hazards, copied from a real `pnpm@11.18.0` shim, and sweep it in the matrix

## Verification

Fixture matrix: **26 ok / 5 failed** before, **31 ok / 0 failed** after. The new style fails without the launcher change.

Real `pnpm@11.18.0 add @nubjs/nub@0.6.0`, two fresh installs differing only in `launch.js`:

```
stock  before=57240cbc after=57240cbc  -> not healed
fixed  before=50a7bfa8 after=3c18b071  -> HEALED
```


colinhacks:

PR #651, opened 2026-08-01
fix(runtime): a .cjs preload must not outrun nub's hooks or run twice

colinhacks:
A `.cjs` entry in the nub.jsonc `preload` array ran before Nub's hooks were installed, and ran twice, on the compat tier (Node 18.19–22.14). Found while verifying something unrelated; no issue filed.

A preload that `require()`s a TypeScript file did not merely misbehave — it aborted the process on the unstripped annotation:

```
SyntaxError: Missing initializer in const declaration
    at Object.<anonymous> (/tmp/fixture/pre.cjs:4:17)
```

## Why

Preloads were injected by extension: `.cjs` → `--require`, everything else → `--import`. Two upstream Node behaviors make that wrong wherever Nub's own preload is an `--import`. Both were measured against plain `node` with no Nub involved, on 18.19.0 through 26.5.0:

1. Node runs **every** `--require` preload before **any** `--import` preload, whatever order the tokens appear in. Nub's compat preload is an `--import`, so a user `--require` always won the race and ran before the hooks existed.
2. Every `module.register()` loader worker **re-runs** the `--require` preloads in its own realm, and the compat tier always registers one. So the preload also ran twice — four times when user code spawned a worker of its own. The worker reads the immutable per-process option store, so scrubbing `process.env.NODE_OPTIONS` or `process.execArgv` before the `register()` call does not suppress it.

## The change

A `.cjs` preload keeps `--require` only when nothing in the invocation pulls in the ESM loader: Nub's own preload is a `--require` (fast tier) **and** no sibling entry rides `--import`. Otherwise it rides `--import`, which is ordered after Nub's token and is skipped in loader workers.

The sibling clause matters on the fast tier too. Given `["./b.mjs", "./a.cjs"]`, Node 22.15 ran `a.cjs` twice and ahead of `b.mjs`, because the `.mjs` makes the runtime auto-select the async loader-worker tier. Rerouting costs nothing there — the sibling has already forced eager ESM-loader init.

Where nothing forces it, `--require` is retained deliberately: routing a CJS preload through `--import` would itself force that init and break the synchronous `Module.runMain` semantics of the user's entry (the R1 cluster).

An oddly-cased `.CJS` stays on `--require` everywhere. Node's ESM loader maps extensions case-sensitively with no unknown-extension fallback, while `require()` loads an unrecognized extension as JS, so rerouting it would trade a preload-ordering bug for an `ERR_UNKNOWN_FILE_EXTENSION` startup crash.

## Verification

Driven end-to-end on 18.19.0, 18.20.4, 20.19.0, 22.13.0, 22.14.0, 22.15.0, 24.17.0 and 26.5.0 — one run, hooks live, declared order preserved, on both the file-run and `nub run` paths. A falsification sweep against a merge-base build also covered worker realms, a throwing preload (exit status and entry suppression unchanged), `require.cache` and `Module.prototype.require` patching, bare specifiers, and `--node` compat mode.

Two behavior notes, both accepted:

- An oddly-cased `.CJS` on the compat tier keeps the original early-and-twice bug, because rerouting it would crash. Uppercase `.MJS` was already broken the same way before this change, so uppercase-extension preloads are a pre-existing family worth its own fix.
- A rerouted `.cjs` preload sees `module.parent` as `undefined` rather than an object. The property is deprecated (DEP0144), and the fast tier is unchanged.

Coverage: five unit tests on the injection policy in `nub-core`, and a `version_tiers` integration test over a new `tests/fixtures/cjs-preload` fixture, run against both a compat-tier and a fast-tier Node.

## Third commit: a Windows-only defect the first CI run caught

The first run was red on `Test (windows-latest, node 24)` and `embed-runtime verify (windows-latest)`. The new injection function takes a `windows` parameter for purity but decided whether a spec needed the `file://` spelling with `Path::is_absolute()`, which answers for the host — so the POSIX fixture `/proj/pre.cjs`, rooted but not drive-qualified on Windows, fell through to the bare-specifier branch and emitted `--import=/proj/pre.cjs`.

Replaced with a pure `is_absolute_path(spec, windows)` beside `strip_verbatim` and `to_file_url`, matching the "pure over `windows` so both branches test on any host" pattern those already document.

Production behavior was never affected on either platform: preload entries are resolved to a platform-native absolute path before reaching this code, and the two predicates agree on those. The defect was test purity, and the coverage gap under it — nothing exercised the Windows spelling, so it was invisible from a Mac. The added test asserts the `windows: true` branch on any host, and was checked to fail against the old predicate with `--import=C:\proj\pre.mjs`.

## The first commit is unrelated

`cargo clippy --all-targets` and `cargo test` fail on `main` right now, before this branch touches anything:

```
error: couldn't read crates/nub-cli/src/../../../site/public/schema/latest.json
```

3539b65db1 withdrew that published schema to hide the nub.jsonc config reference until it ships, but a test still embedded it with `include_str!`, turning a content decision into a compile error for the whole nub-cli test target. Every CI run on main since has been cancelled, so nothing caught it. The test now reads the file at runtime and skips when it is unpublished, keeping its assertions — verified to still fail on a schema drifted from the parser's key set — whenever the file is present.

### DIFF
--- a/crates/nub-cli/src/cli.rs
+++ b/crates/nub-cli/src/cli.rs
@@ -3067,24 +3067,14 @@ pub(crate) fn runtime_node_options(
         if preload.is_empty() || preload.contains('\0') {
             bail!("nub.jsonc `preload` entries must be non-empty paths or module specifiers");
         }
-        let path = Path::new(preload);
-        let is_cjs = path
-            .extension()
-            .and_then(|extension| extension.to_str())
-            .is_some_and(|extension| extension.eq_ignore_ascii_case("cjs"));
-        let value = if path.is_absolute() && !is_cjs {
-            nub_core::node::spawn::path_to_file_url(preload)
-        } else {
-            preload.clone()
-        };
-        options.push(
-            nub_core::node::spawn::PreloadInjection {
-                flag: if is_cjs { "--require" } else { "--import" },
-                value,
-            }
-            .node_options_token(),
-        );
     }
+    // The whole list at once: whether a `.cjs` entry can keep `--require` depends on
+    // whether any SIBLING entry drags in the ESM loader. See user_preload_injections.
+    options.extend(
+        nub_core::node::spawn::user_preload_injections(&runtime.preload, &node.version)
+            .iter()
+            .map(|injection| injection.node_options_token()),
+    );
 
     if let Some(tsconfig) = runtime.tsconfig.as_deref()
         && !Path::new(tsconfig).is_file()
--- a/crates/nub-cli/src/project_config.rs
+++ b/crates/nub-cli/src/project_config.rs
@@ -1926,9 +1926,22 @@ mod tests {
             values.iter().map(|value| (*value).to_string()).collect()
         }
 
-        let schema: Value =
-            serde_json::from_str(include_str!("../../../site/public/schema/latest.json"))
-                .expect("the published nub.json schema must be valid JSON");
+        // Read at RUNTIME, not via include_str!. The schema is a PUBLISHED site
+        // artifact, and the site withdraws it whenever the config reference is
+        // hidden pending ship (3539b65db1) — a compile-time include turns that
+        // ordinary content decision into a build failure for the whole crate's test
+        // target. Absent schema means there is nothing published to be out of step
+        // with, so skip; the assertions return with the file.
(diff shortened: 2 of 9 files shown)


colinhacks:

PR #70, opened 2026-06-22
site: align heading inline-code baseline + strengthen PM security copy

colinhacks:
Two homepage changes in one PR (distinct regions of `site/src/app/(home)/page.tsx`).

## Heading inline-code baseline (measured)

`HeadingCode` (inline monospace inside serif headings, e.g. "A 24× faster `pnpm run`") rendered with its glyph bottoms below the serif baseline. Measured via headless Chrome / CDP `getBoundingClientRect`:

| Heading | before (code − serif) | after |
| --- | --- | --- |
| h2 (40px serif) | +2.13px | −0.09px |
| h3 (25px serif) | +1.27px | −0.13px |

Fix: `vertical-align` `-0.09em` → `-0.035em` on `HeadingCode`. Residual is sub-pixel (±0.13px) across every instance (`using`, `node`, `pnpm run`, `npx`, `pnpm exec`, `pnpm`), verified at both heading sizes. Vercel preview gives the visual confirmation.

## PM band security

The PM band under-sold the supply-chain posture. Now leads with the defenses that are **on by default**, each grounded in code:

- Deny-by-default build scripts (`WARN_NUB_IGNORED_BUILD_SCRIPTS`, `nub approve-builds`)
- OSV malicious-package advisory check, hard-block on a hit (`ERR_NUB_MALICIOUS_PACKAGE`)
- Trust-downgrade refusal (`trustPolicy=no-downgrade`, `ERR_NUB_TRUST_DOWNGRADE`)
- `minimumReleaseAge` cooling window (24h default, matching pnpm)

Terminal demos use real output: the `@ledgerhq/connect-kit` / `MAL-2023-8697` block (real OSV advisory), the `nanoid@3.3.14` trust-downgrade in the engine's exact format, and the captured build-script warning. Error/refusal lines render in ember-red with ❌ via a new `tone` on the `Terminal` component. README PM section aligned to match.

Deliberately NOT claimed (verified against code): install-time cryptographic attestation verification (only metadata-shape presence checks exist), default-on build sandboxing (opt-in via `paranoid`), and install-failing-by-default for the age gate (default is soft fallback).

tsc clean, `next build` green.

https://claude.ai/code/session_01YRvztkcr4fzfg9rD5edUwa

PR #658, opened 2026-08-02
fix(add): stop the similar-name gate refusing popular packages

colinhacks:
Three failures, one root cause chain. The nightly lockfile-mutation differential has been red since 2026-07-30, and every `Build` leg of the release now dies in `build.rs`. Both trace to the popular-name similarity gate that arrived with the aube v1.35.0 sync (#621).

## The gate refuses popular packages

`find_similar_package_name` skipped only an exact self-match and kept scanning, so being *in* the popularity corpus was no defence. Measured against the real 100,000-name corpus, that rejects almost everything worth installing — and the name it points at is frequently the squatter:

| `nub add …` | before | after |
| --- | --- | --- |
| `react` | refused, "did you mean preact?" | ok |
| `lodash` | refused, "did you mean loadash?" | ok |
| `express` | refused, "did you mean expres?" | ok |
| `debug` | refused, "did you mean dbug?" | ok |
| `axios` | refused, "did you mean gaxios?" | ok |
| `ms` | refused, "did you mean ws?" | ok |
| `lodahs` `raect` `axois` | refused, correct suggestion | refused, correct suggestion |

A package among the 100,000 most-downloaded on npm is not a typosquat, so a request the corpus lists at all is now cleared outright instead of being compared against the rest of it. A name the corpus does not list is still scanned against all of it.

The blunt form of that rule is what the precision requires, and the intermediate version is worth recording because it looks sufficient and is not. Restricting candidates to names that **outrank** the request fixes every row above — but still refuses **847 of the top 10,000**, measured by running the gate over the corpus itself:

| rule | false positives in the top 10,000 |
| --- | --- |
| compare against the whole corpus | substantially everything popular |
| candidates must outrank the request | 847 (8.5%) — `ws`→`ms`, `qs`→`ms`, `micromatch`→`picomatch`, `safer-buffer`→`safe-buffer` |
| clear any request the corpus lists | **0** |

Those are all legitimate packages, and this gate hard-fails a non-interactive install. What the strict rule gives up is a squat popular enough to be in the corpus itself (`expres`, #58196) — one with real download volume, which is the OSV advisory check's and the downloads floor's territory rather than a spelling heuristic's.

## The corpus was not a ranking

The gate reads absence from the corpus as "nobody installs this", which is only true of the real 100,000-name list. When `popular-top100000-v1.json` is absent, `build.rs` falls back to the metadata-primer names — which `write_package_blob` derives from a `BTreeMap`, alphabetically, and which number 100 in dev and 2000 in release. There, absence means nothing, and the reported "top-100,000 popularity rank #N" is an alphabetical position.

That is how it came to refuse `nub add ms@2.1.3` in the nightly mutation harness: in the primer's top 100 sorted alphabetically, `qs` sits at position 70, which is verbatim the "popularity rank #70" in the failing run. Now `build.rs` reports whether the corpus is genuinely download-ranked, and the gate declines to run when it is not.

## The corpus was never shipped

Release builds set `AUBE_REQUIRE_PRIMER=1`, under which `build.rs` treats the corpus as mandatory — but nothing produced it. Upstream aube's release workflow passes `--popular-names-out`; nub's copy was not updated in the sync, so four `Build` legs of run [30727744051](https://github.com/nubjs/nub/actions/runs/30727744051) died with:

```
thread 'main' panicked at vendor/aube/crates/aube-resolver/build.rs:262:13:
popular package names are required, but …/data/popular-top100000-v1.json was missing
```

This was latent until the version-check fix in fdd18fe46e unblocked the `verify` gate that had been failing ahead of it. `scripts/remote-build.ts` already records hitting the same panic.

`mac-build.yml` sets the same flag and is broken identically. Having Node on the runner is not enough: `build.rs` regenerates the primer by shelling out to `generate-primer.mjs` **without** `--popular-names-out`, so nothing there ever writes the corpus. Both workflows now generate it explicitly.

Shipping the corpus is what makes the gate live for released binaries for the first time — which is why it lands together with the two fixes above rather than after them.

## Verification

**Against a binary built from this branch with the unranked corpus, which is what `cargo build -p nub-cli` produces in CI:** `ms react lodash express debug axios chalk` all install, and the exact failing fixture `nub add ms@2.1.3` exits 0. `lodahs reactt expresss` are still refused — by the weekly-downloads floor, not the similarity gate. The gates are defence in depth, so declining to run this one on an unranked build does not leave squatters unprotected.

**Against a binary built with the ranked corpus embedded, so the gate is live:** the same popular packages install, and `lodahs reactt expresss` are refused by the similarity gate itself with the correct suggestion. Both false-positive counts in the table above are measured by running each rule over the corpus.

Also verified: the new corpus step run against the live source writes exactly 100,000 rank-ordered names; `option_env!` picks up a conditional `cargo:rustc-env` and leaves no stale value when the flag is removed; both workflow files parse.

Unit tests cover the exemption, a lookalike appearing earlier in the corpus than the request it must not refuse, the unlisted-request scan, and the ranked-flag invariant.

### DIFF
--- a/.github/workflows/mac-build.yml
+++ b/.github/workflows/mac-build.yml
@@ -86,6 +86,23 @@ jobs:
           node-version: 26
       - run: npm install --no-audit --no-fund --loglevel=error
 
+      # The typosquat corpus is a SEPARATE artifact from the primer, and build.rs
+      # treats it as mandatory under AUBE_REQUIRE_PRIMER=1 (set on the build below).
+      # Having node is not enough: build.rs regenerates the primer by shelling out
+      # to generate-primer.mjs WITHOUT --popular-names-out, so nothing else on this
+      # runner ever writes the corpus and the build dies in write_popular_names_blob
+      # with "popular package names are required". `--top 1` is the cheapest way to
+      # reach the script's popular-names branch; the primer output is discarded
+      # outside data/ so build.rs still regenerates the real one itself.
+      - name: Generate typosquat popularity corpus
+        run: |
+          set -euo pipefail
+          cd vendor/aube
+          node scripts/generate-primer.mjs \
+            --top 1 \
+            --out "$RUNNER_TEMP/primer-discard.json" \
+            --popular-names-out crates/aube-resolver/data/popular-top100000-v1.json
+
       # Belt and braces against the silent-primer path above.
       - name: Build the CLI
         env:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -345,18 +345,54 @@ jobs:
           fi
           echo "primer $out is $bytes bytes"
           echo "path=vendor/aube/$out" >> "$GITHUB_OUTPUT"
-      # Upload whichever source won. Exactly ONE primer-* file exists in data/ at
-      # this point (smart .rkyv.zst XOR public .rkyv.json), both named for the
-      # build-expected TOP/CAP/schema; the glob captures it. The downstream build
-      # job downloads this artifact into the same data/ dir, so build.rs picks it up
-      # directly. `if-no-files-found: error` keeps the empty-primer guard: a missing
-      # file here (both sources failed) fails the release loud rather than shipping
-      # an empty primer.
+      # The typosquat corpus, which is a SEPARATE artifact from the primer and is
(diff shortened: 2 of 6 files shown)


colinhacks:

PR #659, opened 2026-08-02
env: stand down from .env loading when a .env.schema owns the environment

colinhacks:
A project with an `@env-spec` schema gets a different environment from `nub app.ts` than from `varlock run -- node app.ts`, and neither reports a problem.

Nub's cascade keys on `APP_ENV`. A schema picks its own selector with `@currentEnv`, so a schema declaring `@currentEnv=$MY_STAGE` resolves against a variable Nub never reads. Measured on a fixture with `MY_STAGE=production`:

| | `GREETING` | `PORT` |
| --- | --- | --- |
| `varlock run -- node app.js` | `hello-from-production` | `8080` |
| `nub app.js` (before this PR) | `hello-from-dotenv` | `null` |

A schema is a graph, not a file list. It resolves through providers, `exec()`, imports and a selector of its own choosing, so any second implementation of "which files win" is a guess that silently disagrees.

## What changed

When Nub finds a `.env.schema` and Varlock is installed, it stops loading `.env*` and spawns Varlock in front of Node:

```
nub app.ts   →   varlock run --path <schema-dir> -- <node> --import=preload.mjs app.ts
```

Nub resolves nothing, injects nothing, and redacts nothing. Varlock owns the environment end to end, exactly as it does when a user invokes it themselves.

- **`crates/nub-core/src/node/spawn.rs`** — new `SpawnConfig::env_owner`, as `(cli, schema_dir)`. When set, `spawn_node` builds the same Node command and puts the loader in front of it. The `arg0("node")` rewrite is skipped on that path.
- **`crates/nub-cli/src/env_owner.rs`** — detection only: find the schema, find the loader in `node_modules/.bin` walking up, else `PATH`.
- **`crates/nub-cli/src/cli.rs`** — the loader-owned branch contributes an empty environment.
- **`__NUB_ENV_OWNER_WRAPPED`** marks the wrap. Varlock's bin is a `#!/usr/bin/env node` script, so its own interpreter resolves through Nub's PATH shim and re-enters Nub; without the marker that recursion does not terminate.

`--path` is load-bearing rather than cosmetic. Varlock takes its entry point from cwd, which is not where the schema is for a workspace member or a run from a subdirectory. Nub already walked up to find the schema before deciding to stand down, so it hands over the directory it walked to; without that a member died with `No .env files found in …/pkgs/web` against a schema Nub had just found at the root.

`nub run`, `nubx` and lifecycle scripts need no wrapping code — their `node` goes through Nub's PATH shim, re-enters Nub, and is wrapped there. Verified with a `#!/usr/bin/env node` bin run through `nub run`: schema values arrive and secrets are redacted.

`nub watch` is the exception and does need its own wrap: it assembles its own `node --watch` command rather than going through `spawn_node`, while detection has already stood Nub's cascade down. Without it every variable arrived `undefined`, silently. It now has a regression test, having produced that same failure twice.

## What this replaces

An earlier revision of this PR loaded Varlock in-process from a Nub-owned preload module. That is dropped entirely, and with it two runtime adapters, the graph parser, the install-shape split, and the marker plumbing across five launch paths. Net **−1114 / +265** lines against the previous head.

The in-process route could not redact, which is the part that matters. Redaction needs Varlock either inside the process or in front of it, and only the second covers everything (measured, piped output):

| | `console.log` | `process.stdout.write` | subprocess |
| --- | --- | --- | --- |
| in-process patches | yes | no | no |
| `varlock run` | yes | yes | yes |

## Behavior when Varlock is absent

Nub keeps loading `.env*` and warns once. It does not stand down on the filename alone: [dotenv-extended](https://www.npmjs.com/package/dotenv-extended) has defaulted to `.env.schema` for an incompatible format since 2016, so Nub reads the file and stays silent unless it is actually `@env-spec`. An explicit `envFile`/`--env-file` still wins, and `--node`/`NODE_COMPAT=1` disable the whole path.

## Verification

Ten integration tests against a stub loader, plus a sweep against real Varlock 1.16.0 covering the file run, `nub run`, a workspace member, a subdirectory run, `--node`, and a project with no schema:

- redaction confirmed on all three channels (`console`, `process.stdout.write`, subprocess)
- TypeScript still transpiles through the wrap
- `@currentEnv=$MY_STAGE` resolves both ways — `production` gives `8080`, `development` gives the schema default
- `@generateTypes` writes `env.d.ts`
- a validation failure exits 1 without running the program
- `process.exit(7)` arrives as 7

The stub loader carries a `#!/usr/bin/env node` shebang deliberately. With `#!/bin/sh` the recursion channel never opens, and the re-entrancy test passes against a Nub whose guard has been removed. Falsification control: with the guard defeated, two tests hang and then fail.

`cargo fmt --check`, `cargo clippy --all-targets --all-features -- -D warnings`, and the `nub-cli` and `nub-core` suites are green.


colinhacks:

PR #663, opened 2026-08-02
fix(core): never trust a config-cache entry read inside the file's own mtime tick

colinhacks:
Fixes the Windows-only flake in `pnpm_v11_surfaces_need_a_declared_v11_incumbent`, which reds CI on `main`.

`MtimeCache` and the `detect_project` memo validate entries on a `(mtime, size)` stamp. File mtimes are quantized to the platform timer tick (0.39-1.4 ms measured on Windows Server 2022/NTFS), so two same-size writes separated only by a read collide on mtime 62-84% of the time and the stale value is served. The test rewrites `pnpm@10.0.0` to `pnpm@11.0.0` — 32 bytes either way.

Both caches now require a two-second granularity slop between a file's mtime and the read that cached it, proving the mtime's quantum had closed. Measured: 0/2000 just-written files trusted, 2000/2000 aged files still hit.

The second commit corrects the first, which was a no-op.

### DIFF
--- a/crates/nub-core/src/config_cache.rs
+++ b/crates/nub-core/src/config_cache.rs
@@ -15,17 +15,24 @@
 //! call-ordering analysis: a cache validated on mtime can never serve a value
 //! older than the file on disk.
 //!
-//! That argument only closes once the RACY window is handled: mtime resolution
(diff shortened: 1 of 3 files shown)


colinhacks:

PR #665, opened 2026-08-02
feat(dist): ship one binary per platform package, carry the verb in __NUB_ARGV0

colinhacks:
Platform packages shipped `bin/nub` and `bin/nubx` byte-identical at 45 MB so the CLI could read its verb off argv[0]. Six of eight exceeded cnpm/npmmirror's 80 MiB cap, so that mirror is pinned at 0.0.31 and installs behind it fail. Now 90.3 → ~45 MiB; release archive 49.0 → ~24.5 MB.

The second name cannot ship: `npm pack` drops symlinks, hardlinks are mangled on install, `exec -a` is a bashism dash rejects.

Deleting the duplicate alone fails silently — call 2 runs `nub`, exit 0. So the trampoline and the Node spawn both set the var; the latter is what makes Windows work. argv[0] stays the fallback; the var is erased at startup.

Verified on ubuntu + windows.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -708,13 +708,16 @@ jobs:
           MATRIX_PLATFORM: ${{ matrix.platform }}
         run: |
           # The runtime is EMBEDDED in the binary (staged + built above), so the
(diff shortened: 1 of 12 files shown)


colinhacks:

PR #666, opened 2026-08-02
fix(install): bootstrap node-gyp lazily, not before every build fan-out

colinhacks:
Any approved build bootstrapped node-gyp up front, needed or not — so an unreachable registry aborted the install after ~70s, with a warm store and nothing invoking node-gyp. The tool dir is in the cache, not the store, so offline CI installs hit it every time.

Routes install through the lazy shim `nub run` already uses. `ensure_cached` still locks the tool dir, so a parallel fan-out converges on one bootstrap.

`jailBuilds=true` is the exception and resolves up front: a jailed script gets a cleared env and a temporary HOME, so a shim re-entry looks for the tool dir under that HOME and cannot refetch (the jail denies network). Best-effort — it warns rather than failing an install that never needed node-gyp.

Also: `AUBE_NODE_GYP_PROJECT_DIR` was never set on the install path, and the sh shim read it under `set -eu`; the cmd shim expanded it literally. Both fall back to cwd now.

Verified on a 13-scenario probe (`probe/node-gyp-shim`) across Windows, Linux and macOS: real addon compiles, PATH precedence, `npm_config_node_gyp`, concurrency, and jailed builds.

### DIFF
--- a/crates/nub-cli/tests/install_engine.rs
+++ b/crates/nub-cli/tests/install_engine.rs
@@ -1224,16 +1224,23 @@ fn approved_build_that_never_calls_node_gyp_installs_with_no_registry() {
 /// resolved up front — but best-effort, so failing to reach a registry cannot
 /// sink an install whose builds never wanted node-gyp.
 ///
-/// Hermetic: the dep is a `file:` dep and the registry is deliberately dead, so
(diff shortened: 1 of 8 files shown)


colinhacks:

PR #668, opened 2026-08-03
fix(registry): write the packument cache with serde_json, not sonic-rs

colinhacks:
Both packument-cache writers serialize through `serde_json::to_vec` instead of `sonic_rs::to_vec`; read path unchanged.

A `serde_json::Value` may only be serialized by serde_json. Under `arbitrary_precision`, sonic-rs writes `Number`'s private token literally, so cached numbers land as `{"$serde_json::private::Number":"…"}` and the next typed read fails `invalid type: map, expected u64`. Revalidation rewrites the poison verbatim, so it never heals.

`nub compile` pulls `rolldown_common`, which requires that feature; Cargo unifies per build and the cache is shared. No released binary can produce it. Also silently corrupts `trustedPublisher`/`approver`.

fmt, clippy, 309 tests pass.

### DIFF
--- a/vendor/aube/crates/aube-registry/Cargo.toml
+++ b/vendor/aube/crates/aube-registry/Cargo.toml
@@ -48,6 +48,9 @@ hickory-dns = ["reqwest/hickory-dns"]
 
 [dev-dependencies]
 tempfile = "3"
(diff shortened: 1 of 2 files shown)


colinhacks:

PR #673, opened 2026-08-05
fix(preload): one NODE_OPTIONS token per flag name, and resolve bare preload entries from the CWD

colinhacks:
Three fixes to how nub delivers preloads through `NODE_OPTIONS`, plus one npm-parity gap found alongside them.

## The bug

Any consumer that re-parses `NODE_OPTIONS` destroys repeated same-name flags. Next.js parses it into a `Record` keyed by option name and reformats it for every forked worker, so `--require=a --require=b` collapses to `--require=b`. nub emitted one token per `nub.jsonc` `preload` entry on top of its own, so a single `.cjs` preload entry silently dropped nub's own preload — and with it the entire augmentation layer.

Measured on a real `next@16.3.0` build, counting the processes nub's augmentation actually reached:

| | `--require` tokens | nub-augmented processes |
|---|---|---|
| before | 2 | **3 of 8** |
| after | 1 | **8 of 8** |

Exit 0 in both cases. Filed upstream as [vercel/next.js#96582](https://github.com/vercel/next.js/issues/96582); the analysis is also on the existing [#77550](https://github.com/vercel/next.js/issues/77550).

## What changed

**One token per flag name.** nub writes a single chainer module that loads the user's entries in declared order, instead of one token each. A `.cjs`-only list rides nub's own `--require` preload, so no second token exists and `--require`'s synchronous entry semantics survive. Anything else gets a `.mjs` chainer on its own `--import`, which also preserves top-level await — `require()` cannot load a TLA module at all.

**Bare entries resolve from the CWD.** Loading entries through a generated file changed where a bare specifier resolves from: the file's directory rather than the CWD. In a workspace where a member shadows a root dependency, `cd packages/web && nub app.js` loaded the ROOT copy where plain Node loads the member's — same specifier, different module, no error. nub now resolves bare entries with `oxc_resolver` using the condition set the channel implies, so the anchor is a decision rather than a side effect of where the file was written. This resolution is cold and pre-V8, so none of the JS-runtime inputs to Node's algorithm exist yet; nub's runtime hooks still delegate every in-process resolution to Node.

**`nub run` honors `node-options`.** npm and pnpm both apply the `node-options` npmrc field to script execution. nub honored it only for lifecycle scripts during PM operations, so `nub run` dropped it — a project raising its heap ceiling for a build would OOM under nub and succeed under pnpm. It is appended to nub's augmentation rather than assigned over it, so nub's preload survives where npm's own handling destroys the ambient value.

## Verification

- **204-cell matrix, all passing**: 3 Node versions (22.14.0 compat tier, 22.15.0 fast-tier floor, 26.5.0) × `{CJS, type: module}` × 17 preload combinations × `{.js, .mjs}` entry. Each cell asserts every entry ran exactly once, in declared order, before the entry, with at most one `--require` and one `--import`, and nub's own token present.
- **Merge-base differential**: an earlier run of the same matrix against the merge-base showed identical execution behavior in every cell — only the token shape changed.
- **Positive controls**: the Next.js failure reproduces on the merge-base and not after; the anchor regression test was confirmed to fail (`ROOT` instead of `MEMBER`) with the fix removed.
- **Edges**: missing preload, throwing preload, `--node` compat mode, cwd outside the project root, no `node_modules`, `nub run`, worker threads, six concurrent runs, and a real Yarn PnP project.
- `clippy --all-targets --all-features` and `fmt --check` clean.

`verify_deps.rs` has three failures on this host (`npmrc_error_aborts_before_running`, `fresh_clone_warns_but_still_runs_the_script`, `version_drift_warns_and_names_the_dependency`). They reproduce identically on the merge-base and CI is green on `main`, so they are environmental rather than introduced here.

## Rough edges, deliberately left

- The chainer is written to `node_modules/.nub/`. Now that entries resolve to absolute paths it no longer needs to live inside the project, so it could move to nub's own cache dir — including for Yarn PnP projects, where nub currently creates a `node_modules` directory that would otherwise not exist.
- An unresolvable bare entry produces Node's own `ERR_MODULE_NOT_FOUND` naming the chainer rather than `nub.jsonc`. The generated file's header comment names its source, so diagnosis is one step.
- Under Yarn PnP, `oxc_resolver` is built without the `yarn_pnp` feature and cannot resolve; the entry passes through and PnP's own resolver handles it. Verified end-to-end, but those entries do not get the CWD anchor.
- `user_preload_injections` in `spawn.rs` and its tests are now unused. They are marked superseded in place rather than deleted, since they still document the two measured Node facts that govern which channel the chainer rides.


colinhacks:

PR #674, opened 2026-08-05
fix(aube): resolve workspace:<path> specifiers, and say where a patch hunk failed

colinhacks:
Two package-manager fixes found while porting a bun-authored workspace. Unrelated to each other beyond where they surfaced, so one commit each.

**`workspace:<path>` was compared as a version range.** bun writes a dependency as `<name>@workspace:packages/plugin`, where the tail is the member's directory. nub compared that tail against the member's version as if it were semver, so `nub install` failed on any bun-authored workspace:

```
no version of @opencode-ai/plugin matches range 'workspace:packages/plugin'
```

`try_workspace_link` now treats a tail that looks like a path — leading `.`, leading `/`, or containing `/` — as a directory reference, alongside the existing empty/`*`/`^`/`~` forms. A tail that is a real range is still compared as one.

**A failed patch hunk named neither where it looked nor what to do.** It now reports the hunk index, the line it searched from, that it searched 20 lines either way, and that the patch needs regenerating against this version. The ±20 tolerance is unchanged — it matches pnpm.

Both were found on `compile-spike` while porting opencode; neither has anything to do with `nub compile`, so they are extracted here rather than waiting on #536. Refs #536.

Tests: 303 `aube-resolver`, 165 `aube-linker`, green on top of main; new test `a_workspace_tail_that_is_a_path_is_not_a_version_range`. `cargo fmt --check` and `clippy --all-targets -D warnings` green.

### DIFF
--- a/vendor/aube/crates/aube-linker/src/patches.rs
+++ b/vendor/aube/crates/aube-linker/src/patches.rs
@@ -535,8 +535,17 @@ fn apply_hunks(base_image: &str, hunks: &[Hunk]) -> Result<String, String> {
                 -fuzzing_offset - 1
             };
             if fuzzing_offset.abs() > 20 {
+                // Refusing is deliberate — see this function's doc comment. The
+                // message says WHERE, because the actionable fact is that the
+                // patch was cut against a different release of the package: bun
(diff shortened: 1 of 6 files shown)


colinhacks:

PR #675, opened 2026-08-05
fix(preload): fold inherited NODE_OPTIONS preloads into the chainer

colinhacks:
Follow-up to #673, which bounded nub's own preload tokens to one per flag name. A preload arriving through an **inherited** `NODE_OPTIONS` still added another.

## The bug

A consumer that re-parses `NODE_OPTIONS` keys it by flag name and keeps one value per key. nub appends the inherited value last, so the token that got dropped was nub's own preload — and with it the whole augmentation layer.

That inherited `--require` is what every APM injector sets (`NODE_OPTIONS="--require dd-trace/init"` is the Datadog and OpenTelemetry k8s shape), so a Next.js app under nub with an agent attached lost its augmentation silently.

Measured on a real `next build` with an ambient `--require` plus a `nub.jsonc` preload, counting processes each preload reached and how many kept nub's augmentation:

| | ambient preload | config preload | nub-augmented |
|---|---|---|---|
| before | 8 procs | 8 procs | **3 of 8** |
| after | 8 procs | 8 procs | **8 of 8** |

## What changed

Inherited `--require` / `--import` values are folded into the same chainers as the config's, and stripped from the forwarded value — so the child sees at most one of each name however many arrive.

Only those two names are folded, because they are the only ones nub emits and therefore the only ones that can collide. Left in place:

- `--loader` / `--experimental-loader` — folding one into an `import` would drop its hook registration. A loader is not an import.
- Yarn PnP — its resolver has to install before nub's preload.
- **nub's own tokens** — a nested nub inherits them, and folding them would defeat the re-entrancy detection that keys on exactly that token.

## Four bugs surfaced while building it

Each was caught by the sweep rather than by reading the code, and each is worth knowing:

1. **A `--require` token re-runs inside the loader worker.** Node re-runs every `--require` preload in a `module.register` worker's realm, which the compat tier always has. Inherited requires ride the ESM chain there instead — the same trade the original per-entry router made for config `.cjs` entries: costs ordering, buys running exactly once.
2. **Static `import` breaks ordering under top-level await.** Node awaits each `--import` token before starting the next, but static sibling imports may interleave, so an entry with TLA let a later one land first. The chainer now emits sequential `await import(...)`.
3. **Entries were silently dropped on the async tier.** The chain was only loaded in `preload.cjs`'s fast-tier branch, so whenever the async tier was selected every entry vanished.
4. **nub forced itself onto the async tier.** `foreignAsyncLoaderFlagPresent()` matched any `--import` in `NODE_OPTIONS` — including nub's own chainer — so nub spawned a loader worker against itself and, via (1), double-ran the CJS chain. This was the root cause of the whole cluster; fixing it took the sweep from 153 failures to 29.

Separately, inherited **relative** entries now resolve from the CWD. Only an absolute spec is already anchored; config entries arrive pre-absolutized, but an inherited `./x.mjs` left verbatim resolved against the chainer's directory. Caught by the existing `import_cjs_require_cache` suite.

## Verification

- **1020-cell sweep, 991 passing.** 3 Node versions (22.14 compat, 22.15 fast floor, 26.5) × `{CJS, type: module}` × 17 preload combinations × `{.js, .mjs}` entry × 5 inherited shapes (none / require / import / both / loader). Each cell asserts every entry ran exactly once, in declared order, before the entry, with at most one `--require` and one `--import`.
- The 29 remaining failures are all the inherited-`--loader` shape and **reproduce identically on `main`** (`ERR_INVALID_RETURN_PROPERTY_VALUE`, config preload double-runs) — pre-existing, untouched here.
- **Ordering preserved exactly.** All five combinations measured against `main` before the change produce byte-identical order after it.
- The new regression test was confirmed to fail without the fold ("got 2 `--require` and 0 `--import`").
- Full `cargo test -p nub-cli`: 38 suites green. The 3 `verify_deps` failures are the known pre-existing ones that reproduce on the merge base. `nub-core` spawn 53/53. `clippy --all-targets --all-features` and `fmt --check` clean.

### DIFF
--- a/crates/nub-cli/src/cli.rs
+++ b/crates/nub-cli/src/cli.rs
@@ -3172,36 +3172,103 @@ fn prepare_preload_chain(
     let Some(root) = runtime.preload_root.clone() else {
         return Ok(None);
     };
-    if runtime.preload.is_empty() {
+    // Preload entries reaching nub through an INHERITED NODE_OPTIONS are folded into
+    // the same chainers as the config's, so the child ends up with at most one
+    // `--require` and one `--import` no matter how many arrive. Without this an
+    // ambient `--require` (what every APM injector sets) adds a second token, and a
+    // consumer that keys NODE_OPTIONS by flag name drops one of them — which, since
+    // nub appends the inherited value last, was nub's own preload.
+    let inherited = std::env::var("NODE_OPTIONS").unwrap_or_default();
+    let (_, inherited_requires, inherited_imports) =
+        nub_core::node::spawn::split_inherited_preloads(&inherited);
+    if runtime.preload.is_empty() && inherited_requires.is_empty() && inherited_imports.is_empty() {
         return Ok(None);
     }
-    let esm = !node.version.supports_augmentation()
+
+    // Which channel the CONFIG entries ride. Unchanged from the per-entry router it
+    // replaced: a `.cjs`-only list keeps `--require`'s synchronous entry semantics,
+    // anything else needs the async channel (and the compat tier is always async).
+    let config_is_esm = !node.version.supports_augmentation()
         || runtime.preload.iter().any(|spec| {
             !std::path::Path::new(spec)
                 .extension()
                 .is_some_and(|e| e.eq_ignore_ascii_case("cjs"))
         });
 
-    let dir = root.join("node_modules").join(".nub");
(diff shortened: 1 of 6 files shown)


colinhacks:
Shipped in v0.7.1: https://github.com/nubjs/nub/releases/tag/v0.7.1

PR #678, opened 2026-08-05
fix(homebrew): install bin/nub and create the nubx alias

### LINKED ISSUE #676: v0.7.0 release archive missing bin/nubx breaks Homebrew upgrade (Errno::ENOENT)
## Bug

`brew upgrade nubjs/tap/nub` (0.6.0 -> 0.7.0) fails on macOS arm64:

```
Error: An exception occurred within a child process:
  Errno::ENOENT: No such file or directory - bin/nubx
```

## Root cause

`Formula/nub.rb` installs both binaries:

```ruby
bin.install "bin/nub", "bin/nubx"
```

but the v0.7.0 darwin-arm64 release archive ships only:

```
bin/
bin/nub
runtime/
```

The v0.6.0 archive contained `bin/nubx` as well, so the formula fails staging the new version.

## Expected

`brew upgrade nubjs/tap/nub` succeeds, or the formula stops referencing `bin/nubx` since the archive no longer includes it (nubx is the same binary selected by argv[0]).

## Environment vars

- macOS arm64 (Apple Silicon), Homebrew
- nub 0.6.0 installed, upgrade to 0.7.0 fails

colinhacks:
Both `brew install` and `brew upgrade` of 0.7.0 fail with `Errno::ENOENT: No such file or directory - bin/nubx`, on macOS and Linux. The 0.7.0 archives ship one binary (#665) and the verb comes from argv[0], so `nubx` is an alias the installer creates; the generator still installed `bin/nubx`, and nothing in CI read the formula it wrote.

- Generator: install `bin/nub`, symlink `nubx`, assert dispatch in `test do`.
- Release workflow: tap-bump installs the formula before pushing it.
- Installers: refresh the `~/.nub/shims` hardlinks, stale once the binary is replaced.
- Windows `nub upgrade`: install the staged `busybox.exe`.

Closes #676

### DIFF
--- a/.claude/skills/ci-triage/SKILL.md
+++ b/.claude/skills/ci-triage/SKILL.md
@@ -1,0 +1,107 @@
+---
+name: ci-triage
+description: >-
+  Find and diagnose RED CI with the gh CLI — answer "is anything failing?" and
+  "why?" correctly, without being handed a URL. Invoke (via the Skill tool)
+  whenever you need to check whether a branch, commit, or PR is green, whenever
+  someone reports failing CI, before cutting a release, and before claiming any
+  commit is green. THE FAILURE THIS SKILL EXISTS TO PREVENT: `gh run list`
+  reports a RUN-level `conclusion`, and a run whose rollup is `cancelled` can
+  contain a job whose conclusion is `failure` — so filtering runs on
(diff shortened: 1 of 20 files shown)


colinhacks:
Shipped in v0.7.1: https://github.com/nubjs/nub/releases/tag/v0.7.1

Issue #677, opened 2026-08-05
Setting supportedArchitectures from PNPM missing


crypto-scythe:
On PNPM you can install the optional dependencies of multiple architectures / operating systems together. For reference: "[supportedArchitectures](https://pnpm.io/settings/dependency-resolution#supportedarchitectures)".

It seems currently nub does not adhere to this configuration and only installs the dependency for the current architecture / operating system.

This breaks some behaviour like for example when you are developing on a macOS host system and need to lint / compile / generate during commit but the actual application needs to run in the container.

I think at least nub could behave similar to this and at best I think this functionality is a nice addition for DX.

colinhacks:
Checked against 0.7.0 on macOS and Linux, versus pnpm 10.15.1: `.npmrc`, `pnpm-workspace.yaml`, and the `package.json` `pnpm.supportedArchitectures` object are all honored, with or without a `pnpm-lock.yaml`, under `nub install` and `nub ci`. On `esbuild@0.25.10` Nub installs the same variants pnpm does.


```ini
supportedArchitectures={"os":["current","linux"],"cpu":["current","x64"]}
```

If you don't have a `pnpm-lock.yaml` then Nub ignores `package.json#pnpm` intentionally.

colinhacks:
We landed flags for this instead in #683: `nub install --os linux --cpu x64`, also on `add` and `nubx`. 

Not released yet; it goes out in the next version.

colinhacks:
Fixed in v0.7.2 (now published): https://github.com/nubjs/nub/releases/tag/v0.7.2

PR #679, opened 2026-08-05
fix(install): host-filter the prewarm graph under the isolated layout too

colinhacks:
The virtual-store prewarm host-filtered its graph clone only on the global-virtual-store path. The isolated path returned early with the widened graph, writing one nested symlink per declared optional edge while only the filtered variants were ever materialized — 25 of esbuild's 26 `@esbuild/*` left dangling on a cold install. A set `CI` auto-selects that layout, so it is the shape CI shipped. The lockfile-reuse path was unaffected.

Hoisting the filter above the layout split makes both paths materialize the same host-only graph. Honored variant sets are unchanged.

Verified on macOS with `CI=1`: 26 nested / 25 dangling to 1 / 0 plain, and 26 / 22 to 4 / 0 with `supportedArchitectures`. The GVS and lockfile-reuse paths are unchanged, and the regression test fails without the fix.

Refs #677

### DIFF
--- a/crates/nub-cli/tests/install_engine.rs
+++ b/crates/nub-cli/tests/install_engine.rs
@@ -1307,3 +1307,56 @@ fn jailed_build_that_never_needs_node_gyp_survives_a_failed_bootstrap() {
         "the approved build script must still have run: {stderr}"
     );
 }
+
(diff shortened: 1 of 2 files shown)


colinhacks:

PR #680, opened 2026-08-05
fix(pm): abort on npm's legacy-bundling, not just install-strategy=nested

colinhacks:
The npm config definition flattens `legacy-bundling=true` to `install-strategy=nested` before anything reads it, so the two spellings request the same nested tree ([`workspaces/config/lib/definitions/definitions.js`](https://github.com/npm/cli/blob/latest/workspaces/config/lib/definitions/definitions.js)):

```js
flatten (key, obj, flatOptions) {
  if (obj[key]) {
    obj['install-strategy'] = 'nested'
    flatOptions.installStrategy = 'nested'
  }
}
```

The unsupported-config scan matched only the modern key, so an npm-incumbent project with `legacy-bundling=true` in `.npmrc` passed the scan and installed a divergent tree with no diagnostic — the outcome the scan exists to prevent.

This adds a fatal arm for the alias, reported under the key the user wrote rather than the one it expands to. The truthy gate matches npm's own `if (obj[key])`, so the bare key and `=true` abort while `legacy-bundling=false` (npm's default) still installs.

## Verification

Against an npm-incumbent fixture, running the install with `--lockfile-only --offline`:

| `.npmrc` | Exit | Result |
| --- | --- | --- |
| `legacy-bundling=true` | 1 | Aborts, `ERR_NUB_UNSUPPORTED_CONFIG` naming `legacy-bundling` |
| `legacy-bundling` (bare key) | 1 | Aborts, same error |
| `legacy-bundling=false` | 0 | Installs |
| `install-strategy=nested` | 1 | Aborts, unchanged |

```
Error: nub: `legacy-bundling` (npm) is not supported — npm expands it to `install-strategy=nested`; nub installs a hoisted/isolated tree, and npm's nested layout can change which version a require() resolves to. remove `legacy-bundling` from .npmrc [ERR_NUB_UNSUPPORTED_CONFIG]
```

The new test was confirmed to fail without the fix: with only the fatal arm commented out, `scan_fatal_on_legacy_bundling_the_nested_alias` fails through its own assertion (21 passed, 1 failed); with it restored, 22 pass.

Gates run: clippy `--all-targets --all-features` and `cargo fmt --check` are clean, and both `unsupported_config` and `layout_axis_scoping` pass.


colinhacks:
Superseded by #698, which makes layout nub's own axis and removes the `install-strategy=nested` abort this PR extends. Under that policy `legacy-bundling` is a setting nub ignores like every other layout setting — disclosed in the install header rather than refused.

PR #682, opened 2026-08-06
fix(resolver): fall back to the newest mature release when the age gate blocks latest

### LINKED ISSUE #681: nubx should fall back to the newest mature package version when latest is blocked by minimumReleaseAge
When running a package without specifying a version, `nubx` (or `nub dlx`, `nub x`) currently attempts to install latest. If that version hasn't passed minimumReleaseAge, the command fails with ERR_NUB_NO_MATURE_MATCHING_VERSION, even if older acceptable versions exist.

Example:

```bash
~$ nubx skills list
The skills bin is not installed locally.
Install and run from the remote registry?
ERR_NUB_NO_MATURE_MATCHING_VERSION

  × dlx install failed
  ├─▶ failed to resolve dependencies
  ╰─▶ no version of skills matching latest is older than 1440 minute(s) (minimumReleaseAgeStrict=true)
  help: blocked by age gate: 1.5.22
        to bypass for this run: `--minimum-release-age=<duration>` to shorten the window (`0` turns it off), or `--minimum-release-age-exclude=skills` to exempt just this package
        to bypass persistently: shorten `minimumReleaseAge` in .npmrc (`0` turns it off), or add `skills` to `minimumReleaseAgeExclude`
```

`nubx skills list` fails because skills@latest is too new, even though the previous release satisfies the configured age requirement.

Instead, nubx should resolve to the newest version that satisfies minimumReleaseAge, only failing if no acceptable version exists. This preserves the supply-chain protection while maintaining the convenience of nubx.

Thanks for your time!

colinhacks:
A request with no version synthesizes a `latest` range, which resolved to the tagged version's exact range — one candidate, so a `minimumReleaseAge` window covering it had nothing older to fall back to. A blocked `latest` now widens to `<=<tagged>`. Lenient mode had the mirror bug: the lone candidate got re-picked, bypassing the gate.

Bounded at the tag, and `latest` only. pnpm 11 repopulates every tag unbounded, which can hand back a retracted release. Rationale in the commit message.

Verified on the real registry: a merge-base build reproduces the error; this branch runs cowsay@1.5.0 with a notice. 10 new tests, each confirmed red for the right reason.

Closes #681

### DIFF
--- a/site/content/docs/install/index.mdx
+++ b/site/content/docs/install/index.mdx
@@ -516,7 +516,7 @@ $ nub add some-tool
 + some-tool@2.3.0  latest 2.4.0   # ✓ 2.4.0 is still inside the window
 ```
 
-The fallback stops at whatever the publisher currently tags `latest`. A higher version that was published and then untagged is a release they withdrew, so Nub never falls back to one.
+The fallback stops at whatever the publisher currently tags `latest`. A higher version that was published and then untagged is a release they withdrew, so Nub never falls back to one. And when `latest` points at a prerelease, a blocked release fails outright rather than falling back — the stable release below it belongs to a line the publisher has moved off, and the flags below are the way through.
 
 Two flags adjust the window for a single command, so getting past a block never means editing config:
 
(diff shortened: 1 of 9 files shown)


colinhacks:

PR #683, opened 2026-08-06
feat(install): --os / --cpu / --libc for per-invocation platform selection

colinhacks:
Selecting which platform-specific optional dependencies get installed was reachable only through pnpm's `supportedArchitectures` config object. That surface is an override of platform detection dressed as a project property: it is normally wanted for one command, it silently persists once committed, and a `nub.lock` project has no manifest home for it.

pnpm already ships `--os`/`--cpu`/`--libc` on install/add/dlx, so this closes a parity gap rather than inventing a spelling. Repeatable, comma-separated, accepts `current` and `*`, overrides config per axis, never persisted.

Two additions over pnpm, affecting only input pnpm rejects: `*` selects every value on an axis (pnpm installs zero for `*`, `any`, `!name`, and `[]`, silently), and a value naming no platform warns.

Verified against pnpm 10.15.1 on the same fixtures: `--os=linux --cpu=x64` gives an identical result; `--os linux` against config `{os:[darwin],cpu:[arm64,x64]}` yields linux-arm64 + linux-x64, confirming per-axis override; config-only behavior is unchanged.

Refs #677

### DIFF
--- a/crates/nub-cli/src/cli.rs
+++ b/crates/nub-cli/src/cli.rs
@@ -2218,6 +2218,13 @@ fn value_consuming_flags(subcommand: &str) -> &'static [&'static str] {
             "--cwd",
             "--minimum-release-age",
             "--minimum-release-age-exclude",
+            // Platform selection is value-taking too, and the sibling list in
+            // `install_to_add_args` needed the same three. `nubx --os linux -p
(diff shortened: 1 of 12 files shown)


colinhacks:

PR #684, opened 2026-08-06
fix: name NODE_OPTIONS as the constraint refusing a nodeOptions entry

colinhacks:
Project config validates `nodeOptions` entries against the binary's `allowedNodeEnvironmentFlags` — narrower than what Node supports, since the field ships through NODE_OPTIONS and Node keeps V8 and test-runner flags command-line-only.

The refusal blamed Node's support, not the channel. An entry of `--stack-size=2000` reported:

    is not supported by Node 26.5.0

Node supports that flag, and `v8Flags` delivers it. Now:

    is not accepted in NODE_OPTIONS by Node 26.5.0
      V8 flags that Node takes only on the command line go in `v8Flags`

Adds coverage for the real-flag case; the existing test used a typo, where the old wording was accidentally true.

### DIFF
--- a/crates/nub-cli/src/cli.rs
+++ b/crates/nub-cli/src/cli.rs
@@ -3475,7 +3475,15 @@ fn validate_runtime_node_option(
     if let Some(accepted) = accepted
         && !accepted.contains(name)
     {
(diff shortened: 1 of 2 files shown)


colinhacks:

PR #686, opened 2026-08-06
fix(sandbox): copy instead of hardlink for Windows-jailed installs

colinhacks:
The Windows build jail refused every confined install script, which is why win32 has produced no corpus records. Node started, entered its own ESM loader, and died at `getSourceSync` with `EPERM` reading the package's own `install.js`.

**Mechanism.** Windows attaches the security descriptor to the file OBJECT, not the directory entry, so a store file hardlinked into the install keeps the descriptor it was created with under the CAS. MEASURED on `windows-latest` (build-jail-corpus run 31126235717): the denied file carried `SYSTEM`/`Administrators`/`<user>` full control with not one ace marked `(I)`, while `fsutil hardlink list` named four entries for that one object, two of them in the CAS. A descriptor that never auto-inherited is one advapi32's propagation pass also skips, so the jail's inheritable grant lands on every sibling the linker created and never on the linked one.

**Fix.** On Windows, when the embedder's lifecycle sandbox will confine, resolve `LinkStrategy::Copy`, and restore the side-effects cache by copy under the same condition — the second site of the same defect. A copied file is created inside the granted directory, which restores the create-then-inherit invariant the Windows grant path is built around.

**Rejected: acing the file object.** It rewrites the descriptor every OTHER name for that object sees, including the machine-global CAS entry every project on the machine hardlinks from. For a private registry that is source disclosure between packages.

**Cost.** Any Windows install that runs a lifecycle script loses CAS dedupe for that install. Narrowing it to the scripts' read closure needs the dep graph at link time and is not attempted here.

**Verification.** Red-then-green on the new unit test (inverting `jail_forces_copy` fails it). Root `clippy --all-targets --all-features`, `fmt --check`, and `cargo test -p nub-sandbox` green. The real proof is on Windows: build-jail-corpus run 31127907595 builds this branch and measures three arms, with a jail-OFF control that must still show a link count above one.

**Two corrections to the record:**
1. `win-acl-probe` step 0 self-invalidated — MSYS turned `icacls /grant` into `C:/Program Files/Git/grant`, so the hardlink-inheritance premise was never measured.
2. `win-acl-probe` restores a cached `nub.exe` under a prefix key, so re-running it cannot measure a new commit. The new `win-jail-copy-verify` workflow keys on the resolved SHA and builds on a miss.

**Side defect, not fixed here:** `set_ace` (`windows.rs:2381`) feeds the DACL it read — inherited aces included — back through `SetEntriesInAclW`, converting them to explicit aces on every granted path. That residue is visible on `C:\jail\a2` in the run above.

Refs the Windows build-jail bring-up.

### DIFF
--- a/vendor/aube/crates/aube-linker/src/lib.rs
+++ b/vendor/aube/crates/aube-linker/src/lib.rs
@@ -311,7 +311,7 @@ pub struct Linker {
 }
 
 /// Strategy for linking files from the store to node_modules.
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum LinkStrategy {
     /// Copy-on-write (APFS clonefile, btrfs reflink). Selected only by
     /// explicit `packageImportMethod = clone` / `clone-or-copy`, whose
--- a/vendor/aube/crates/aube/src/commands/install/lifecycle.rs
+++ b/vendor/aube/crates/aube/src/commands/install/lifecycle.rs
@@ -3,7 +3,7 @@ use miette::{Context, IntoDiagnostic, miette};
 use super::bin_linking::materialized_pkg_dir;
 use super::node_gyp_bootstrap;
 use super::side_effects_cache::{
-    Confinement, SideEffectsCacheConfig, SideEffectsCacheEntry, SideEffectsCacheRestore,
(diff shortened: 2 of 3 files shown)


colinhacks:
Closing — sandbox/integration is a prerelease feature branch, which per the repo's own flow lands by worktree merge rather than PR. The work is unchanged and will merge directly once verified on a real Windows box.

PR #690, opened 2026-08-07
runtime: refuse an explicit env file when Varlock owns the environment

colinhacks:
An explicit `--env-file` or `envFile` alongside a Varlock schema is two contradicting instructions. The explicit one used to win silently, dropping the loader's half with no diagnostic. Now it errors, naming both sides. `--no-env-file` and `envFile: false` are not conflicts — they load nothing, which standing down already does.

Runs on all three surfaces; `nub watch` never called `check_schema_usable` at all.

Verified against real varlock 1.16.0: both conflict forms exit 1, both no-load forms run redacted, `--env-file` still works with the schema removed. Tests, clippy and fmt clean.

Docs: rewrites the Varlock page; adds `@env-spec` highlighting.


colinhacks:

PR #691, opened 2026-08-07
env: decide .env.schema ownership by loader resolution, and refuse when it is missing

colinhacks:
Removes the `@env-spec` syntax sniff: Nub gated the hand-over on a `# ---` or `# @decorator` line, guessing at a format it does not parse. Whether varlock resolves is now the whole test. A declared `dotenv-extended` is the sole carve-out, and blocks the hand-over rather than only the warning.

Makes an unreadable schema fatal. It previously fell back to `.env*` — a different answer, not a softer one, and silent. Both `SchemaProblem` variants bail; `is_fatal` and the unused `warn_once` are gone. Only the file run and `nub run` are gated, so `nub install` still works.

### DIFF
--- a/crates/nub-cli/src/cli.rs
+++ b/crates/nub-cli/src/cli.rs
@@ -3656,48 +3656,25 @@ fn runtime_child_env(
     Ok(result)
 }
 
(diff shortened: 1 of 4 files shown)


colinhacks:

PR #692, opened 2026-08-07
runtime: apply tsconfig customConditions as Node export conditions

colinhacks:
TypeScript resolves types out of a package's `exports` map using `compilerOptions.customConditions`, but nothing carried them to the runtime. A project on the "live types" layout — a `source` condition pointing at `.ts` — type-checks against the source and then loads the built copy, or fails to resolve at all.

Nub now reads them and emits `--conditions=` NODE_OPTIONS tokens, union'd with `nub.jsonc` `conditions`. Nothing to declare twice:

```jsonc title="tsconfig.json"
{
  "compilerOptions": {
    "customConditions": ["source"]
  }
}
```

No other runner does this. [privatenumber/tsx#574](https://github.com/privatenumber/tsx/issues/574) is the same request, open since June 2024; Vite and Vitest need a separate `resolve.conditions`.

## Mechanism

Passing `--conditions` through `NODE_OPTIONS` reaches **both** the ESM and CJS resolvers — measured on 18.19.0, 20.19.0 and 26.5.0. So one spawn-time flag covers the whole support band with no tier split, and it rides through to workers and child `node` processes. Whether CJS could take conditions at all is the thing the tsx thread stalled on.

Discovery is anchored at the CWD, where `nub.jsonc` discovery already starts and where Node resolves a bare `--import` specifier from, and it follows the `extends` chain so a condition declared in a shared base config counts. Skipped in compat mode along with every other config-derived flag. An empty or whitespace-bearing entry in a tsconfig is skipped rather than fatal — `tsc` tolerates it, and it is not Nub's file to reject.

## Why a new crate

The get-tsconfig port moves out of `nub-native` into a napi-free `crates/nub-tsconfig`, and the addon keeps only the `#[napi]` wrapper. Since `customConditions` conventionally lives in a base config that leaf packages `extends`, a reader in the CLI would otherwise have reimplemented the whole `extends` chain — package-style `extends` through `node_modules`/`exports` included — and then drifted from the addon's copy.

The move is verbatim: 11 lines differ from the old file, all of them napi attributes and two `crate::` paths. `tsconfig_hash` is byte-identical, so warm transpile caches survive. That crate now also owns the pinned JSONC dialect and nesting bound both readers share, and — unlike the cdylib it came from — it can carry unit tests.

## Verification

Remote `cargo test --workspace` and `cargo clippy --all-targets --all-features` over both workspaces, rc=0.

- Five unit tests on the reader: `extends` inheritance, child-replaces-base, option absent, non-string members dropped, explicit path.
- The shared `project_runtime_config` fixture gained a package keyed on a tsconfig-declared condition, so every route asserts it — file run, `run`, `exec`, `nubx`, watch, and the `node` shim.
- One test for the find-up plus `extends` layout, with `--node` as the negative control.
- One unix-only test that a condition may point at TypeScript source in a linked workspace package. The symlink is load-bearing: Node resolves to the realpath, so the `.ts` lives outside `node_modules`. A package that is a real directory there instead dies in Node's own stripper with `ERR_UNSUPPORTED_NODE_MODULES_TYPE_STRIPPING` — existing transpile scope, unchanged here.


colinhacks:

PR #693, opened 2026-08-07
fix(add): report missing packages before reputation warnings

colinhacks:
Nub now verifies a registry package exists before any heuristic reputation warning or prompt. A missing package such as `@types/nub` fails immediately with `ERR_NUB_PACKAGE_NOT_FOUND`; an unscoped typo on the public npm registry can carry a non-interactive suggestion in that authoritative error.

Npm scopes are owned namespaces, so package basenames within one scope no longer trigger unrelated suggestions such as `@types/nub` → `@types/nib`. Unscoped comparisons remain unchanged, and different scopes are still compared by full name. Existing packages retain the prior similarity-before-age challenge order. Names exempt from the reputation gates are checked during normal resolution and preserve the same coded 404. Names routed to private registries never receive suggestions from the public npm popularity corpus.

The lightweight abbreviated-packument existence checks run concurrently. Full packuments remain sequential when the age policy needs their creation timestamps.

## Verification

- Exact repro: exit 40 with `ERR_NUB_PACKAGE_NOT_FOUND`, with no similarity or package-age output
- Reputation-bypass repro: the later manifest fetch preserves the same coded 404
- Missing unscoped public npm typo: `lodaszh` exits 40 and suggests `lodash` without prompting
- Private-registry 404: no suggestion from the public npm corpus
- Focused aube suites: 43 supply-chain tests and 8 manifest/cache tests passed
- Root and native format and all-target/all-feature clippy checks passed
- Static brand checks passed


colinhacks:
Shipped in v0.7.4: https://github.com/nubjs/nub/releases/tag/v0.7.4

PR #694, opened 2026-08-07
config: add init and global scope grammar

colinhacks:
## Summary

- add `nub config init` with commented, behavior-neutral project and user templates
- replace config `--location` with project-default `-g` / `--global` and `nub global config` grammar
- keep protected npm credentials in the user `.npmrc` by default, with explicit `--local` for project credentials, and redact credential values from write reports
- resolve project init at the same root as `config set`, preserve no-clobber behavior, and cover both flag positions
- remove the accidental root `nub.jsonc` that forced this repository into Node compatibility mode

## Testing

- `cargo check -p nub-cli --tests`
- focused config init and credential-scope integration tests
- pre-push site build
- CI


colinhacks:
Shipped in v0.7.4: https://github.com/nubjs/nub/releases/tag/v0.7.4

PR #698, opened 2026-08-08
fix(pm): make layout nub's own axis, with nub.jsonc as its canonical surface

colinhacks:
No branded config file directs the `node_modules` tree now. Drops the last two exceptions: pnpm 11's `pnpm-workspace.yaml` layout keys, and npm's `install-strategy=nested` abort.

`install.linker` / `install.publicHoist` in `nub.jsonc` now apply under every identity; `install.minimumReleaseAge` stays nub-identity-only. A reserved `linker` value is now rejected in a compat project rather than skipped.

Verified against a pre-fix build:

| Case | Pre | Post |
| --- | --- | --- |
| pnpm 11 `nodeLinker` in workspace YAML | `hoisted` | `isolated` |
| pnpm incumbent + `nub.jsonc` linker | ignored | `hoisted` |
| npm `install-strategy=nested` | exit 1 | exit 0 |


colinhacks:

PR #671, opened 2026-08-04
feat(launcher): drop a real nub.exe beside npm's shims on Windows (add-only)

colinhacks:
Every Windows `nub` call is `cmd.exe → nub.cmd → node → spawn nub.exe`; the node boot is ~58 ms of it, and the POSIX heal cannot help there. Hardlinking the binary as `<binDir>\nub.exe` lets PATHEXT resolve it ahead of `nub.cmd`. windows-latest, N=40: **95.6 → 35.8 ms**.

**Deliberately add-only.** npm's `.ps1` and extensionless shims are untouched. Deleting them is faster everywhere but would make nub the first package to modify files npm owns. Accepted cost: sh-family shells gain nothing, including `nub run` under busybox (170.3 → 169.0).

Guarded by `cmdShimLeadsToUs` — an unrelated `nub@1.0.0` exists on npm.

Verified: windows e2e 10/0, POSIX matrix 70/0.

PR #699, opened 2026-08-08
run(windows): carry the bundled POSIX shell with the relocated nub.exe

### LINKED ISSUE #687: missing busybox.exe
Hello, I resintalled nub to 0.7.2 via `npm i -g @nubjs/nub`, then `nub run` not work:

```
> nub -v
v0.7.2
» node v24.13.0 (resolved from package.json#engines.node)

> nub run test
Error: nub's bundled POSIX shell (busybox.exe) was not found next to the nub executable at \\?\C:\Users\xxx\AppData\Roaming\npm\busybox.exe. Reinstall nub, or set `script-shell` in .npmrc to a POSIX shell on PATH.
```

Any help? should i also download nub-win32-x64 and move busybox.exe to Path?

colinhacks:
`nub run` resolves its bundled busybox `sh` relative to the running executable. 0.7.0 (#671) began hardlinking `nub.exe` into npm's global bin dir, away from that sidecar, so on Windows npm installs every `nub run` after the first failed with "bundled POSIX shell (busybox.exe) was not found".

The launcher now carries `busybox.exe` into a `nub-sh/` subdir beside the relocated binary, and the resolution accepts that layout. A subdir, not a sibling: the target dir is on PATH, where a bare `busybox.exe` would shadow the user's own.

Tests: candidate-order unit test; the Windows probe runs its matrix against both layouts plus a negative control.

Also removes a stray root `nub.jsonc` (`nodeCompat: true`) that reached main in 7ae35e252b and has had the whole Test matrix red since eeb66d86e0. Removing it: 219 passed / 0 failed.

Closes #687

### DIFF
--- a/.github/workflows/busybox-run-probe.yml
+++ b/.github/workflows/busybox-run-probe.yml
@@ -1,9 +1,13 @@
 # Branch-scoped Windows probe (see .claude/skills/ci-adhoc-test) — NO PR needed.
 # Pushing to the busybox-shell branch runs this on a real windows-latest runner.
-# It builds a real nub, stages busybox.exe next to nub.exe exactly as the win32
-# package lays it out, and drives the integrated `nub run` (production sidecar
-# resolution, /tmp glue, .bin shim resolution, --script-shell override) — the
-# acceptance test for the bundled busybox script shell that macOS/Linux cannot run.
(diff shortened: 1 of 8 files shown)


colinhacks:

PR #702, opened 2026-08-08
fix: bound unbounded retention in the run supervisor, blob URLs, and the caches

colinhacks:
Seven memory/disk defects, each reproduced and re-measured against plain Node.

| Defect | Before | After |
| --- | --- | --- |
| `run -r` retained every output line | 45→325 MB/50s | flat 15 MB |
| Non-TTY buffered the whole run | climbing | flat 29 MB |
| Object URLs kept a decoded heap copy | +43 MB/20k | +5.6 MB |
| Blobs pinned source buffers | 504 MB ext | 4 MB |
| Sweep never ran for sync runs | never | 527→383 MB |
| Compile-cache dir never evicted | 6.9 GB | swept daily |
| tsconfig memo unbounded | unbounded | 8192 cap |

Also fixes `data:` URLs with an extension-shaped tail throwing ERR_INVALID_URL_SCHEME.

Gates clean, 220/220 integration, bootstrap module list byte-identical.

A `blob:` worker source over 4 MB is no longer accepted; documented.

### DIFF
--- a/crates/nub-cli/src/cli.rs
+++ b/crates/nub-cli/src/cli.rs
@@ -5359,7 +5359,7 @@ fn run_single_script_prefixed(
         if let Some(pre_cmd) =
             nub_core::workspace::scripts::resolve_script(&project.manifest, &pre_name)
         {
-            let (code, _) = spawn_script_prefixed(
+            let code = spawn_script_prefixed(
                 &pre_cmd,
                 project,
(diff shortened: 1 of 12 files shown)


colinhacks:
Shipped in v0.7.5: https://github.com/nubjs/nub/releases/tag/v0.7.5

PR #705, opened 2026-08-09
fix(upgrade): strip the Windows verbatim path prefix, and skip a no-op upgrade

### LINKED ISSUE #664: [Fetaure Request] `nub upgrade` run version check first
Every time I run `nub upgrade`, it just upgrades itself directly without any actions. But honestly, it would make more sense to first check the result, report whether it's already up to date, and then allow the user to choose whether to proceed with the upgrade. For example:

```bash
> nub upgrade
Your version is v0.6.0
The latest version is v0.7.0 (as example)
Are you sure to upgrade? (Y/N):
```
This can also save on download bandwidth.

### LINKED ISSUE #704: windows path
~> nub upgrade
upgrading to v0.7.4 (win32-x64)
installed v0.7.4 to \\?\C:\Users\<username>\.nub

paths aren't written correctly

colinhacks:
`current_nub_binary` canonicalizes, and Windows canonicalization returns `\\?\C:\…`, so every path `nub upgrade` printed carried that prefix. A new `display_path` strips it at the print sites only; the `PathBuf` used for the swap keeps the `MAX_PATH` exemption.

The stable self-owned channel now stops with `already on the latest release (vX)` instead of re-downloading. An explicit `--version` still reinstalls, and `--stable` off a canary still downloads.

`--yes` advertised a prompt that never existed; the help text now matches.

Closes #704
Closes #664

### DIFF
--- a/crates/nub-cli/src/cli.rs
+++ b/crates/nub-cli/src/cli.rs
@@ -441,6 +441,16 @@ fn strip_windows_verbatim_prefix(path: &str) -> String {
     }
 }
 
+/// A path spelled for a HUMAN to read. `current_nub_binary` canonicalizes, and
+/// Windows canonicalization returns the extended-length `\\?\C:\…` form, so any
+/// path derived from it lands in output with a prefix no user typed and no shell
+/// echoes back (#704). Strip it at the print site only: the `PathBuf` itself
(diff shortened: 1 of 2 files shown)


colinhacks:
Shipped in v0.7.5: https://github.com/nubjs/nub/releases/tag/v0.7.5

PR #714, opened 2026-08-11
fix(config): write nub.jsonc at the workspace root

colinhacks:
`nub config init` from a workspace member created `packages/app/nub.jsonc`. It now anchors at the workspace root.

The change is in `project_file()`, the resolver behind `init`, `set`, `get --local` and `delete`, so those cannot disagree about where the config lives. An existing file still wins by up-tree discovery. The root comes from `detect_project`, keeping the pnpm-incumbency gate on `pnpm-workspace.yaml` in one place.

Also isolates `HOME`/`XDG_CONFIG_HOME` and `__NUB_DEPS_CHECKED` in the `verify_deps` tests: three failed on a dev box, and `an_inherited_checked_marker_skips_the_gate` could not fail.

### DIFF
--- a/crates/nub-cli/src/config_fields.rs
+++ b/crates/nub-cli/src/config_fields.rs
@@ -32,6 +32,7 @@
 use std::path::{Path, PathBuf};
 
 use jsonc_parser::cst::CstInputValue;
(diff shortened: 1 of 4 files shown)


colinhacks:

PR #717, opened 2026-08-11
fix(pm): resolve workspace aliases and relative-path workspace specs

### LINKED ISSUE #713: native package manger doesn't resolve workspace aliases
In a workspace, one might want to add some alias to a package (useful in case there are multiple versions of a package). For instance,

```json
{
  "name": "pkg1",
  "version": "1.0.0",
  "dependencies": {
    "pkg2-alias": "workspace:pkg2@*"
  }
}
```

This fails to resolve. npm aliases (`npm:somepkg@*`) works correctly

colinhacks:
`"card": "workspace:card-impl@*"` and `"card": "workspace:../card"` name a member the dependency key does not. Both failed with `package not found`. When the key also named a different member, the locator branch bound to that one — exit 0, wrong package linked and locked.

Parses the tail with pnpm's grammar (an alias needs `@`; a leading `.` is a path) and binds the target before the key lookup.

Also fixed: `link:`/`portal:` deps never got their bins shimmed; a `workspace:` spec binding no member now errors instead of hitting the registry.

Verified against pnpm 10.15.1. New tests confirmed RED before the fix.

Closes #713

### DIFF
--- a/crates/nub-cli/tests/install_engine.rs
+++ b/crates/nub-cli/tests/install_engine.rs
@@ -1480,3 +1480,161 @@ fn linked_esbuild_variants(dir: &Path) -> Vec<String> {
     out.sort();
     out
 }
+
+/// Build a workspace whose consumer `app` declares `deps`, alongside members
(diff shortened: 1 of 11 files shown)


colinhacks:

PR #720, opened 2026-08-12
pm: garbage-collect the virtual store and extracted-tree tiers

colinhacks:
`store prune` only walked the CAS. Nothing collected the global virtual store or `<store>/v1/trees/`. Both key on the dep path plus its graph hash, and `calc_deps_hash` folds each child's hash into every ancestor, so one dep bump strands that package and every ancestor reaching it. One dev machine held 5,999 and 26,143 entries, 6.8%/11.7% of them churn duplicates.

Installs now record the project at `<virtual-store>/.projects/<hash>`; prune marks what registered projects reach and sweeps the rest, as pnpm's `pruneGlobalVirtualStore` does. An empty registry prunes nothing. Also splits out the CAS sweep, which returned early when the CAS root was absent and skipped the tiers.

6 unit tests, each proven to fail when its guard is removed, plus 11 e2e cases in `tests/store-prune/`.


colinhacks:

PR #723, opened 2026-08-12
fix(registry): bound and surface a stalled packument fetch

### LINKED ISSUE #715: Resolver hangs indefinitely on large workspace installs — 0% CPU, in-flight registry requests never time out
### Summary

`nub install` hangs forever mid-resolution on a large workspace — progress stops at roughly a third of the way through (`2321/4487 pkgs · resolving`, exact numbers vary a little), CPU drops to 0%, and nothing ever times out. Left for >10 minutes on multiple runs; it never recovers and never errors. Small projects are unaffected; the odds of the hang seem to approach 1 as the resolution graph grows past a few thousand packages.

Standalone repro (manifests-only mirror of our monorepo, sanitized, hangs on nearly every run): **https://github.com/iamnafets/nub-resolver-hang-repro** — clone, `nub install`, watch it stop advancing within ~60–90s.

This is the same monorepo behind #562. Practical impact for us: with a lockfile-satisfied manifest, installs complete — but any operation that needs re-resolution (adding a single dependency, an override/manifest change) enters this path and hangs, so we can't add a dependency.

### Environment

- nub 0.7.5 (also happens on the fresh-resolve path with no lockfile at all)
- macOS arm64, Darwin 25.2.0
- Node 24.12.0 via `.node-version`
- registry.npmjs.org, no proxy/VPN; fast residential fiber (downloads run 70–100 MB/s when they work)

### State while hung

- `sample` of the process: main thread parked in `pthread_cond_wait`, tokio worker threads all idle. Total CPU time consumed after minutes of "running": <1s.
- `lsof -p <pid> -i`: exactly 2 `ESTABLISHED` TLS connections to the npm CDN (104.16.0.0/12) that never close and never deliver anything. It looks like in-flight metadata requests whose responses are never received, with no client-side timeout to break the wait.
- Tried: `--minimum-release-age 0` (no change), deleting the lockfile (no change), warm vs cold caches (no change), several reporters/log levels (no change).

Possibly the same underlying connection state, surfacing as an error where it's handled: on runs that reach downloading, we intermittently get hard failures like

```
× stream error for @esbuild/freebsd-arm64@0.20.2: HTTP error: error decoding response body
× stream error for @sentry/cli-win32-arm64@2.58.4: HTTP error: error decoding response body
```

on a different package each time, and a retry proceeds past it.

### No single culprit package — graph-size sensitivity

I bisected the workspace's dependency surface looking for one bad package and didn't find one:

| tree | ~pkgs resolved | result |
|---|---|---|
| full workspace (75 members) | ~6,000 | hangs (~2,300 in) |
| repro repo as published (`devDependencies` trimmed to one member's) | ~4,500 | hangs (~2,300 in) |
| same tree, all `devDependencies` removed | ~3,500 | completes |
| bare project with just `@remotion/cli@4.0.499` (the devDep whose presence tips this tree over) | ~1,200 | completes |

The full-size resolve did complete once across many attempts on the real repo, which also points at a race (lost waker / connection-pool slot never released?) whose probability scales with resolution concurrency, rather than a deterministic cycle.

Happy to run instrumented builds or capture more state (we can reproduce this reliably within ~90 seconds).

colinhacks:
Closes #715

An install froze mid-resolve at 0% CPU with no output when a registry connection was accepted and then went silent. Three defects on the packument path, measured against a registry that accepts a connection and never answers.

## Nothing bounded a stalled stream

The `fetchTimeout` setting is a whole-request budget, and there was no idle bound, so a silent stream burned all 5 minutes of it before anything failed. Adds `fetchStallTimeout` (default 60000, `0` disables), applied via reqwest's `read_timeout`. That clock resets on every delivered frame, so a large-but-progressing packument is never cut off, and it also gates the wait for response headers — a socket that is accepted and never answered fails on it too.

## The timeout cap never reached the packument path

`TIMEOUT_RETRY_CAP` was applied in the two tarball helpers in `request.rs` and in none of the four copied retry loops in `packument.rs`, so a stalled packument spent the whole attempt budget.

| | before | after |
|---|---|---|
| attempts against a silent registry | 3 × 300s | 60s + 1 retry |
| wall-clock before an error | 970s | ~130s |

All twelve retry arms now route through a new `retry_policy` module, so the cap, the backoff and the wording cannot drift per call site again.

## A stall produced no output

The `slow_metadata::record` call runs only after a request returns, so a request that never returned printed nothing at all. In-flight metadata requests are now tracked, and a ticker reports them every 10s:

```
WARN still waiting on 9 registry metadata requests (longest: packument @prisma/client, 40s)
WARN packument playwright: response body could not be read — retrying in 10s (attempt 2 of 3)
```

## Verification

- 296 `aube-registry` tests pass. Three new ones cover the retry cap, the idle bound, and a positive control that a slow-but-progressing response is not cut off.
- Reverting the timeout classification makes the cap test fail with 6 attempts, which is the pre-fix behavior.
- clippy `--all-targets --all-features -D warnings` and `cargo fmt --check` pass in both workspaces.
- Driven end-to-end against the reporter's repro tree through a proxy that stalls a packument mid-body.

Also corrects a comment claiming the `fetchTimeout` default matches npm's 60s; it is 300000.

### DIFF
--- a/tests/registry-stall/README.md
+++ b/tests/registry-stall/README.md
@@ -2,14 +2,16 @@
 
 This directory is the working system for nub's behavior when a registry accepts a connection and then goes quiet. It exists because the defect class here is invisible to unit tests: a stalled stream is not an error, so nothing fails, nothing logs, and the resolver simply blocks at 0% CPU until some bound expires. The bug that motivated it ([#715](https://github.com/nubjs/nub/issues/715)) presented as "`nub install` hangs forever" and was really a 970-second wait — a distinction only wall-clock can make.
 
-`wiki/` carries no decision record for this; the mechanism is documented in `crates/aube-registry/src/client/retry_policy.rs` and the `fetchStallTimeout` entry in `settings.toml`. This README documents the *loop*.
+`wiki/` carries no decision record for this; the mechanism is documented in `vendor/aube/crates/aube-registry/src/client/retry_policy.rs` and the `fetchStallTimeout` entry in `vendor/aube/crates/aube-settings/settings.toml`. This README documents the *loop*.
 
 ## The loop
 
 1. Build a binary: `scripts/rust-build.sh build -p nub-cli --profile fast`.
 2. Run the matrix: `tests/registry-stall/run-stall-matrix.sh "$(scripts/rust-build.sh --print-target)/fast/nub"`.
 
-It takes about three minutes, needs no network, and exits non-zero on the first case outside its window.
+It takes about three minutes, needs no network, runs all five cases, and exits non-zero if any of them missed its window. It is deliberately not fail-fast: which bounds moved *together* is what localises a regression, and that is only visible if every case runs.
+
+Each case runs with the ambient `npm_config_*` / `AUBE_*` / `NUB_*` environment stripped, because `env` outranks the project `.npmrc` in the settings precedence chain — without that, a developer who exports `npm_config_registry` would silently measure something other than what the case claims.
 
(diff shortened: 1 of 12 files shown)


colinhacks:

PR #725, opened 2026-08-12
wiki: adopt lat.md as the corpus knowledge graph

colinhacks:
Pilots [lat.md](https://github.com/1st1/lat.md) over the public `wiki/` corpus: cross-linked markdown with `lat check` as a required CI gate, so a doc naming a symbol that no longer exists fails the build instead of rotting.

**`/wiki` is kept.** The graph directory name is hardcoded in lat and is not configurable, but a tracked `lat.md -> wiki` symlink works — verified. The only visible cost is that lat derives the root index filename from the directory basename, so the root index is `wiki/lat.md`.

## What is here

| Change | Notes |
| --- | --- |
| `lat.md -> wiki` symlink + three directory indexes | `wiki/lat.md`, `wiki/design/design.md`, `wiki/research/research.md` |
| `wiki/research/README.md` removed | `research.md` replaces it. lat enforces that index's completeness, so it cannot silently go stale; the hand-maintained README could |
| 11 benchmark JSON files → `benchmarks/data/native-http/` | lat rejects non-markdown inside the graph. Their 11 links are repointed |
| ~800 sections given a summary paragraph | lat requires every heading to carry a ≤250-char summary; it is what `lat search` returns |
| `npm run lat:check` + `.github/workflows/lat-check.yml` | `npx` fetches the checker at a pinned `0.12.2`; the root install is byte-identical to main. The pin covers `lat.md` itself, not its ~158 transitive packages, which npm re-resolves per run |
| Three doc→code links in `wiki/design/architecture.md` + three `@lat:` backlinks in `nub-core` | `lat refs` resolves the Composition section back to `spawn.rs` |
| `wiki/design/compat-mode-tests.md` | four `require-code-mention` specs, each backed by a real compat-mode test |
| `.claude/skills/lat-md/SKILL.md` + an `AGENTS.md` section | |

The test-spec gate was verified by breaking it: deleting one `@lat:` backlink makes `lat check` exit 1 and name the orphaned spec; restoring it clears the error.

## Costs, measured

- **802 errors on first run, across 71 of 73 files.** 432 sections had no summary paragraph, 356 had one over 250 characters (worst: 1,530), plus the 11 non-markdown files and 3 missing indexes.
- **69 of those 432 were `## Changelog` sections.** nub's own doc convention collides with the rule, and the checker has no exemption, so each got a near-identical line. That is the least defensible part of this diff.
- **lat does not enable GFM, so a pipe table parses as one paragraph.** A table-led section is therefore reported as an over-long "leading paragraph" — one was 16,259 characters for a 137-row table. Splitting is impossible; the fix is a real summary above the table, which leaves the table byte-identical.
- **The checker cannot be a root devDependency.** It brings ~185 transitive packages, and every `ci.yml` test leg runs the root `npm ci` through the Socket Firewall shim. That step's own comment records the failure mode — an sfw crash that exits 0 leaves a half-built tree behind a green install step — and adding that much tripped the following "assert root deps actually installed" guard on every leg. The first push of this branch did exactly that. It now uses `npx` at a pinned version, so `package-lock.json` is unchanged.
- **Semantic search costs a 128 MB index and a 5m56s first build** (1,394 sections, offline WASM model, no API key). It is paid once per clone, on the first `lat search`, and the index lands in `wiki/.cache/`, which the repo's existing `.*` rule already ignores. The result quality is good: "why is the user's Node spawned instead of embedded" returns `node-embedding-vs-spawn` first.
- **A Rust symbol inside a `mod` block cannot be wiki-linked.** lat's extractor walks top-level items only, so 972 of 1,522 `#[test]` functions are unaddressable that way. `@lat:` backlinks are a comment scan and work anywhere, including inside `#[cfg(test)] mod tests`, so test specs are unaffected.

## Worth a look before merging

The inserted summary sentences are the substance of this diff. Each was written by reading the section rather than its heading, and no finding, number, table cell, code sample, or conclusion changed — but it is a lot of new prose in the public corpus, and it is the part to spot-check.

## What else moved

`AGENTS.md` now lives in the graph as `wiki/agents.md`, reached through a root symlink, so the repo's most-read document is inside the gate. That immediately caught a dangling `[[node-hijack-compat-design]]` it was already carrying. Its nine broken `wiki/` references are gone too — six de-linked to the private corpus, three repointed — and every document reference in it is now either a checked `[[…]]` link or a path that resolves.

The corpus also gained **25 `@lat:` backlinks** from source to the research that explains it, and its **221 cross-document markdown links became `[[…]]`**, so they are validated rather than invisible. Five links stay as markdown because their targets sit outside the graph.


colinhacks:

PR #727, opened 2026-08-12
fix(pm): order dependency lifecycle builds

colinhacks:
## Summary

- run dependency lifecycle scripts in children-first graph phases
- preserve bounded parallelism for independent packages and collapse dependency cycles
- cover ordering through non-building graph nodes and an end-to-end parent/child build

## Verification

- `make verify`
- fresh `spectron@17.0.0` install with an isolated Nub store and Electron cache
- control regression fails on `main` when the parent starts before its child finishes


colinhacks:

PR #734, opened 2026-08-14
types: cover the six polyfilled proposals an es2024 lib can't reach

colinhacks:
`@nubjs/types` exists so that code Nub runs also typechecks. A consumer on the lib `nub init` scaffolds — `"lib": ["es2024"]` — could not type seven features Nub provides across its entire Node floor.

Found while answering a question about Explicit Resource Management support. ERM itself turned out fine: `@types/node` carries `/// <reference lib="esnext.disposable" />`, so `using` types correctly today. Probing the rest of `feature_matrix.rs` against a faithful scaffold surfaced these.

| Feature | Reached by |
| --- | --- |
| `Set.prototype.union` and the other set methods | `es2025.collection` / `esnext.collection` |
| `Array.fromAsync` | `esnext.array` |
| `RegExp.escape` | hand-declared |
| `Error.isError` | hand-declared |
| `Promise.try` | hand-declared |
| `Map`/`WeakMap` `getOrInsert`, `getOrInsertComputed` | hand-declared |
| `Iterator.concat` | hand-declared |

A library reference is only safe where the name resolves in every TypeScript the entry point serves; anything older is hand-declared instead. The sibling `Iterator.zip`/`zipKeyed` stay out on purpose — their element type depends on the `mode` option, so declaring them honestly needs discriminated overloads.

Raising the scaffold's `lib` to `esnext` was the alternative and is rejected: `common.d.ts` states the package's design is to type this surface *at* the ES2024 target, and `esnext` floats with the TypeScript release.

## The bug review caught

The first version referenced `lib.esnext.error`, which first exists in TypeScript 5.9. `typesVersions` routes **every** TypeScript at or below 5.9 to the same entry point, and an unknown `reference lib` is a hard `TS2726` that fails the consumer's whole build — so it broke 5.7 and 5.8 outright, where `main` compiles clean.

That is why the fixture matrix gains a **TypeScript 5.8 leg**: the oldest version `typesVersions` routes to that file, and the only leg that can see this class of defect. Verified it catches the original bug — re-adding the reference turns 5.8 red while 5.9 stays green.

## Verification

| Check | Result |
| --- | --- |
| Fixtures | 18 runs across TypeScript 7.0.2, 6.0.3, 5.9.2, 5.8.3 |
| All seven features typecheck | 5.7.3, 5.8.3, 5.9.2, 6.0.3, 7.0.2 |
| All seven at **runtime** | Node 18.19, 20.19, 22.14, 26.5 |
| Every fixture pin | verified red against a specific weaker signature |

Two step-aside fixtures skip on the 5.8 leg — its own `lib.dom` disagrees with the pinned `@types/node` about `TextDecoder`, independently of this package. The skip is printed, never silent.

## Also

Corrects the launch post's claim that `using` is "native where V8 has shipped it, safely down-transpiled otherwise". Nub lowers the syntax unconditionally at the hardcoded es2022 target, including on Node 26 where V8 has ERM natively — verified by stack probe on 18.19/22.14/24.0/26.5.


colinhacks:

PR #735, opened 2026-08-15
runtime: let envFile take precedence over the Varlock hand-over

colinhacks:
`envFile` now takes precedence over a `.env.schema` hand-over, and a bare-string path is refused in favour of an array.

A `.env.schema` file is inferred intent — its presence is what makes nub stand down and put Varlock in front of Node. An `envFile` value is declared intent. Declared now wins.

## What changes

| Input | Loader runs | nub loads |
| --- | --- | --- |
| no `envFile`, schema present | yes | nothing — unchanged |
| `envFile: true` | no | the `.env*` cascade |
| `envFile: false` | no | nothing at all |
| `envFile: ["custom.env"]` | no | `custom.env` |
| `envFile: "varlock"` | yes | nothing |
| `--env-file` / `--no-env-file` | no | the named files / nothing |
| global `envFile`, schema present | yes | nothing — a global value cannot displace a project schema |

Three behaviours are replaced:

- **`envFile` beside a schema was a hard error** telling the user to pick one. A project wanting a schema in CI and a plain `.env` locally could not say so, and the error also fired at projects whose only fault was a global setting they never wrote.
- **`envFile: false` and `--no-env-file` did nothing** in a schema project. They were classified as non-conflicting because standing down already loads nothing — which read the hand-over as the absence of loading rather than as its own answer. Asking for no environment now means no environment.
- **A declared `envFile` also clears the missing-Varlock error**, so a schema project is runnable on a machine where Varlock will not install without giving up augmentation via `--node`.

## Breaking: a path goes in an array

`envFile: ".env"` is now an error naming `[".env"]`. A string is a mode name, and `"varlock"` is the only one.

This makes `envFile` consistent with every other list-valued field (`preload`, `nodeOptions`, `v8Flags`, `conditions`, `install.publicHoist`), each of which already refuses a bare string, and with the two string-valued fields (`verifyDeps`, `install.linker`), each of which treats a string as a closed mode list. `envFile` was the only field where a bare string was data.

The exception was already costing a diagnostic. `nub config set envFile .env,.env.local` was accepted and wrote a single path named `.env,.env.local`, failing much later at file resolution; the same input against `preload` is refused on the spot. Both now behave the same way.

Freeing the string slot is also what allows `"varlock"` to be spelled without a sigil. `$varlock` was not available: `envFile` values run through `${VAR}`/`$VAR` expansion, so `$varlock` already resolves against the environment.

## Scope

Only a project file, the environment layer, or a CLI flag displaces a schema. A global `envFile` does not. `nub config set --global envFile false` is a documented personal default, and letting it reach into every checkout would empty the environment of every schema project on the machine — silently, far from the config that caused it, and with no committed `.env` to fall back on.

A global value that is overridden this way is ignored without a warning. A warning would fire on every run in the project; if that turns out to bite, it is a cheap follow-up.

## Implementation

The rule suppresses detection rather than unpicking a hand-over downstream, so the diagnostics, the child environment, the spawn chain, and the watch path all see the same `None` they already handle for a project with no schema. That also made two branches dead — a `Sources` value can no longer coexist with an owner — and both are removed along with the comments describing the refusal they existed to plug.

`RuntimeEnvFile` deliberately gains no `Varlock` variant. It is a cross-version wire format, and an unknown tag fails the whole deserialize, so a new variant would abort the run on a nub of a different version mid-upgrade. Selecting the loader is an outer-process decision and never travels on that wire.


colinhacks:

Issue #450, opened 2026-07-13
isolated node-linker: peer types not deduped like pnpm, breaking astro-icon <Icon> props


afonsojramos:
## Summary

Under the default `isolated` (pnpm-style) node_modules layout, `nub install` doesn't make a dependency's peer types resolve to the *same* instance the project uses, the way pnpm's own isolated layout does. This breaks `astro-icon`: its `<Icon>` component loses the HTML attributes it inherits from `astro`'s `HTMLAttributes`, so `astro check` fails. pnpm resolves the identical lockfile/tree cleanly; `nub install --node-linker hoisted` also works. So this looks specific to nub's isolated symlink layout, not the lockfile.

## Environment

- nub 0.4.11 (macOS arm64, Node 24)
- Shared `pnpm-lock.yaml` (lockfileVersion 9) — unchanged by nub (`git status` clean after install)
- astro `^7`, astro-icon `^1.1.5`, typescript `^6`

## Repro

A minimal Astro project using `astro-icon`:

```
{
  "dependencies": { "astro": "^7", "astro-icon": "^1.1.5", "@iconify-json/lucide": "^1" }
}
```

```astro
---
import { Icon } from 'astro-icon/components';
---
<Icon name="lucide:check" class="h-4 w-4" />
```

1. `nub install`          # default isolated layout
2. `astro check`

**Actual** (nub, isolated):

```
error ts(2322): Type '{ name: string; class: string; ... }' is not assignable to type 'IntrinsicAttributes & Props'.
  Property 'class' does not exist on type 'IntrinsicAttributes & Props'.
```

**Expected:** `astro check` passes (0 errors), matching pnpm.

## What resolves it

- `pnpm install` (pnpm's own isolated layout) → 0 errors
- `nub install --node-linker hoisted` → 0 errors
- `node-linker=hoisted` in `.npmrc` (nub honors it) → 0 errors, and pnpm honors it too, so CI stays consistent

## Likely cause

`astro-icon`'s `Icon` Props extends `astro`'s `HTMLAttributes<'svg'>`. That inheritance only works if `astro-icon` resolves the *same* `astro` types the project does. Under nub's isolated symlinking the peer `astro` seen by `astro-icon` appears to differ (or the augmentation isn't deduped), so the inherited HTML-attribute props are dropped. pnpm's isolated layout dedupes the peer; nub's doesn't. `hoisted` sidesteps it by flattening.

Happy to share a full minimal repo repo if useful.

[...]

PR #736, opened 2026-08-15
fix(pm): apply the disk-materialize eject in workspace projects too

### LINKED ISSUE #711: Type-phantom ejection (#452) is skipped in projects with workspace members
Follow-up to #450 / #452 ("Fixed in v0.4.12"): the type-phantom detector doesn't run when the project has at least one workspace member. No package is ejected from the global store, so `@storybook/react`'s realpath stays in `~/.cache/nub/pm/store`, TS's upward walk never re-enters the project, `@types/react` is unreachable, and `Meta`/`StoryObj` degrade to `any` — same `TS7006` signature as the TanStack case in #450.

**One empty workspace member is enough to disable it.** Same directory, same files, nub 0.7.5 (macOS 15 arm64, Node 22.23.2, `pnpm-lock.yaml` v9 from pnpm 10.34.5, typescript 5.9.3):

| Fixture | pnpm | nub GVS | packages ejected |
|---|---|---|---|
| no `pnpm-workspace.yaml` | 0 errors | 0 errors | 9 |
| workspace file, glob matches no member | 0 errors | 0 errors | 9 |
| workspace file + one empty member | 0 errors | **TS7006** | **0** |
| real monorepo, 23 members, 2589 pkgs | 0 errors | **17 errors** | **0** |

So the trigger is workspace membership, not the workspace file.

## Repro

`pnpm-workspace.yaml` with `packages: - packages/*`, one `packages/dummy/package.json` (`{"name":"dummy","private":true,"version":"0.0.0"}`), root devDeps `@storybook/react@8.6.18 storybook@8.6.18 react@19.2.8 react-dom@19.2.8 @types/react@19.2.18 @types/react-dom@19.2.4 @types/node@22.19.4 typescript@5.9.3`, tsconfig with `strict`, `jsx: react-jsx`, `moduleResolution: Bundler`, `skipLibCheck: true`, and:

```tsx
const meta: Meta<typeof Button> = { component: Button, args: { label: "Click me" } };
export default meta;
export const Outline: StoryObj<typeof meta> = { render: (args) => <Button {...args} /> };
```

```
pnpm install && npx tsc -p tsconfig.json                  # 0 errors
rm -rf node_modules && nub install --frozen-lockfile
./node_modules/.bin/tsc -p tsconfig.json                  # TS7006: Parameter 'args' implicitly has an 'any' type
```

`Button` is a trivial component with `{ label?: string; variant?: "default" | "outline" }` props. Happy to attach the full fixture as a repo/gist.

## Underlying cause (as in #450)

With `--skipLibCheck false`: `TS7016 Could not find a declaration file for module 'react'` from `~/.cache/nub/pm/store/@storybook+react@8.6.18_.../dist/public-types-*.d.ts`. The walk goes `~/.cache/nub/pm/store` → `~/.cache/nub/pm` → `~/.cache` → `$HOME`, never back into the project; pnpm's walk finds `node_modules/.pnpm/node_modules/@types/react`. The sidecar tier exists but is unreachable: 2317 of 2318 `.store` entries are symlinks into `~/.cache`, so no realpath passes through `.store`.

## Ruled out

`@types/react` in the root `package.json` (still 0 ejections), `hoist-pattern[]=*` / `public-hoist-pattern[]=*types*`, `tsc --preserveSymlinks` (162 errors). Resolution fidelity matches pnpm's package set; lockfile untouched (`git status` clean after install).

## Workaround

`node-linker=hoisted` — but 25.8 s per install vs 14.4 s pnpm and 2.0 s nub GVS, so it costs the entire speed gain.

colinhacks:
## Problem

Resolving a single workspace member disabled every disk-materialize eject. A project with one member — even an empty one — kept every registry package symlinked into the machine-global store, so no package realpath returned to the project and each eject class silently stopped working:

- **Type-phantom (#450, #452)** — `@react-pdf/renderer`'s `.d.ts` imports `react`, the project's `@types/react` sits only at the top level, and TypeScript's upward walk from the store never re-enters the project. `TS2786`/`TS2607`, or `TS7006` on the TanStack shape.
- **Undeclared-phantom** — `@hookform/resolvers` and similar keep 404ing their undeclared target.
- **Project-context (#457)** — `simple-git-hooks` and the other host-project mutators run detached from any project again.

Legacy-vite (#315, #318) was unaffected: it rides `project_local_dep_paths`, a separate channel the workspace path already honored.

Reproduced on `main` with one fixture varied only by workspace shape:

| Fixture | Ejected | `tsc` |
|---|---|---|
| No workspace | 4 | 0 errors |
| `workspaces` glob matching no member | 4 | 0 errors |
| Same glob plus one empty member | 0 | `TS2786` |
| `pnpm-workspace.yaml` plus one member, pnpm incumbent | 0 | `TS2786` |

## Cause

The install routes on `has_workspace`, which is true as soon as one member resolves. That picks `link_workspace` over `link_all`, and only `link_all` consulted the eject set — its GVS populate pass had the `disk_materialize_matches` branch, and the workspace twin never got one. The install digest still printed the eject because it records the plan, not what the linker ran.

The two passes were otherwise identical line for line, and this is the second bug that duplication has shipped: #566/#576 was the `EntryState::Stale` arm landing in `link_all` alone, which wedged Windows workspaces on os-183.

## Fix

Both passes now call one `Linker::gvs_populate_entry`, so a branch can no longer reach one and miss the other. Sharing the body rather than mirroring it a second time is the point of the change.

## Reaching the projects that already broke

The linker change alone reaches nobody who hit this. Their tree carries a current install state, and nothing about the fix moves the hash — lockfile, manifest, resolved settings and the disk-materialize seed are all identical, since only which code path consults the seed changed. So the first install after upgrading short-circuits on the warm fast path, reports "Already up to date", and the all-symlinks layout survives. Installing a workspace fixture with a pre-fix binary and then running a post-fix one confirmed it: nothing ejected, `TS2786` still there. Layout verification does not catch it either, because it checks entry existence and package identity, never entry type.

So a `GVS_EJECT_ALGO_VERSION` is folded into the settings fingerprint, the seam already carrying the scanner version and the #457 curated-eject token for this same class of staleness, and the shape aube's `hoisted_layout_algo` salt already uses. It is deliberately not gated on a non-empty `diskMaterializePackages`: that setting is almost always empty here because the eject set arrives through the expand hook, so gating there would skip the salt in exactly the broken cases.

The cost is one full install on the first run after upgrading, and that is more than a relink. The dependency side is cheap — no refetch, no rebuild, no side-effects-cache bust and no lockfile churn, all of which stay gated on content-hash deltas. Root lifecycle hooks are not: they gate only on the fast path being missed, so `preinstall` and `install`/`postinstall`/`prepare` re-run once per importer, which is worth knowing for a workspace whose members build from `prepare`. That is accepted rather than overlooked — the alternative is leaving every already-installed workspace on the broken layout, and scanner-version bumps already carry the same cost. Narrowing the salt to fire only when the eject closure is non-empty is not available, because the closure needs the resolved graph and this hash is computed before resolution. The auto-install path behind `nub run` does not pay it at all, since it skips the settings-hash check.

Measured on the upgrade path: pre-fix install ejects 0 and fails `TS2786`, the post-fix binary relinks to 9 ejected and 0 errors, and a third run reports "Already up to date" — so the salt fires once and settles.

## Validation

- Unit — a workspace twin of the existing `link_all` eject test, asserting the listed package becomes a real project-local dir while an unlisted one stays a shared-store symlink. It fails on the pre-fix linker. The fingerprint test pins the new algorithm fold.
- End to end, real `nub install` — all four fixtures above eject and type-check, across both the neutral `workspaces` field and the pnpm-incumbent path; `simple-git-hooks` and `@hookform/resolvers` eject in a workspace again; `vite` is unchanged.
- Cost — on a five-member React/eslint/vitest workspace, 21 of 245 packages eject, inside the existing bounded-subtree guard and matching what the same dependency set already ejects without a workspace.

Fixes #711

### DIFF
--- a/crates/nub-cli/src/dynamic_phantom.rs
+++ b/crates/nub-cli/src/dynamic_phantom.rs
@@ -92,7 +92,9 @@ fn eject_disabled(raw: Option<&str>) -> bool {
 /// members are injected inside the expand hook, past aube's `disk_materialize_packages`
 /// settings fold, so folding the list token here is what invalidates a warm tree on
 /// the initial ship AND on any future list edit (else the stale symlinked shape is
-/// accepted and #457 stays unfixed on existing installs).
+/// accepted and #457 stays unfixed on existing installs). It also folds
+/// [`GVS_EJECT_ALGO_VERSION`], which covers the third way a warm tree goes stale:
+/// the plan is unchanged but the LINKER writes it differently (nub#711).
 ///
 /// The token still branches on [`enabled`] SOLELY for the internal A/B seam: when
 /// an agent flips [`INTERNAL_EJECT_DISABLE_VAR`] the token changes, so a warm tree
@@ -107,7 +109,7 @@ pub(crate) fn settings_fingerprint() -> String {
 fn settings_token(enabled: bool) -> String {
     if enabled {
         format!(
-            "phantom_scanner={PHANTOM_SCANNER_VERSION};project_context={}",
+            "phantom_scanner={PHANTOM_SCANNER_VERSION};project_context={};gvs_eject_algo={GVS_EJECT_ALGO_VERSION}",
             crate::pm_engine::phantom_closure::project_context_eject_token()
         )
     } else {
@@ -308,6 +310,25 @@ pub(crate) fn phantom_cache_dir() -> Option<PathBuf> {
 /// is structural, nothing else to remember.
 pub(crate) const PHANTOM_SCANNER_VERSION: u32 = 4;
 
+/// Version of what the linker's GVS-populate pass WRITES TO DISK for a given eject
+/// set — bumped when the same plan produces a different on-disk shape.
+///
+/// Distinct from [`PHANTOM_SCANNER_VERSION`] (which plan is computed) and from
+/// aube's `disk_materialize_packages` fold (which NAMES are in the seed): both of
+/// those are unchanged when only the EXECUTOR changes, so neither invalidates.
+/// nub#711 is the case in point — `link_workspace` never consulted the eject set,
+/// so every workspace install produced an all-symlinks tree. Fixing the linker
+/// moves no hash: the lockfile, the manifest, the settings and the seed are all
+/// identical, so `try_install_fast_path` reports "Already up to date" and the
+/// broken layout survives the upgrade. Only the users who filed the bug have such
+/// a tree, so without this salt the fix reaches nobody until unrelated churn
+/// (a lockfile edit, `--force`) happens to bust the state.
+///
+/// Same shape and same remedy as aube's `hoisted_layout_algo` salt, which exists
+/// because a hoisted-layout algorithm change likewise left the graph hash
+/// identical. Bump on any future change to what that pass materializes.
+pub(crate) const GVS_EJECT_ALGO_VERSION: u32 = 1;
+
 /// THE single source of truth for a phantom sidecar's location: the versioned
 /// subdir `<phantom_cache_dir>/s<PHANTOM_SCANNER_VERSION>/<fingerprint>.json`.
 /// Both halves derive their path HERE — the extract-time PRODUCER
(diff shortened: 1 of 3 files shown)


colinhacks:

PR #737, opened 2026-08-15
fix(install): an optional dependency's build failure no longer fails the install

### LINKED ISSUE #660: A failing optionalDependency build fails the whole install; npm and pnpm exit 0
When an `optionalDependencies` entry has a lifecycle script that exits non-zero, `nub install` fails. npm and pnpm complete the install.

Fixture: `optionalDependencies: { optfail: "file:../dep" }`, dep postinstall `exit 3`.

| | exit | dep in `node_modules` |
|---|---|---|
| npm 11.17.0 | 0 | removed |
| pnpm 10.15.1 | 0 | left in place |
| nub 0.6.0 | **1** | removed |

Optional dependencies exist so a failure is tolerable — a platform-specific binary that will not build here is the common case. Failing the install removes that property.

Neither reference tool names the cause, so a warning naming the package and its failing script would improve on both.

macOS arm64, Node 26.5.0.

colinhacks:
An `optionalDependencies` entry whose lifecycle script exits non-zero failed the whole install. npm and pnpm both continue.

Optionality is derived from the graph's edges via a new non-mutating `optional_only_packages`. Reading `LockedPackage::optional` would be inert on a frozen install off an npm/bun/yarn lockfile, where it is always `false`. One fully-required path keeps a package required, so a required build failure still fails the install.

Skips warn under `WARN_NUB_OPTIONAL_BUILD_FAILED` with the build error. That warning comes from the install that runs the build; a later up-to-date install skips the package, so it is not repeated.

The failed dependency stays in `node_modules`, matching pnpm — npm removes it. Case 4 of `wiki/design/install-script-failure-readout.md` recommends npm's removal, so this records what shipped against that case and what the unlink still owes rather than closing it silently. The behavior is documented under lifecycle scripts, and the new code is authored in the `aube-codes` registry with the derived JSON regenerated from it.

Verified against npm, bun and yarn-classic lockfiles, a frozen `nub ci`, both linkers, a multi-importer workspace, and a native addon whose implicit `node-gyp` build fails — each with a required-dependency control.

Ported from `92ed4cc78f` on `sandbox/integration`, narrowed to drop its dependence on that branch's sibling-drain change so it lands standalone.

Closes #660

### DIFF
--- a/site/content/docs/install/index.mdx
+++ b/site/content/docs/install/index.mdx
@@ -469,7 +469,7 @@ That field applies in any project, whoever owns it. A project owned by another p
 A failed build fails the install. The one exception is a package reachable only through `optionalDependencies`, which the project has declared it can work without: that failure is reported and the install continues, matching npm and pnpm.
 
 ```console
-# captured: optfail@1.0.0 whose postinstall exits 3, allowBuilds entry present
+# captured: nub 0.7.5, optfail@1.0.0 whose postinstall exits 3, allowBuilds entry present
 $ nub install
 WARN optfail@1.0.0 is an optional dependency and failed to build; continuing
      without it: lifecycle script postinstall failed for optfail@1.0.0: script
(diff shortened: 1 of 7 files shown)


colinhacks:

Issue #728, opened 2026-08-13
Submit nub to the official brew core tap


fraluc06:
Hi, I'm a Homebrew contributor. I opened a [PR](https://github.com/Homebrew/homebrew-core/pull/298614#pullrequestreview-4928132875) to include `nub` in the official `homebrew/core` tap. After the PR is merged, you can update the docs and point the installation via brew directly to the official tap, so you won't have to maintain a personal tap anymore and it will be installable via a simple `brew install nub`.

colinhacks:
Done in #821. The README, docs install page, FAQ, node-shim page, agent start guide, and homepage install tabs all show `brew install nub` now.

The `nubjs/tap` tap stays live and the release workflow keeps bumping it, so existing tap installs keep upgrading from the tap. The README carries a migration note for switching over.

colinhacks:
Shipped in v0.8.3: https://github.com/nubjs/nub/releases/tag/v0.8.3 — the docs now point at the core formula. Thanks again for the contribution.

PR #738, opened 2026-08-15
embed-runtime: require the vendored runtime deps at build time

colinhacks:
`build.rs` validated only `preload.mjs` when embedding the runtime, so a stage missing the five vendored packages produced a green build and a degraded binary. The extracted tree has no parent `node_modules` to resolve through, so that binary fails with `MODULE_NOT_FOUND` on any file needing a transpile helper, and silently drops Temporal, URLPattern and Float16Array on Node versions that lack them natively.

Nothing downstream caught it. The integrity hashes cover only the entrypoints, and a smoke test on a current Node exercises none of those paths — Node 26 has all four natively, so both the intact and degraded runtimes pass there.

Found while triaging #728: the homebrew-core formula ran `npm install` from `runtime/`, which resolves up to the repository root because `runtime/` has no package.json, so `runtime/node_modules` was never created. That formula fix is separate and already merged into the contributor's branch.

## What this does

The build now fails and names the absent packages. `NUB_ALLOW_INCOMPLETE_RUNTIME=1` opts out, for the sites that build the feature but never run the binary.

| Opts out | Stages the packages, gate stays on |
| --- | --- |
| `ci.yml` `clippy` / `embed-runtime` / `macos-embed-roundtrip`, `make verify`, `scripts/remote-build.ts` (job prep and image bake) | `release.yml`, both `compile-native.yml` jobs, `ci.yml` `test` and `windows-embed-roundtrip`, `flake.nix`, `npm/build-local.sh` |

The `check` job needs no grant — its only `--all-features` run is in `crates/nub-native`, a separate workspace without the feature.

`scripts/runtime-deps.test.mjs` pins the set across the gate and its producers: a package the release stages but the gate does not require is one a packager can omit without the build failing, which is this bug returning by another route.

The docs commits update every `--all-features` recipe the gate invalidated — AGENTS.md's pre-push loop, the `rust-build` / `dev-loop` / `aube-bump` / `remote-build` / `implementation-thread` / `address-issue` / `orchestrator` playbooks, `scripts/rust-build.sh`'s usage header and `.worktreeinclude`.

## Verified

Build-script behaviour is not unit-testable, so the gate was exercised by hand with `cargo check -p nub-core --features embed-runtime`:

| Stage | Opt-out | Result |
| --- | --- | --- |
| none | unset | exit 101, names all five |
| partial (3 of 5) | unset | exit 101, names exactly the two absent |
| none | set | exit 0 |
| complete | unset | exit 0 |

`make verify`'s real clippy line, on a tree with `runtime/node_modules` deleted: exit 101 without the grant, exit 0 with it.

Against the shipped nub on Node 22.15 with `runtime/node_modules` removed from the extracted cache, `await using` fails with `MODULE_NOT_FOUND` and the three globals read `undefined`; restoring it fixes both.

On the merged head: `cargo fmt --check`, `cargo clippy --all-targets --all-features -- -D warnings`, `npm run test:scripts` (124 tests) and `npm run lat:check` all clean, plus a dev-binary smoke run of a decorator + `enum` fixture.

Refs #728


colinhacks:

PR #739, opened 2026-08-15
cli: parse a help flag after the verb in the pm and node groups

### LINKED ISSUE #653: nub pm shim --help runs the shim installer instead of printing help
### What happens

`nub pm shim --help` does not print help. It runs the shim installer: it creates six shim executables under `~/.nub/shims` and appends a `PATH` export to the user's shell profile.

```
$ nub pm shim --help
6 entries in <home>/.nub/shims (6 created)
  added <home>/.nub/shims to PATH in <home>/.profile
  restart your shell, or run: source <home>/.profile
```

The output is byte-identical to running `nub pm shim` with no arguments — the flag is ignored rather than handled. `-h` behaves the same way.

### What it writes

Starting from an empty home directory, the command leaves nine new entries:

```
~/.nub/shims/npm      ~/.nub/shims/npx
~/.nub/shims/pnpm     ~/.nub/shims/pnpx
~/.nub/shims/yarn     ~/.nub/shims/yarnpkg
~/.profile
```

and `~/.profile` gains:

```sh
# nub shims
export PATH="$HOME/.nub/shims:$PATH"
```

So a user typing `--help` to find out what a command does ends up with `npm`, `pnpm` and `yarn` shadowed on their `PATH` in new shells.

### It is isolated to this subcommand

The neighbouring help paths are all well-behaved, which rules out a global argument-parsing problem:

| Command | Result | Home entries created |
|---|---|---|
| `nub --help` | prints help | 0 |
| `nub pm --help` | prints help | 0 |
| `nub pm shim --help` | **runs the installer** | 9 |
| `nub pm shim -h` | **runs the installer** | 9 |

`nub help pm shim` is not an alternative — it exits with `error: unexpected argument 'shim' found`. So there is currently no way to ask what `nub pm shim` does without it doing it.

### Why it is worth fixing beyond the surprise

`--help` is the conventional way to inspect an unfamiliar command before running it, and it is what people reach for precisely when they are unsure whether a command is destructive. This subcommand edits shell startup files, so the blast radius outlives the process.

Secondary, but relevant to anyone sandboxing a run: `~/.nub/shims` is derived from the home directory directly, so it is not reachable by `XDG_DATA_HOME` / `XDG_CACHE_HOME` / `XDG_CONFIG_HOME`. Redirecting all three still leaves the shims and the profile edit in the real home directory. (Happy to open that separately if it is a distinct concern — it only came up because it is what made the accidental write non-recoverable by redirection.)

### Reproduction

Released `@nubjs/nub@0.6.0` from npm, `node:26-slim` container, `HOME` pointed at a directory that starts empty:

```sh
docker run --rm node:26-slim bash -lc '
  npm install -g @nubjs/nub@0.6.0 >/dev/null 2>&1
  export HOME=/probe/home; mkdir -p $HOME
  find $HOME -mindepth 1 | wc -l     # 0
  nub pm shim --help
  find $HOME -mindepth 1 | wc -l     # 9
'
```

### Expected

`nub pm shim --help` and `nub pm shim -h` print the subcommand's help and exit without touching the filesystem.

colinhacks:
Every `nub pm` and `nub node` sub-verb ignored `--help`/`-h` and ran the verb. `nub pm shim --help` installed six shims and added a PATH block to every shell startup file; `nub pm unshim --help` removed them. The two `nub node` shim verbs behaved the same way.

The top-level scan stops matching nub's own flags once a subcommand is seen, so `nub run build --watch` reaches the script. Both groups bypass clap and guarded help on argv[0] only, so past it the flag was dropped.

Both now recognize a help flag at any position. Neither forwards argv to a child, and their argument consumers take a package-manager name or a version, so no help flag was valid there.

Closes #653

### DIFF
--- a/crates/nub-cli/src/agent/mod.rs
+++ b/crates/nub-cli/src/agent/mod.rs
@@ -51,8 +51,11 @@ const INDEX_SLUG: &str = "/docs";
 
 /// Entry point for `nub agent …`, dispatched from `dispatch_subcommand`.
 pub fn run(args: &[String]) -> Result<i32> {
+    // Same guard as the other two non-forwarding groups: a help flag counts after
+    // the verb too, so `nub agent docs --help` prints usage instead of failing on
+    // an unexpected argument, and `nub agent skill --help` stops dumping the skill.
     let verb = args.first().map(String::as_str);
-    if matches!(verb, None | Some("help") | Some("--help") | Some("-h")) {
+    if verb.is_none() || crate::cli::group_help_requested(args) {
         print_usage();
         return Ok(0);
     }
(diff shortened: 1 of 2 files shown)


colinhacks:

PR #740, opened 2026-08-15
pm: honor NUB_CACHE_DIR for the engine cache, and report the env tier in nub config

### LINKED ISSUE #654: cacheDir env aliases (npm_config_cache_dir / NPM_CONFIG_CACHE_DIR) are silently ignored; only the .npmrc form works
### What happens

`cacheDir` can be set through a project `.npmrc`, and that works. The environment-variable forms of the same setting are accepted silently and have no effect.

Measured on released `@nubjs/nub@0.6.0`, one knob per container, one real `nub add lodash` each, counting files in the override directory and in the default cache afterwards:

| How `cacheDir` was set | Override dir | `~/.cache/nub/pm` |
|---|---|---|
| `npm_config_cache_dir=<dir>` (env) | **0 files** | 1056 files |
| `NPM_CONFIG_CACHE_DIR=<dir>` (env) | **0 files** | 1056 files |
| `cache-dir=<dir>` in project `.npmrc` | 1053 files | 3 files |
| `cacheDir=<dir>` in project `.npmrc` | 1053 files | 3 files |

So the setting is plainly wired up — the two `.npmrc` spellings both relocate the cache correctly. It is specifically the environment path that does not take.

Nothing is printed in the env cases: no warning, no "unknown setting", no note that the value was discarded. The install succeeds and writes to the default location.

### Why this is worth more than a missing feature

Documented precedence is **env above npmrc**. Here the higher-precedence source is the one that does not work, which inverts the outcome rather than merely losing a knob:

Someone who sets `npm_config_cache_dir=/fast/nub` in CI *and* `cache-dir=/project-local` in `.npmrc` will get `/project-local` while reasonably believing the environment override won. Nothing in the output contradicts them. That is a harder failure to spot than the knob simply not existing, because the cache does move — just not to where they asked.

`nub config list` is consistent with the actual behaviour (it shows the `.npmrc` value and shows nothing for the env vars), so it does not mislead — but it also will not tell you your environment variable was dropped unless you already suspect it and go looking.

### The general shape

This is the same class as a package manager accepting `npm_config_store_dir` and quietly doing nothing with it: the tool recognises the knob's *name* well enough to not complain, but does not honour it. A one-line warning when a recognised setting arrives on a source that is not consulted would turn a silent wrong-directory install into an obvious misconfiguration. That seems more valuable than the fix itself, since the same pattern can recur for any setting whose source list and read path drift apart.

### Reproduction

```sh
docker run --rm node:26-slim bash -lc '
  npm install -g @nubjs/nub@0.6.0 >/dev/null 2>&1
  export HOME=/probe/home; mkdir -p $HOME
  export npm_config_advisory_check=off
  export npm_config_cache_dir=/probe/over/c ; mkdir -p /probe/over/c
  mkdir -p /tmp/p && cd /tmp/p && echo "{\"name\":\"t\",\"version\":\"1.0.0\"}" > package.json
  nub add lodash --allow-low-downloads >/dev/null 2>&1
  echo "override: $(find /probe/over/c -type f | wc -l) files"
  echo "default : $(find $HOME/.cache/nub/pm -type f | wc -l) files"
'
```

Prints `override: 0 files` / `default: 1056 files`. Swapping the env var for a `.npmrc` line containing `cache-dir=/probe/over/c` prints `override: 1053 files` / `default: 3 files`.

### Expected

Either the environment aliases relocate the cache the way the `.npmrc` forms do, or nub warns that a recognised setting arrived on a source it does not read.

colinhacks:
Closes #654.

Env-set `cacheDir` was inert in 0.6.0: an `.npmrc`-only gate ran ahead of the settings resolver. #621 dropped it and v0.7.0 shipped the fix, with nothing pinning it. Two related defects were still live.

- `resolved_cache_dir` reads `config_env("CACHE_DIR")` ahead of the settings chain, so `NUB_CACHE_DIR` moves the engine cache, not just the resolver primer. Byte-identical for standalone aube.
- `nub config get` / `config list` gained the env tier. They read files only, so an install moved the cache while `config get cache-dir` printed `undefined`. Only config-carrying vars surface; ambient ones (`CI`, `http_proxy`) do not, matching npm. Scoped reads unchanged.
- New `pm_env_matrix` test pins both. Four stale comments corrected; the cache directory is now documented.


colinhacks:

PR #741, opened 2026-08-15
fix(linker): never let a .bin wrapper resolve to itself

### LINKED ISSUE #656: nub install node@26.5.1 hangs forever: the .bin/node shim execs itself
`nub install node@26.5.1` never terminates: the generated `.bin/node` shim execs itself.

## What happens

`node@26.5.1` ships a bin named `node`. The POSIX shim generator prefers a locally installed interpreter:

```sh
basedir=$(dirname "$0")
if [ -x "$basedir/node" ]; then
  exec "$basedir/node" "$basedir/../.store/node@26.5.1/node_modules/node/bin/node" "$@"
```

`$basedir` is `node_modules/.bin`, so `$basedir/node` **is the shim**. It execs itself, forever.

Measured on macOS/arm64: `ps` shows `/bin/sh .../node_modules/.bin/node .../node_modules/.bin/node` running its own path, and the install spins past two minutes with no output. Unrelated to the build jail — reproduced with `dependenciesMeta.node.sandbox = false`.

## Reproduce

```sh
mkdir -p /tmp/nodehang && cd /tmp/nodehang
echo '{"name":"f","version":"1.0.0","dependencies":{"node":"26.5.1"}}' > package.json
nub install       # never returns
```

## What other package managers do

npm installs it (exit 0). pnpm installs it and explicitly skips the shim, warning `Failed to create bin at .../.bin/node` — the target does not exist until the package's own `preinstall` downloads it.

## Two defects, one masking the other

**1. The self-exec (fixed on `sandbox/integration`).** Guarded by comparing `$0`'s basename against the interpreter name, in the shell rather than the generator, so every caller is covered without a signature change. Applied to both `generate_posix_shim` (unix) and `generate_sh_shim` (windows) — the two carry the same template and differ only by a `#[cfg]`.

**2. Still open: `failed to detect Node version`.** With the hang fixed the install terminates and fails cleanly, because nub runs `.bin/node --version` against a package whose binary the `preinstall` has not downloaded yet. Two candidate directions, neither validated:

- follow pnpm and skip creating a bin shim whose target does not exist at link time, or
- do not treat a dependency named `node` as a Node installation to probe.

## Scope

Any package whose bin name equals its shebang interpreter hits (1). `node` is the obvious case; the pattern is not limited to it.

colinhacks:
`nub install node@26.5.1` never terminated. That package's bin is named `node`, so its wrapper landed at `.bin/node` and preferred `$basedir/node` — itself. Measured exit 124 after 100s.

Three fixes in the same area, previously split across #751 and #753 and now folded in here.

**The hang.** Two independent defects:

- The linker declines an *unscoped* bin whose name is its interpreter — both its launch paths resolve back to it. A scoped `@scope/node` is not: it lands in `.bin/@scope/`, which PATH never carries. The post-build relink writes the right wrapper after.
- Node discovery accepted `.bin/node` as the project runtime; it now skips `node_modules/.bin`.

The interpreter preference also broke every other bin there. It now needs a native delegate, and the probe fails closed.

**The red release gate.** `Verify version consistency` has failed on every push to `main` since #735, which added a Varlock enum to `latest.json` without refreshing the pinned `v0.7.json`. It gates the release workflow, so nothing could ship — including this. Refreshed the snapshot, preserving its own `$id`.

**The missing PR-time gate.** `make version-check` ran only in `release.yml`, which is why that drift merged green and broke on `main`. Added a `Version consistency` job to `ci.yml`, wired into the `CI gate` aggregate.

Verified: control hangs, fixed exits 0 in 5s; Windows arm cross-checked. Version gate red-controlled — exit 2 against the drifted snapshot, exit 0 with it refreshed.

Closes #656


colinhacks:

PR #743, opened 2026-08-15
outdated: apply minimumReleaseAge to both reported versions

### LINKED ISSUE #722: [Feature Request] Take minimumReleaseAge into account for nub outdated
Currently, when running `nub outdated`, we get:

```
Package         Current  Wanted  Latest
@biomejs/biome  2.5.7    2.5.8   2.5.8
```

However, seeing the same output again right after running `nub update` is confusing when `minimumReleaseAge` is configured.

I think it would be better to either:

- Follow pnpm's approach: Hide updates that do not meet the `minimumReleaseAge` requirement.
- Display the age explicitly.

For example:

```
Package         Current  Wanted  Latest  Age
@biomejs/biome  2.5.7    2.5.8   2.5.8   8h
```

nub v0.7.5

colinhacks:
`nub outdated` read `Wanted` from the manifest range and `Latest` from the dist-tag, consulting `minimumReleaseAge` for neither. Nub enables that window by default and pins it strict, so the report offered upgrades `nub update` declined, and exited 1 permanently.

Both columns now resolve through `pick_version_for_add`, the picker `add` and `install` use. A held version keeps its row, is marked, and is named with the time it becomes installable. A project whose only pending upgrade is inside the window exits 0.

Closes #722

Verified: reporter's fixture goes exit 1 to 0; window off is unchanged; `minimumReleaseAgeExclude` un-holds the row; a mature upgrade beside a held one still exits 1. 16 unit tests plus a network-gated integration test that drives the binary. Clippy and fmt clean.

### DIFF
--- a/crates/nub-cli/tests/info_engine.rs
+++ b/crates/nub-cli/tests/info_engine.rs
@@ -478,6 +478,56 @@ fn outdated_reports_registry_drift_and_exits_one() {
     assert_no_engine_branding(&[("stdout", &stdout), ("stderr", &stderr)]);
 }
 
+/// `outdated` under a release-age window that admits nothing (#722).
+///
+/// The report used to offer an upgrade the resolver would then refuse, so it
+/// exited 1 forever and no `nub update` could clear it. The columns now agree
+/// with the resolver, the refused version is disclosed rather than dropped,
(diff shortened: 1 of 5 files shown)


colinhacks:

PR #744, opened 2026-08-15
run: signal every concurrent child on Ctrl-C, and make --color take effect

### LINKED ISSUE #685: [Fetaure Request] Improve concurrent / parallel run
### Without nub

Usually we use [concurrently](https://github.com/open-cli-tools/concurrently) to run multiple scripts:

```json
{
  "scripts": {
    "serve": "concurrently --names '🤖,🌼' 'pnpm --filter backend serve' 'pnpm --filter frontend serve'"
  }
}
```

### With nub 0.7.0

We can remove the `serve` script from the main `package.json` and do:

```sh
nub run -r serve
# or
nub run -r --parallel serve
```

But we lose the color output (`--color=always` doesn't work) and we cannot customize the prefix (we get something like `frontend serve: ...` for each line).

### Feature requests

- Allow color output in concurrent / parallel mode
- Allow to customize the prefix to reduce the output noises
- Allow to distribute SIGINT to all running script once (when pressing Ctrl+C)

colinhacks:
Fixes #685 plus four adjacent defects found while verifying it. Measured against pnpm 10.15.1. Recursive concurrency defaults to `min(4, cpus)`, so this hits plain `nub run -r`, not only `--parallel`.

- Ctrl-C signalled one child: a single `AtomicI32` target each spawn overwrote, zeroed by the first exit. Siblings were orphaned and nub hung. Now a set; `group_on_spawn` untouched, so #26/#27/#463 hold.
- `--color=always` was inert. Now exports `FORCE_COLOR=1` to scripts and `--no-color` exports `0`, matching pnpm; `--no-color` also parses after the verb, as pnpm allows.
- `NO_COLOR` was ignored, and `FORCE_COLOR` now follows Node's value table (`''` and `false` were both wrong).
- `--reporter-hide-prefix` stripped nub's own `Done` label; pnpm keeps it.

Reverting the signal fix turns three tests red. 12 new tests.


colinhacks:

PR #752, opened 2026-08-17
pm/node: honor XDG_DATA_HOME for the shim dirs on a fresh install

### LINKED ISSUE #750: Shim directories ignore XDG_DATA_HOME while the cache and config roots honor XDG
`nub pm shim` and `nub node shim` derive their install directory from the home directory alone, so redirecting the XDG variables still writes to `~/.nub/shims` and edits the real shell profiles.

```sh
XDG_DATA_HOME=/probe/data XDG_CACHE_HOME=/probe/cache nub pm shim
# 10 entries under $HOME, 0 under either
```

nub honors XDG elsewhere: `cache_dir()` reads `XDG_CACHE_HOME`, `config_dir()` reads `XDG_CONFIG_HOME`, and the profile writer reads `XDG_CONFIG_HOME` to find fish's config. Only the two shim dirs ignore it.

Redirecting `HOME` does contain nub fully, so this is a consistency gap, not a containment hole. Raised by @jdalton in #653.

colinhacks:
A fresh install now uses `$XDG_DATA_HOME/nub/<leaf>` when that variable is set. An existing `~/.nub/<leaf>` always wins: relocating an installed shim dir would leave the profile's PATH line naming the old path, so a bare `npm` would silently keep running pre-upgrade binaries.

Three consumers assumed a fixed path and are fixed with it:

- `invoked_as_persistent_node_shim` — a mismatch would have made the global `node` AUGMENT every node process.
- `which_node`'s recursion guard — a mismatch would have let nub resolve its own shim as the real node.
- Removal — unshimming an XDG install from a shell without the variable orphaned the binaries.

Closes #750


colinhacks:

PR #753, opened 2026-08-17
ci: gate version consistency on pull requests, not just releases

colinhacks:
`make version-check` runs in exactly one place — `release.yml`'s verify job — which fires on `v*` tags and on pushes to main, i.e. always after the offending merge. So #735 edited `site/public/schema/latest.json` without refreshing its pinned snapshot, merged green, and surfaced only on main, where it blocked every release until #751.

Adds a `Version consistency` job and wires it into the `CI gate` aggregate, in the same shape as `oxc lockstep`: checkout plus one Make target that shells a single `node -e`. No toolchain to install, no path filter to keep in sync — the drift it catches spans `package.json`, `Cargo.toml`, `runtime/version.mjs` and the published schema snapshots at once.

Verified: the command exits 2 on current main, catching the live drift, and 0 with #751 applied. `actionlint` clean.

## Merge #751 first

A `pull_request` checkout resolves `refs/pull/N/merge` — base merged with head — so this gate evaluates against current `main`, which fails `make version-check` today. Landing this before #751 would turn `CI gate` red on every other open PR until the snapshot is refreshed. #751 itself stays mergeable, since its own merge ref carries the fix.

This PR's own `Version consistency` job is red for that same reason, which is the gate catching the bug it exists for. No rebase is needed once #751 lands — the merge ref already reflects `main`, so re-running the checks is enough.

Refs #751

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1288,6 +1288,23 @@ jobs:
           persist-credentials: false
       - run: node scripts/check-oxc-lockstep.mjs
 
+  # `make version-check` also runs in release.yml's verify job, but that fires on
+  # `v*` tags and on pushes to main — i.e. always AFTER the offending merge. #735
+  # edited site/public/schema/latest.json without refreshing its pinned snapshot,
+  # merged green, and surfaced only on main, where it blocked every release until
+  # #751. This job is what makes that fail on its own pull request instead. Same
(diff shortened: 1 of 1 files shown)


colinhacks:
Folded into #741, which now carries all three fixes in one PR. Closing in favour of it.

PR #756, opened 2026-08-17
ci: make the embed fixtures able to detect a missing vendored runtime

colinhacks:
oxc inlines `enum`, so the round-trip fixture never touched the vendored `node_modules` in the blob. Two `embed-runtime` binaries differing only in whether `runtime/node_modules` was staged:

| Fixture | Complete | Empty |
| --- | --- | --- |
| `enum` only | exit 0 | exit 0 |
| plain `.js` compile | exit 0 | exit 0 |
| legacy decorator | exit 0 | exit 1, `Cannot find module '@oxc-project/runtime/helpers/decorateMetadata'` |

The pre-publish gate gets the same check, because `smoke.sh`'s decorator case installs from npm and so runs only after publish.

Down-levelling a decorator is version-independent; resolving its helper from a CJS entry is not, below require(esm). The unpinned release gate therefore uses an ESM entry; `ci.yml`'s fixture stays CJS because that job pins Node 24. `macos-embed-roundtrip` builds incomplete deliberately and is untouched.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -452,20 +452,36 @@ jobs:
           if (-not (Test-Path $nub)) { throw "nub.exe not found at $nub" }
 
           # Isolated work dir (no package.json/.nvmrc) so nub uses the PATH Node and
-          # no project pin interferes. `enum` is NON-ERASABLE, and that is the whole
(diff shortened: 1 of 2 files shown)


colinhacks:

PR #297, opened 2026-07-03
feat(node): `nub node shim`/`unshim` — persistent global node shim

colinhacks:
Opt-in persistent global `node`→nub shim so a fresh machine with no Node can use nub as its default node runner.

Vanilla by default (signed off 2026-07-03): the shimmed `node` does version resolution + provisioning only, no augmentation — per the node-hijack contract and to bound the blast radius of a machine-wide `node`. The per-invocation hijack still augments; `--node`/`NODE_COMPAT` unchanged.

Reuses the PM-shim engine (parameterized on a `ShimBlock`; PM block unchanged), a dedicated `~/.nub/node-shim` dir + marker, a `which_node` recursion guard, and upgrade re-link parity. Tests + docs included.

Refs #294

https://claude.ai/code/session_01Xuh9u8b93eMNu53fTQTWX7

PR #757, opened 2026-08-17
docs: point the node-shim fresh install at the tap

colinhacks:
`brew install nub` resolves nothing today — Nub is not in Homebrew's core tap, so the command errors. This is the `nub node shim` fresh-install story, aimed at a reader whose machine has no Node, which makes it the worst place for an install command that fails.

Every other install surface already says `nubjs/tap/nub`. This one was missed when `nub node shim` landed in #297.

`tsc --noEmit` exits 0.


colinhacks:

Issue #258, opened 2026-06-30
ERR_NUB_UNSUPPORTED_ENGINE


that70schris:
Running `nub install` getting `ERR_NUB_UNSUPPORTED_ENGINE` when `package.json#engines>node:24` and I'm on node `26.3.1` it's supposed to respect that right?

PR #760, opened 2026-08-17
fix(resolver): emit the popular-names corpus for source builds too

colinhacks:
Release builds generate `data/popular-top100000-v1.json` before compiling, so they never reach `generate()`. Every source build does — homebrew-core, Nix, Fedora, AUR — and without it `AUBE_POPULAR_NAMES_RANKED` stays unset, so `nub add`'s similar-name check declines to run.

`nub add expresss`:

| Binary | Result |
| --- | --- |
| release | `SIMILAR_PACKAGE_NAME`, rank #258 |
| source, before | `LOW_DOWNLOAD_PACKAGE` — check never ran |
| source, after | `SIMILAR_PACKAGE_NAME`, rank #258 |

Best-effort, after the primer succeeds: offline builds return before reaching it, failure is swallowed, and the host is one the primer already needs.

### DIFF
--- a/vendor/aube/crates/aube-resolver/build.rs
+++ b/vendor/aube/crates/aube-resolver/build.rs
@@ -191,9 +191,66 @@ fn generate(manifest_dir: &Path, source: &Path, top: usize) -> bool {
         return false;
     }
 
+    generate_popular_names(workspace, manifest_dir);
+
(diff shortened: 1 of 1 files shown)


colinhacks:

PR #761, opened 2026-08-17
Harden source builds of nub against silent degradation

colinhacks:
Supersedes #738, #756, #757, #760. homebrew-core builds nub from source (#728), and a source build can come out quietly degraded.

- `nub-core/build.rs` — an `embed-runtime` build with no vendored `node_modules` now fails instead of shipping a runtime that cannot resolve transpile helpers.
- `aube-resolver/build.rs` — source builds emit the popular-names corpus, restoring `nub add`'s similar-name refusal for Homebrew, Nix, Fedora, AUR.
- `ci.yml` / `release.yml` — the embed fixture and pre-publish gate now pull a helper, so they can see that degradation.
- `runtime-deps.test.mjs` — pins the vendored set across `build.rs` and its producers.
- `docs/node.mdx` — `brew install nub` resolves nothing today.

Refs #728.


colinhacks:

PR #762, opened 2026-08-17
docs: correct the env-precedence attribution — the ecosystem is split, not a consensus

### LINKED ISSUE #732: docs: env precedence wrongly attributed to dotenv-flow/Vite (behavior actually matches Next.js/Bun) in three places
The .env precedence is inverted relative to the dotenv-flow / Vite convention Nub claims to mirror. .env.local overrides the mode file, so a production build can silently pick up local overrides.

Repro. Project with these four settings:

```
.env:                 FOO=base
.env.local:           FOO=local
.env.production:      FOO=prod
.env.production.local: FOO=prodlocal
```

Run `NODE_ENV=production nub main.ts` where main.ts prints process.env.FOO:

- With all four files, FOO prints `prodlocal` (correct)
- Delete .env.production.local, FOO prints `local`, not `prod` (wrong)
- Same inverted behavior for development: with .env.development (FOO=dev) plus .env.local, FOO prints `local`, not `dev`

So .env.local beats .env.{mode}. In Vite and dotenv-flow the order is `.env` then `.env.local` then `.env.{mode}` then `.env.{mode}.local`, with the mode file overriding .env.local. Nub has local above mode.

Source. `crates/nub-core/src/workspace/env.rs`, `env_file_names_for_mode` builds the list:

```
.env.{mode}.local
.env.local
.env.{mode}
.env
```

and the merge is first-writer-wins (`result.entry(key).or_insert(value)`). So .env.local is treated as higher priority than .env.{mode}. The comment above it says the cascade mirrors the dotenv-flow / Next / Vite convention, but that convention puts the mode file above .env.local.

Effect. A machine-specific .env.local can silently override environment-specific values, and the environment-specific file never wins unless .env.{mode}.local also exists. Would expect .env.local to load before .env.{mode} (mode file wins), matching Vite.

colinhacks:
Nub's `.env*` precedence matches Next.js and Bun. Five surfaces claimed it mirrored one ecosystem convention that also covers Vite and `dotenv-flow`, which rank `.env.{mode}` above `.env.local`. Checked against Vite v8.1.2, dotenv-flow v4.1.0, @next/env 16.3, Bun bbe3f6a2.

No behavior change; the ordering and the tests pinning it are untouched.

Corrects the claim in `env.rs`, `introducing-nub.mdx`, `env-file-loading.md`, `docs/runtime/env.mdx` and `docs/watch.mdx`. The last two were not in the report.

Closes #732

### DIFF
--- a/crates/nub-core/src/workspace/env.rs
+++ b/crates/nub-core/src/workspace/env.rs
@@ -145,9 +145,11 @@ const CLAMPED_NODE_ENV_MODES: &[&str] = &["development", "production", "test"];
 
 /// The `.env*` filenames Nub loads, in descending priority order (the file
 /// listed first wins a key over later ones). Driven by the resolved *mode*. The
-/// `.env` / `.env.local` / `.env.{mode}` cascade mirrors the dotenv-flow / Next /
-/// Vite ecosystem convention — NOT Node core, which has no mode cascade (its
-/// `--env-file` loads named files only). Shared by [`load_env_files`]
+/// four-file taxonomy is ecosystem-wide, but the ecosystem SPLITS on where
+/// `.env.local` sits: Next.js and Bun rank it above `.env.{mode}`, Vite and
+/// dotenv-flow below. Nub follows Next.js / Bun, so `next build` under Nub sees
+/// the environment `@next/env` would have computed. Node core has no mode cascade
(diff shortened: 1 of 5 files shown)


colinhacks:

PR #763, opened 2026-08-17
lockfile: treat an empty importer as drift, not as a specifier-less format

### LINKED ISSUE #657: Install reports "Already up to date" when package-lock.json omits a declared dependency
When `package-lock.json` omits a dependency declared in `package.json`, `nub install` skips it, prints `✓ Already up to date`, and exits 0. npm installs it and rewrites the lockfile.

Manifest declares `"is-odd": "3.0.1"`; lockfile has only `"packages": { "": { … } }`.

```
$ nub install
nub 0.6.0 · ✓ Already up to date
$ ls node_modules/is-odd
No such file or directory
```

A lockfile lagging the manifest is ordinary (hand edit, merge, rebase). `npm install` reconciles, `npm ci` refuses; both are actionable.

`ERR_NUB_LOCKFILE_DECLARATION_MISMATCH` in `pm_engine/mod.rs` fires on other inputs, so the question is why not here.

nub 0.6.0, macOS arm64.

colinhacks:
The importer drift check skipped itself when every recorded direct dep carried no specifier — how a yarn berry lockfile parses. That is vacuously true on an empty list, so an importer recording nothing skipped too and reported Fresh whatever package.json declared: `nub install` linked nothing and printed "Already up to date", `nub ci` exited 0 with an empty tree.

Requiring one recorded dep before the skip closes it for every format; a depless importer still reads Fresh.

Closes #657

Verified: four lockfile shapes now install, `nub ci` exits 16 with ERR_NUB_OUTDATED_LOCKFILE, warm and depless projects unchanged. 563 + 12 tests pass; reverting the guard turns the new test red.

### DIFF
--- a/crates/nub-cli/tests/install_engine.rs
+++ b/crates/nub-cli/tests/install_engine.rs
@@ -1638,3 +1638,59 @@ fn workspace_spec_naming_a_non_member_fails_without_reaching_the_registry() {
         );
     }
 }
+
+/// A lockfile that records the root importer with ZERO direct deps while
+/// `package.json` declares one must read as drift, in every format
+/// (nubjs/nub#657). Two shapes reach it: a `package-lock.json` whose root
+/// entry declares a dep with no matching `node_modules/<name>` package node,
+/// and a `pnpm-lock.yaml` written as `importers: { .: {} }`. Both used to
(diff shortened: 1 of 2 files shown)


colinhacks:

PR #768, opened 2026-08-19
fix(tsconfig): keep the local compilerOptions when an extends target is missing

### LINKED ISSUE #731: Unresolvable tsconfig extends silently drops the whole compilerOptions block
When `tsconfig.json` has an `extends` pointing at a file that cant be resolved, Nub silently discards the whole `compilerOptions` block. Path aliases stop working, decorator flags stop working, and nothing reports the problem.

Repro:

```json
// tsconfig.json
{ "extends": "./does-not-exist.json", "compilerOptions": { "baseUrl": ".", "paths": { "@/*": ["./*"] } } }
```

```ts
// main.ts
import { thing } from "@/util";
console.log(thing);
```

```ts
// util.ts
export const thing = "alias-resolved";
```

Run `nub main.ts`:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package @/util imported from ...
```

The `@/*` alias resolves fine with a valid `extends`, and with no `extends` at all. Only the unresolvable `extends` breaks it. I ran the failing case five times and it fails every time.

The real issue is that a failed `extends` is treated the same as "user set no compilerOptions", so paths, decorators, everything silently falls away. `tsc` reports `TS5083: Cannot read file` for the missing config but still applies the local `compilerOptions`. Nub reports nothing and applies none of them.

It would help if a broken `extends` either surfaced an error or fell back to the local `compilerOptions` like `tsc` does, instead of silently dropping all of it.

colinhacks:
A failed `extends` propagated out of `inner_parse`, so `build_loaded` returned the "no tsconfig" shape and dropped the whole file — `paths`, decorator flags and `customConditions` alike.

| Case | Before | After |
| --- | --- | --- |
| `extends` + `paths` | ERR_MODULE_NOT_FOUND | resolves |
| `extends` + decorators | false "Stage 3 decorators" error | runs |
| `extends` + `customConditions` | exit 0, wrong module | correct module |
| invalid root JSON | silent | reports why |

The loop now records and skips the unresolvable base, keeping what the file declares, as `tsc` (TS5083) and bun do. Diagnostics reach stderr once per config path.

Closes #731

### DIFF
--- a/crates/nub-cli/src/cli.rs
+++ b/crates/nub-cli/src/cli.rs
@@ -3809,6 +3809,18 @@ fn npmrc_script_node_options(project_root: &Path) -> Vec<String> {
         .unwrap_or_default()
 }
 
+/// The tsconfig warnings this process already wrote, ready to hand to the child.
+///
+/// Both the CLI and the addon parse the project's tsconfig, so without this the
+/// user reads the same warning twice — the exact problem that made tsx drop its
+/// warning entirely. Call it AFTER the node options are built, since that is what
+/// parses the config. A spawn path that skips this prints the warning twice rather
+/// than losing it, so missing one degrades the output instead of the behavior.
+fn tsconfig_reported_env() -> Option<(String, String)> {
+    let reported = nub_tsconfig::reported_config_paths();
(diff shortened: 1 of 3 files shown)


colinhacks:

PR #770, opened 2026-08-19
node: auto-unflag `import defer` on Node 26.4+

colinhacks:
`import defer` now works on Node 26.4+ with no flag or config.

V8's `--js-defer-import-eval` is refused in `NODE_OPTIONS` and absent from `allowedNodeEnvironmentFlags`, so the `Unflag` shape drops it when the probe succeeds and aborts startup when it does not. Adds `Mitigation::UnflagArgv`, which rides argv beside `v8Flags`.

The floor is a Node floor — Node wired the phase through in nodejs/node#63712. Measured: 25.9 rejects the flag, 26.0-26.3 abort in `to_phase_constant`, 26.4/26.5 defer.

The flag enters V8's flag hash but never `NODE_OPTIONS`, so #246 cannot recur. Below the floor the error stays a clean `SyntaxError`. Compat mode and `--no-js-defer-import-eval` suppress it. Dynamic `import.defer()` aborts upstream, identically on plain `node`.


colinhacks:

PR #774, opened 2026-08-19
runtime: let a global envFile displace a .env.schema too

colinhacks:
A `.env.schema` is a default: it decides the environment exactly when nothing else does. So a declared `envFile` at any scope displaces it, the global layer included, and `CLI > environment > project > global > defaults` runs unchanged.

This reverses the one carve-out shipped in #735, where a global `envFile` could not displace a project's schema.

## Why the carve-out was wrong

It was wrong twice, and both reasons are about coherence rather than the failure mode it was avoiding.

A schema is not a config field and does not sit in the precedence chain. Exempting one layer put a non-config signal at a novel position *inside* that chain — beating `global`, losing to `project` — a position no other key holds, and one that cannot be stated without describing the wedge.

And #735 exists to stop an explicit `envFile` being silently ignored. Carving out one layer where it still is contradicts the premise: `envFile: false` meaning "load nothing" is not worth much if a layer quietly exempts itself.

Accepted cost: a global `false` now empties a schema project. That fails loudly at an undefined variable rather than silently, and one project line fixes it.

## `envFile: "varlock"` becomes load-bearing

Under the carve-out it was purely declarative — an absent `envFile` reached the same place. It is now how a project keeps the hand-over when a machine-wide `envFile: false` is set, which is what the test added here pins.

The accessor is renamed `scoped_env_file_setting` → `declared_env_file_setting`, since no scope filter remains in it.

## Relative paths, documented rather than changed

Also adds a `Relative paths` section to the config reference. A relative path resolves against the directory of the file that set it, so a relative path in the global file points inside your config directory — nearly never what is meant.

That matches uv, Cargo, ESLint and tsconfig's default. Cargo documents the same footgun in a note rather than special-casing it, and this does the same. Considered and rejected: resolving global paths against the project root (pnpm's `store-dir` model, inconsistent with how a project file's own paths resolve), and a `${projectDir}` token (tsconfig's `${configDir}` model, speculative surface). The survey is recorded in the research doc.

Only three fields are path-valued — `preload`, `envFile`, `tsconfig` — so the policy stays cheap to revisit.

## Pre-existing failures on the remote builder

Two unrelated red signals showed up while gating this, both established by control rather than assumed:

- Three `nub-launcher` tests fail identically on pristine `origin/main` (70 passed, 3 failed) — pre-existing, not from this change.
- `brand_boundary_no_globals_no_env` fails only under `scripts/remote-build.ts`, which began exporting `NUB_ALLOW_INCOMPLETE_RUNTIME=1` in #761 today. The same test passed on the same harness before that commit. CI does not set that variable for its test jobs.


colinhacks:

PR #775, opened 2026-08-19
linker: claim hidden-hoist names shallowest-first, matching pnpm's depth ordering

colinhacks:
Hidden-hoist alias claims were made by iterating `graph.packages`, keyed by dep_path — lexicographic, tracking nothing about the graph (`1.0.0` beat `2.0.0`; `10.0.0` beat `9.0.0`). Now root direct deps claim first, then shallowest wins, dep_path breaking ties.

The depth tier matches pnpm; the tie-break does not. pnpm sorts parents and walks their children, so its effective key is the parent's `(depth, node id)` — two same-depth parents bringing different versions resolve differently there than here.

- `dependency_depths` seeds from every importer; `root_deps()` reads only `importers["."]`, leaving a member's direct dep unreachable.
- It walks optional edges too: `LockedPackage` documents active optionals as mirrored into `dependencies`, but `yarn/berry.rs` does not merge them. The readers are the underlying bug; that is a separate change.
- Adds a `hidden_hoist_algo` settings-hash salt; without it the change is inert on warm trees. One-time relink.

Claim order also changes for the GVS-off tree; intended.

### DIFF
--- a/vendor/aube/crates/aube-linker/src/link.rs
+++ b/vendor/aube/crates/aube-linker/src/link.rs
@@ -1595,8 +1595,11 @@ impl Linker {
         //     shared store, so its walk ascends the store, never the project — it
         //     can NEVER consume this tree. So the ejected subset gets pnpm's
         //     blanket phantom tolerance while the symlinked majority is untouched.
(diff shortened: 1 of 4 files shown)


colinhacks:

PR #776, opened 2026-08-20
compile: probe the artifact from a short path when --out is deep

colinhacks:
`nub compile --out <deep path>` builds a working binary then fails verifying it: the self-probe cannot spawn the staged artifact.

Measured on Windows Server 2022 with the gate's own binary — `--out` at 285 chars compiles clean, at 351 the staged image fails with os error 3. It reproduces both where `LongPathsEnabled` is 0 (a default Windows box) and 1 (GitHub runners), so long-path support does not decide it. `\\?\` makes it worse: measured from Rust it turns a working spawn into os error 123.

The probe asks whether the produced bytes run, not where they sit, so it retries from a short copy. The staged artifact does not move — its directory shares a filesystem with the destination so publish stays an atomic rename. An ordinary `--out` copies nothing.

Evidence: branch `probe-win-longpath-spawn`.

### DIFF
--- a/crates/nub-cli/src/compile/mod.rs
+++ b/crates/nub-cli/src/compile/mod.rs
@@ -1523,10 +1523,35 @@ fn verify_artifact(bin: &Path, target: &TargetPlatform) -> Result<()> {
     } else {
         Path::new(".").join(bin)
     };
-    let out = std::process::Command::new(&bin)
-        .env("__NUB_COMPILED_LAUNCHER_MODE", "probe")
-        .output()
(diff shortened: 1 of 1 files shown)


colinhacks:
Closing: the premise is disproven. Windows CAN CreateProcess past MAX_PATH here — a Rust binary spawns a 297-char image fine, LongPathsEnabled=1, and the \\?\ prefix breaks a working spawn (os error 123). The real failure reproduces on a fresh compile, but path length, the .exe.tmp name, the staging dir and the artifact are each individually exonerated. Evidence on branch probe-win-longpath-spawn.

PR #777, opened 2026-08-20
node: keep version-gated flags out of NODE_OPTIONS so Electron survives

colinhacks:
`NODE_OPTIONS` is inherited by the whole process subtree, and nub's flag set matches the HOST Node. Any descendant on an older Node aborts at startup. Electron 42 embeds Node 24.17, so a host on 26.5 fed it `--experimental-import-text` and one on 26.3 fed it `--experimental-ffi`.

Measured against real Electron 42.5.1 — `nub run dev`, `nub exec electron` and `nubx electron` all exited 9, on released v0.7.4 and on main alike. Refs #246, the same family caught earlier only in its V8-snapshot SIGTRAP form.

Both NODE_OPTIONS builders now carry only flags that cannot abort an older descendant. The second matters because `.bin/electron` is itself a Node script, so `nub exec electron` reaches `spawn_node` and the real binary inherits that string.

The split is by floor, not by category. `--enable-source-maps` (floor 12.12, below nub's 18.19 support floor) stays. The feature flags move to argv, and so does `--disable-warning=ExperimentalWarning`, whose 20.11 floor is above that support floor — a latent instance of the same bug.

Argv delivery happens OUTSIDE the augment block, which is skipped on a re-entrant spawn — every `node` a script launches through the shim. Leaving it inside silently disabled every version-gated feature (`vm.Module` true → false).

Cost, accepted: a tool spawning Node by absolute path bypasses the shim and loses the version-gated features and ExperimentalWarning suppression. It keeps the preload and source-map remapping. Anything launched as `node` is unaffected.

Verified with a working positive control (#246 still reproduces as exit 133): all three entry points exit 0, `vm.Module` restored, real-apps green, nub-core 432 and nub-cli 1039 passed.


colinhacks:

PR #778, opened 2026-08-20
fix(tsconfig): refuse the run when the project's tsconfig cannot be read

colinhacks:
Follow-up to #768, which reported an unreadable tsconfig and carried on. Continuing still runs the program under options its author never wrote — `extends` is where a project factors out `strict`, `target` and its path aliases — so this refuses the run instead, as `tsc` (TS5083) does.

```console
$ nub main.ts
Nub: /app/tsconfig.json: File './base.json' not found.
Error: the project's tsconfig.json could not be read in full (see above).
Fix the config, or use --node to run without Nub's TypeScript features, path aliases included.
```

Two anchors: the CWD, and the entry file's own directory, so `nub sub/main.ts` from the parent answers the same as that project run from inside `sub/`.

The salvage from #768 stays and now only governs a config the CLI never saw — a dependency's own tsconfig reached mid-resolve by the addon, where aborting is not available.

Refs #731

### DIFF
--- a/crates/nub-cli/src/cli.rs
+++ b/crates/nub-cli/src/cli.rs
@@ -3950,6 +3950,36 @@ fn tsconfig_reported_env() -> Option<(String, String)> {
     (!reported.is_empty()).then(|| (nub_tsconfig::REPORTED_ENV.to_string(), reported.join("\n")))
 }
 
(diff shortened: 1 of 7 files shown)


colinhacks:

PR #779, opened 2026-08-20
outdated: honor the release-age window silently

### LINKED ISSUE #722: [Feature Request] Take minimumReleaseAge into account for nub outdated

colinhacks:
The release-age window is the project's own configured policy, and `install` and `update` honor it without remark. The report announced it instead — a `*` on each held cell and a footer with the version and its ETA — so one command editorialized about a policy the others simply apply. Under the 24-hour default that is most runs.

Now a version inside the window is not offered, and a project whose only pending upgrade sits inside it reports nothing and exits 0.

The gating that fixes #722 is unchanged: both columns still resolve through `pick_version_for_add`, and the absent-dist-tag guard stays. Silence removes the wall case, so `Hold`, `Pick`, the walled flags and the publish-time arithmetic all go — net 512 lines deleted.

Refs #722


colinhacks:

PR #780, opened 2026-08-20
settings: resolve the namespace in a default against the active embedder

colinhacks:
Refs #654.

`nub config list --all` advertised the cache default as `$XDG_CACHE_HOME/aube`, a directory nub never uses. The listing renders each setting's `default`, a build-time constant on the shared table, so a directory default written as a literal is baked with the engine's namespace and printed by whichever host embeds it.

Both directory defaults now carry `{cache_namespace}` / `{data_namespace}` tokens resolved against the active embedder at display time. Standalone aube renders what it did before; an untokenized default is returned borrowed and unchanged.

`storeDir` is tokenized though it does not leak today — nub's embedder default masks the literal rather than correcting it.


colinhacks:

PR #781, opened 2026-08-20
lockfile: resolve a read against the declared package manager, not filename precedence

colinhacks:
## The defect

In a project declaring npm, a `bun.lock` sitting beside `package-lock.json` made `nub install` rewrite `package-lock.json` without the `resolved`, `license` and `engines` of exactly the entries `bun.lock` also names. It also installed bun's pins rather than the project's own. Remove the `bun.lock` and the same fixture round-tripped byte-for-byte.

```sh
mkdir /tmp/repro && cd /tmp/repro
cp <nub-repo>/{package.json,package-lock.json,bun.lock} .
cp package-lock.json /tmp/before.json
nub install
diff -u /tmp/before.json package-lock.json   # 6 entries lose resolved/license/engines
```

The same shape reproduces with a stray `pnpm-lock.yaml` or `yarn.lock`, and with `npm-shrinkwrap.json`.

## Root cause

Reads and writes were resolving different files. The write path asks the declaration-aware `resolve_project_lockfile_kind`, which honors `packageManager` and `devEngines`. The read path walked candidates in raw filename precedence — `pnpm > bun > yarn > npm-shrinkwrap > npm` — and took the first that existed. Since `package-lock.json` sorts last, the stray file won the read while npm still won the write, so bun's thinner graph was serialized into npm's file.

Confirmed with `RUST_LOG=debug`: the install logged `peer-context pass (lockfile=Bun)` before, `lockfile=Npm` after. Nothing merged the two lockfiles — the npm one was never read.

The read candidates are now ordered so the declaration's family leads. Reordering rather than filtering keeps this from inventing a failure mode: when a declaration contradicts the disk, or several tools' lockfiles coexist undeclared, detection errors and the read falls back to the old precedence, leaving the write path to raise the error it already raised.

## Second defect, independent of any foreign lockfile

The npm writer's root importer entry mirrored only the manifest's dependency fields. Real npm also copies `license`, `bin` and `engines` there, so a genuine rewrite dropped all three even with no stray lockfile present. Both npm normalizations are reproduced: an object `license` collapses to its `type`, a string `bin` expands to `{ <name-without-scope>: <path> }`. Scoped to the root importer — a workspace member's entry has no manifest available on the path that matters.

## The signal-exit hoist: intended, left alone

The same rewrite moved `node_modules/write-file-atomic/node_modules/signal-exit` to `node_modules/signal-exit`. Differential on the same fixture:

| tool and state | placement |
| --- | --- |
| npm, existing lockfile kept | `node_modules/write-file-atomic/node_modules/signal-exit` |
| npm, no lockfile (fresh resolve) | `node_modules/signal-exit` |
| nub, genuine write | `node_modules/signal-exit` |

The npm writer recomputes the layout from the flat graph on every write and places each package in the shallowest legal slot. Only one version of `signal-exit` is demanded in the tree and the root slot is free, so the hoist is what npm itself produces on a fresh resolve. npm differs only in that its tree build is incremental and preserves an existing valid placement.

## Verification

- Reported reproduction: read resolves `Npm`, `package-lock.json` byte-identical.
- Genuine write (`nub add is-odd@3.0.1`) with `bun.lock` present: zero `resolved`/`license`/`engines` lost, against 17 fields before.
- Fresh project carrying `license`, `bin` and `engines`: nub's `package-lock.json` byte-identical to npm 11.17's.
- A 13-case declaration-by-lockfile matrix run against the merge-base build and this one: only the declared-with-stray cases changed; ambiguity and declaration-mismatch errors are untouched.
- Both new tests confirmed red with their fix reverted.
- Root clippy `--all-targets --all-features`, `cargo fmt --check`, root `cargo test` (59 binaries), and `cargo test --workspace` in `vendor/aube` (56 binaries) all pass.

### DIFF
--- a/vendor/aube/crates/aube-lockfile/src/detect.rs
+++ b/vendor/aube/crates/aube-lockfile/src/detect.rs
@@ -177,7 +177,7 @@ fn declared_lockfile_kind(name: &str) -> Option<LockfileKind> {
 /// The PM family a lockfile kind belongs to, for the ambiguity check.
 /// Kinds owned by the same tool (npm's shrinkwrap + package-lock,
 /// yarn classic + berry) never make a project ambiguous together.
-fn family(kind: LockfileKind) -> &'static str {
+pub(crate) fn family(kind: LockfileKind) -> &'static str {
     match kind {
         LockfileKind::Aube => "aube",
         LockfileKind::Pnpm => "pnpm",
--- a/vendor/aube/crates/aube-lockfile/src/io.rs
+++ b/vendor/aube/crates/aube-lockfile/src/io.rs
@@ -498,7 +498,7 @@ pub fn parse_lockfile_with_kind_and_options(
     options: ParseOptions,
 ) -> Result<(LockfileGraph, LockfileKind), Error> {
     reject_bun_binary(project_dir)?;
-    for (path, kind) in lockfile_candidates(project_dir, /*include_aube=*/ true) {
+    for (path, kind) in read_candidates(project_dir, /*include_aube=*/ true) {
         if !path.exists() {
             continue;
         }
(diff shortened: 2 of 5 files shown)


colinhacks:

PR #788, opened 2026-08-24
runtime: pass non-file URLs through the load hook untouched

colinhacks:
nub’s load hooks dispatch on the URL extension, and every branch they reach calls `fileURLToPath`. A non-`file:` URL ending in a plain-JS extension entered those branches and threw `ERR_INVALID_URL_SCHEME` — masking Node’s `ERR_UNSUPPORTED_ESM_URL_SCHEME` and breaking every custom-protocol ESM loader plain Node serves.

An extension is now reported only for a `file:` URL. The gate sits in `transform-core`, shared by both tiers.

Three corpus tests go 1 → 0 under nub on Node 26.7: `test-esm-dynamic-import.js`, `test-esm-loader-default-resolver.mjs`, `test-esm-loader-http-imports.mjs`. Both new tests were confirmed red before the fix.


colinhacks:

PR #789, opened 2026-08-24
env: stop expanding $VAR in --env-file values, matching Node

colinhacks:
Node's `--env-file` never expands variable references. Nub ran `expand_env_map` over the parsed map, so any value holding a literal `$` was silently rewritten: `A="{ port: $MISSING_VAR}"` arrived as `{ port: }`, and `PASSWORD=foo$bar` truncated to `foo`. The divergence survived `--node` — it was in the Rust CLI, not the preload.

The fix drops the expansion call on the explicit `--env-file` path. Auto-discovered `.env*` files are untouched and still expand, which is the split recorded in `wiki/research/env-file-loading.md`: expand in defaults, match Node for the Node-compat flag.

**This reverses a recorded maintainer verdict**, so it is your call. `wiki/research/nub-v0.5-augmentation-regressions.md` Cause 8 (2026-07-24) kept the expansion as an intentional bun-parity divergence. Grounds for reversing:

1. `AGENTS.md` holds byte-for-byte Node compatibility as paramount and forbids implementing a Bun-specific non-spec behavior for parity's sake. `--env-file` is Node's own flag, not a nub DX surface.
2. The premise was half-true: re-measured against bun 1.3.14, nub did not match bun either (`{ port: }` vs `{ port: `), so the divergence bought neither parity.
3. Any env value with a literal `$` — passwords, connection strings — was silently corrupted.

Cause 8 keeps the superseded verdict and adds the reversal; the changelog carries a `REVERSAL:` entry.

Verified: Node's `test/parallel/test-dotenv.js` goes 1 → 0. New regression test fails before, passes after. `cargo fmt --check`, `clippy --all-targets --all-features`, and `cargo test -p nub-cli` all green.


colinhacks:

PR #791, opened 2026-08-25
tests: make recursive clones skip the node-suite submodule

colinhacks:
Vercel clones every HTTPS-reachable submodule before a site build, so each deploy fetched all of nodejs/node (48,844 files) and then deleted it under .vercelignore. Clones on recent preview deploys took 1.5 to 32 minutes, and one hung for 46 minutes and failed the deployment.

- `.gitmodules` sets `update = none`, so recursive clones and a bare `git submodule update --init` skip `tests/node-suite`. A probe deploy on this branch cloned in 3.2s and removed 1,272 ignored files instead of 50,104.
- The consumers that need the corpus pass `--checkout`, which overrides the setting: the release workflow's gate, the two Rust corpus gates' hints, and `tests/run-node-compat.sh`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -502,7 +502,7 @@ jobs:
         with:
           persist-credentials: false
       - name: Init node-suite submodule (shallow)
-        run: git submodule update --init --depth 1 tests/node-suite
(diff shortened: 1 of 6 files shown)


colinhacks:

PR #797, opened 2026-08-27
runtime: keep the stack of a REPL module-resolution error

colinhacks:
The nub REPL printed a failed `require()` with no stack frames, where node prints 8:

```
> require("./bar")
Uncaught [Error: Cannot find module './bar' ...] { code: 'MODULE_NOT_FOUND' }
```

The REPL cuts a trace at the last null-named frame — normally its own eval frame. nub's wrappers reach the saved `Module._resolveFilename` / `Module._load` through `.call()`, which nulls those frames' names, and nub's added resolve frames push the REPL's own frame past the default `Error.stackTraceLimit` of 10. The delegated original was then the only null-named frame left.

Naming the originals also restores `at Module._resolveFilename` / `at Module._load`, which printed as `at Module.<anonymous>`. No frame added or removed.


colinhacks:

PR #798, opened 2026-08-27
node: keep Node's default test-file coverage exclusion when nub injects its own

colinhacks:
Node applies its default coverage exclusion (`kDefaultPattern` in `lib/internal/test_runner/utils.js`) only when no `--test-coverage-exclude` is set. Nub injects one to keep its preloaded runtime out of the report, which switched that default off, so `nub --test --experimental-test-coverage` listed the user's own test files where plain node does not.

Both injection sites now emit Node's default pattern beside the runtime exclude — but only on 23.5.0+, where Node has such a default (it was never backported to 22.x), and only when the user supplied no exclude of their own.

Known limit: a grandchild spawned by absolute `process.execPath` never passes through Nub, so its own exclude cannot suppress the pattern in `NODE_OPTIONS`.


colinhacks:

PR #799, opened 2026-08-27
runtime: forward caller arguments through the revokeObjectURL wrap

colinhacks:
`URL.revokeObjectURL()` with no arguments returned silently under nub; Node throws `ERR_MISSING_ARGS`. The blob-URL wrap called the saved native function with a fixed arity of one, so the zero-arg call passed `undefined` through. It now spreads the caller's arguments.

The `createObjectURL` wrap has the same fixed arity but no divergence: Node throws `ERR_INVALID_ARG_TYPE` for both zero args and explicit `undefined`, and ignores extras. Left as is.

New fixture plus integration test, verified RED before the fix. Upstream `test/parallel/test-url-revokeobjecturl.js` (v26.7.0) goes from exit 1 to exit 0.

### DIFF
--- a/crates/nub-cli/tests/integration.rs
+++ b/crates/nub-cli/tests/integration.rs
@@ -4419,6 +4419,27 @@ fn worker_blob_wrapping_preserves_file_instanceof_blob() {
 }
 
 #[test]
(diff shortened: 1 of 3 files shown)


colinhacks:

PR #800, opened 2026-08-27
transpile: report a file: URL as a stripped script's sourceURL

colinhacks:
Node reports `pathToFileURL(filename).href` as a stripped script's `//# sourceURL`, and the inspector surfaces it as `Debugger.scriptParsed`'s `url` — the key editors and DevTools match breakpoints against. nub emitted the bare path, so every transpiled file reached them under a different script identity than under node.

The percent-encoding matches `pathToFileURL` byte for byte. Its set is wider than the WHATWG path set, so it comes from a sweep against v26.7.0.

The cache schema goes to "7": the sourceURL is baked into the cached body.

Upstream `test-inspector-strip-types.js` now passes 4 of 5 scenarios, matching plain node on the same checkout.


colinhacks:

PR #801, opened 2026-08-27
build: isolate the target dir when runtime/ diverges

colinhacks:
A shared-bucket `nub` binary resolves `runtime/*.cjs` from the tree that compiled `nub-core` (its baked `CARGO_MANIFEST_DIR`), so a worktree editing only runtime files ran a sibling worktree's copy — three concurrent fix branches hit this, with red/green verdicts flipping as siblings rebuilt the shared bucket.

Fix: add `runtime` to the divergence and content-key pathspecs, so a worktree with any tracked or untracked `runtime/` change gets an isolated target dir. A key move costs the first worktree one cold build (a shared bucket is never seeded from foreign content — cargo's mtime-based path-dep freshness would accept foreign rlibs). `--print-target` is now a pure read, and the disk-reduction sweep's duplicated key formula is synced so its positive control passes.


colinhacks:

PR #802, opened 2026-08-27
node: route Node 23.0-23.4 to the compat tier

colinhacks:
Nub crashes at startup on Node 23.0 through 23.4 with `TypeError: module_.registerHooks is not a function`.

`module.registerHooks` is a semver-minor that shipped on the 23.x line at 23.5.0 and on the 22.x LTS line only later, at 22.15.0. The tier gate compared against 22.15.0 alone, so 23.0-23.4 sorted above the fast floor without having the API and was handed `--require preload.cjs`.

The fix excludes the band in `supports_augmentation`, the one predicate every downstream gate keys on. Two JS gates carried the same defect and now probe the capability instead of the version.

Verified on real 18.19.0, 22.14.0, 22.15.0, 23.0.0, 23.3.0, 23.4.0, 23.5.0, 23.6.0, 24.0.0, 24.19.0. Both new tests confirmed red without the fix.


colinhacks:

PR #803, opened 2026-08-27
runtime: restore the node: scheme on builtins that require it

colinhacks:
CommonJS `require("node:test")` — plus `node:sqlite`, `node:sea` and `node:test/reporters` — threw ERR_INVALID_RETURN_PROPERTY_VALUE under nub. Measured per release: broken on 22.15.0–22.17.1, 23.5.0–23.11.1 and 24.0.0–24.3.0; fixed in 22.18.0+, 24.4.0+ and 25+.

The defect is upstream and reproduces on plain Node with a pass-through hook: with a sync resolve hook registered, Node recomputed the URL through a helper that returned a bare `test` for an id only requirable with the scheme, so the load validator rejected the default step's null source. Nub hooks on every fast-tier run, so its users hit this unconditionally.

Nub now re-prefixes the bare id, a no-op where Node is already fixed.


colinhacks:

PR #805, opened 2026-08-28
Fix lat.md errors in anticipation of 0.13

1st1:


colinhacks:
Merged `main` in and resolved the conflict rather than asking for an amend.

`wiki/research/node-embedding-vs-spawn.md` was deleted from `main` in d50b1c2, so that hunk is dropped; the other two fixes apply unchanged. `native-resolver-prior-art.md` went in the same commit, so the review's third-file finding no longer applies. `lat.md` is still published at 0.12.2, so the pin stays where it is.

Verified the reflow through GitHub's own renderer: the packed form emits a literal paragraph and leaves every reference unlinked, the split form resolves. All 23 label mappings in `cold-start.md` are byte-identical after the split and all 23 are referenced.

colinhacks:
Shipped in v0.9.0: https://github.com/nubjs/nub/releases/tag/v0.9.0 — thanks for the contribution.

PR #807, opened 2026-08-28
repo: make the root nub-identity on nub.lock, and bootstrap CI with nub

colinhacks:
Two root lockfiles with no declaration made the repo ambiguous, so `nub install` refused nub's own repo. `package-lock.json` and `bun.lock` go; `nub.lock` replaces them. `tests/conformance/frontdoor/` still covers all seven incumbents.

All eight `npm ci` calls ran before nub is built. `ci.yml`, `busybox` and `compile-native` bootstrap a released nub via `./setup`; `wpt-worker` uses the nub it already built; `release.yml` builds a host nub, so a broken release cannot wedge the workflow that publishes its fix.

`--node-linker hoisted` is required: the build script panics on a symlinked store.

`ci.yml`'s root install no longer passes through the Socket Firewall shim.


colinhacks:

PR #808, opened 2026-08-28
build: cap the host at two compiling builds over six rustc tokens

colinhacks:
The rustc-count cap still let 10 rustc run 7 LLVM threads each at 1-3 GiB apiece. This adds a build-level layer: at most `NUB_BUILD_SLOTS` (default 2) cargo invocations compile at once machine-wide, FIFO, sharing `NUB_RUSTC_LIMIT` (default 6) rustc tokens. Slot = outermost cargo ancestor; reclaimed when the holder dies or idles `NUB_BUILD_IDLE` (120s), place kept on re-queue; rust-analyzer exempt; fail-open in place (wiped state dir, dead cargo, `make build-slots-off`); one compound command, so an older installer's copy cannot truncate a running wrapper; a build queued 20s says so once.

`qos-global` installs by rename; `build-status` shows holders, queue, stale installs.

Fake-cargo harness (15 scenarios, mutation-checked, ubuntu + macOS CI) and live on the host.


colinhacks:

PR #809, opened 2026-08-28
runtime: scan for .env.schema up to the workspace root

colinhacks:
`.env.schema` discovery checked exactly two directories — the project root and the workspace root. That covers a standalone project, a workspace root, and a member, and silently misses the one between them: a schema at a package that **contains** the package being run. `apps/web/.env.schema` was invisible from `apps/web/functions/`, because that directory has its own manifest and is therefore the project root, so the run fell through to nub's own `.env*` cascade with no diagnostic — the exact substitution `SchemaProblem` exists to refuse.

## What changed

`search_roots` now yields the project root and each parent up to and including the workspace root, nearest first, and `detect` walks it. The loader lookup already walked with that same ceiling for the same "nothing above the workspace is this project's" reason, so the two now share one walk and cannot drift apart.

Any `.env.schema` inside that ceiling counts, with or without a `package.json` beside it. Requiring a manifest was tried and dropped: the ceiling is what bounds the walk, and within it — usually one to three directories — a schema is a file somebody placed on purpose, so ignoring it would be a silent substitution. It also bought nothing mechanically, since `--path` needs only that the directory exists and explicitly ignores the `varlock.loadPath` it would otherwise read from a manifest.

Varlock does no ancestor walk of its own — `--path`, else `varlock.loadPath` from the `package.json` in the working directory, else that directory. Nub walks only because it resolves the project root while Varlock resolves the working directory, and `--path` is Varlock's own answer for that gap. So the walk supplies an input Varlock leaves to its caller rather than overriding an opinion Varlock holds.

The ceiling is bounded by `Path::starts_with` rather than an equality test against the ceiling, so a `workspace_root` that is not an ancestor of `project_root` yields the project root alone instead of walking to `/`. Component-wise comparison also makes it immune to a trailing slash, a `.` component, and the `/a/bb` vs `/a/b` prefix trap — probed rather than assumed.

`EnvOwner` carries the walked chain, and the two near-identical manifest readers collapse into one `declares`. This is fallout of the walk, not extra scope: it now stops at a schema rather than at a package boundary, so the schema's own directory need not carry a `package.json` — and both readers read exactly that manifest. Left alone, a manifest-less schema directory reports "declares nothing", which names the wrong fix in the diagnostic and, worse, misses a declared `dotenv-extended` and routes the run through `varlock run` against a schema written in another format.

The two readers take **different reaches**, and the difference is load-bearing. `rival_schema_tool_declared` reads only what the schema governs — the package being run, anything between it and the schema, and the schema's own directory. Directories above the schema answer to a different schema or to none, so their dependencies are not evidence about this file. `loader_declared` reads the whole chain, because a devDependency at a workspace root is how a monorepo declares the loader for every member, and that reader only chooses between two fatal messages rather than changing what runs.

Manifests are read through `strip_utf8_bom`, matching every other `package.json` reader in the tree. A chain scan makes skipping it worse: one BOM'd manifest anywhere would read as declaring nothing.

## Behaviour

| Schema location | Run from | Before | After |
| --- | --- | --- | --- |
| Standalone project root | that root | found | found |
| Plain subdirectory of a package | that subdirectory | ignored | ignored |
| No `package.json` anywhere above | that directory | ignored | ignored |
| Workspace root | a member | found | found |
| Directory with no `package.json` beside it | a member under it | ignored | **found** |
| A parent package's root | a package nested inside it | ignored | **found** |
| Above the workspace root | a member | ignored | ignored |

**The nested-package rows require a declared workspace.** The ceiling is `workspace_root.unwrap_or(project_root)`, and there is no workspace root without a `workspaces` field or an incumbent `pnpm-workspace.yaml`. In a plain repo the chain is a single entry, so `myapp/.env.schema` is still invisible from `myapp/functions/` — the Firebase and Vercel layout. That follows from the ceiling rationale (nothing above the workspace is this project's), and covering it would need a different stopping rule, which is a separate decision. The Varlock docs now state the precondition.

## Docs

The Monorepos callout in the Varlock page stated the old two-directory rule.

Three passages in the environment page were stale from #735 and wrong since it merged — worst of them a `console` block showing an error nub no longer emits (`` `--env-file` conflicts with .env.schema ``). Verified against the built binary with a control: `--env-file=.env.ci` beside a schema gives `FROM_CI=yes FROM_LOADER=<unset>`, and with no flag `FROM_CI=<unset> FROM_LOADER=yes`. The top callout ("nothing on this page applies") and the `--no-env-file` description are corrected too — both flags displace the hand-over.

## Verification

Six unit tests cover the walk directly: the nested-package gap, a schema with no manifest beside it still being read and still reading the chain for its diagnostic, the ceiling, a rival declared at the running package blocking an ancestor schema, and a rival hoisted ABOVE the schema not blocking it. One integration test drives the nested-package case through the real binary with a stub loader, asserting the hand-over happens, nub's cascade stands down, and the loader is pointed at the schema's directory.


colinhacks:

PR #810, opened 2026-08-28
loader: standalone tsx-style npm loader package

colinhacks:
Ships the runtime's resolve + transpile hooks as a standalone npm package: `node --import @nubjs/loader`, `--require @nubjs/loader`, `@nubjs/loader/esm`. Slim entries under `runtime/loader-*` arm only the shared `transform-core` / `preload-common` hooks; the addon rides 8 `@nubjs/loader-<platform>` packages resolved at runtime (`__NUB_ADDON_PATH`, probed last). Version/lockstep/release wiring, a docs page, and a tarball-install matrix under `tests/loader/` are included.

Name: `@nubjs/loader` (bare `nubjs` is refused by npm's similarity guard; scoped names bypass it). Before the first release that ships it, run `scripts/npm-first-publish-loader.sh` once — npm OIDC cannot create a package (npm/cli#8544), so the nine new packages must exist and be trusted first.

Verified from a clean tarball install: `--import` green on Node 18.19–26.7; `--require` on 20.19+/22.12+.


colinhacks:

PR #811, opened 2026-08-28
config: stop offering and writing engine settings nub does not consume

colinhacks:
`aubeNoAutoInstall` gates the engine's pre-run staleness check, reached only from the engine's `run`/`exec`/`restart`. None is an `ENGINE_VERB` — nub runs scripts through its own frontend and gates freshness in `verify_deps` — so it decided nothing here, yet `nub config set aubeNoAutoInstall true` wrote that key to the user's `.npmrc` and `config list --all` advertised it.

The embedder profile gains `unsupported_settings`. A listed setting is absent from `meta::find`/`meta::all`, so every generated accessor falls through to the default, and all three write routes refuse it, pointing at `verifyDeps` / `verify-deps-before-run`. Empty for standalone aube.


colinhacks:

PR #812, opened 2026-08-28
node: keep --experimental-webstorage off the inherited NODE_OPTIONS channel

colinhacks:
`--experimental-webstorage` does not exist before Node 22.4, which is above nub's 18.19 support floor, and `NODE_OPTIONS` is inherited by the entire process subtree. So any descendant on an older Node aborted at startup with exit 9 on a flag it could not parse.

That was reachable rather than theoretical: a host on the 22.4–24 band running Electron 34 or older (which embeds Node 20.18.1) hit exactly this. `strip_unsupported_node_options` never covered it, because that applies only to the *inherited* string, never to nub's own freshly pushed tokens. Issue #7 was the same flag reaching an older child through a nested `.nvmrc`, and it was closed by the node_modules pin guard rather than by taking the flag off this channel.

The flag now rides argv, from the site outside the augment block that the version-gated feature flags already use since #777. A child `node` still gets it: that invocation re-enters nub through the PATH shim and is given the flag on its own argv.

**The neutralize signal moves with it, and has to.** `runtime/polyfills.cjs` deliberately does *not* delete `__NUB_NEUTRALIZE_LOCALSTORAGE`, and that inheritance was what covered the subtree. Once the flag is per-process, a level can no longer be assumed to inherit a signal for a flag it received directly, so both are set at one site. A nearby comment claiming the preload deletes the var was already stale and is corrected.

### Verification

Measured against two independent controls — current `main` and released v0.7.4, both of which carry the flag in `NODE_OPTIONS`:

| Scenario | Result |
| --- | --- |
| direct run, hosts 22.15 and 24.17 | absent from `NODE_OPTIONS`, present on `execArgv`, `sessionStorage` live |
| 3-level grandchild chain | `localStorage` neutralized **and** `sessionStorage` live at all three levels |
| `nub run`, including a nested `sh -c 'node ...'` | identical to `main` |
| `--localstorage-file` opt-in | store round-trips, identical to `main` |
| `--no-experimental-webstorage` | disable respected |
| below-floor descendant (real Node 20.18.1), script path | `main` exits 9; this branch survives |

### Tests

`webstorage_flag_rides_argv_and_never_node_options` is the wiring test: the version-band and injection-policy unit tests all pass with the injection wired to the wrong channel, because none of them observe which channel a real spawn used. Confirmed to fail for the right reason by running it against a pre-change binary, where it panics on the `NODE_OPTIONS` assertion specifically.

`localstorage_neutralization_reaches_grandchildren` could not fail before. It asserted only that `localStorage` is `undefined` at each level, which is equally true of a process that never received the flag at all — so it passed whether or not the flag still reached the subtree, which is exactly how this change could have broken it silently. It now asserts `sessionStorage` is `object` alongside, which is true only when the flag really arrived.

### Not changed

`--test-coverage-exclude` (floor 22.5) stays on `NODE_OPTIONS`, so a host on 22.5+ still cannot run Electron 34 or older. It is the one token that must share a channel with the preload, since a coverage grandchild nub never spawns inherits the preload through that string alone. Keeping it was decided explicitly and is recorded in `spawn.rs` with the declined alternatives, so it is not silently "fixed" later.

The docs commit also corrects `wiki/design/architecture.md`, which still described the pre-#777 mechanism where version-gated flags travelled in `NODE_OPTIONS`.

Refs #7, #246


colinhacks:

PR #813, opened 2026-08-28
outdated: cover the undatable-registry warning end to end

colinhacks:
A registry serving no per-version publish times makes the release-age gate fail closed: `install`/`update` hard-error, while `outdated` warns on stderr and exits 0. That predicate sat behind a registry fetch with no test reaching it — the one thread left open on #779.

Three cases over an in-process registry (std-only, offline, no new dependency): the warning fires and stays off stdout; a positive control with every version dated proves it is not unconditional; and a stale `latest` tag alone does not warn, since only the manifest range predicts plain `update`.

Test-only — no production code changes.

Refs #722


colinhacks:

PR #814, opened 2026-08-29
lockfile: drop non-optional peerDependenciesMeta entries from the pnpm writer

colinhacks:
pnpm 12's lockfile reader requires `optional` on every `peerDependenciesMeta` entry. Nub wrote an empty mapping for a peer declared `optional: false` (`vitest` does this for `vite`), so `pnpm@12 install --frozen-lockfile` failed with `ERR_PNPM_BROKEN_LOCKFILE`; pnpm 10 accepted the file. pnpm drops those entries and records only `optional: true`; the writer now does the same. Writer test added in `aube-lockfile`.

Verified on a `vitest@4.1.11` fixture: this build's lockfile passes `pnpm@12.0.0 install --frozen-lockfile --offline`; the shipped 0.7.4 binary's fails.

### DIFF
--- a/vendor/aube/crates/aube-lockfile/src/pnpm/tests.rs
+++ b/vendor/aube/crates/aube-lockfile/src/pnpm/tests.rs
@@ -5270,3 +5270,79 @@ snapshots:
         Some("is-odd")
     );
 }
+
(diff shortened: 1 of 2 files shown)


colinhacks:

PR #815, opened 2026-08-29
node: restate the default coverage exclusion on argv only; remeasure the 26.7.0 corpus on merged main

colinhacks:
Re-measures node + nub on the Node 26.7.0 corpus against a release build of current main. Regressions vs node: 164 → 146; 0 untriaged config entries.

- `sequential/` runs one file at a time after the parallel pool, as Node's runner orders it; the three `test-net-*` failures were pool artifacts.
- The #798 default coverage pattern stays on argv but leaves `NODE_OPTIONS`: Node excludes on any glob match, so a grandchild's own `--test-coverage-exclude=!test/**` could not undo it (six coverage snapshots reported an empty table). Regression test added.
- Triage classes from the divergence review added to the regen script.


colinhacks:

Issue #270, opened 2026-07-01
nubx @modelcontextprotocol/inspector fails with ERR_NUB_TRUST_DOWNGRADE, while npx @modelcontextprotocol/inspector works


omenlog:
When launching the MCP Inspector through nubx, install resolution fails with `ERR_NUB_TRUST_DOWNGRADE`. The same package works when invoked through **npx**.

This is the error

```
nub 0.2.9 — resolving  [░░░░░░░░░░░░░░░░░░░░]    35/293  pkgs · ETA …
ERR_NUB_TRUST_DOWNGRADE

  × dlx install failed
  ├─▶ failed to resolve dependencies
  ╰─▶ trust downgrade for tailwind-merge@2.6.1 (trustPolicy=no-downgrade): earlier published version 3.4.0 had trusted publisher but this version has no trust evidence
  help: a trust downgrade may indicate a supply-chain incident — the publisher's previous releases carried trusted publisher but tailwind-merge@2.6.1 does not.
        to bypass: pin a version that retains evidence, set `trustPolicy = off` in .npmrc / pnpm-workspace.yaml, or add `tailwind-merge` (or `tailwind-merge@2.6.1`) to `trustPolicyExclude`.
```

omenlog:
<img width="1955" height="300" alt="Image" src="https://github.com/user-attachments/assets/b6c5d5c0-c89e-41b2-9ce9-e89ab479aa6d" />

colinhacks:
Fixed in #277 — nub now defaults `trustPolicyIgnoreAfter` to a 14-day window, so the no-downgrade gate no longer flags a freshly published version as a downgrade.

colinhacks:
Shipped in v0.3.0: https://github.com/nubjs/nub/releases/tag/v0.3.0

PR #819, opened 2026-08-30
resolver: backtrack past a trust-refused version instead of failing the install

colinhacks:
`trustPolicy=no-downgrade` refused the version a range resolved to and aborted the whole install, even when an older version satisfying the same range kept its trust evidence. The age gate already behaves the other way — `pick_version` scans the satisfying set and keeps the newest release clearing the cutoff — so the two gates deadlock on any package that publishes one version manually between two attested ones.

`fast-glob@^3.3.2` is that case today, and shipped nub cannot install it:

```console
$ nub install                       # fast-glob -> @nodelib/fs.walk -> fastq
ERR_NUB_TRUST_DOWNGRADE

  ╰─▶ trust downgrade for fastq@1.20.2 (trustPolicy=no-downgrade): earlier
      published version 1.20.0 had trusted publisher but this version has no
      trust evidence
```

`fastq@1.20.3` is too new for the release-age floor, `1.20.2` was hand-published with no attestation, and `1.20.1` — 249 days old, trusted publisher, SLSA provenance — satisfies the range and was never considered.

## What changed

On a refusal the resolver walks down the versions the range admits and takes the newest that clears both gates, naming what it skipped:

```console
$ nub install
WARN skipped fastq@1.20.2 (trustPolicy=no-downgrade): earlier published version 1.20.0 had trusted publisher but this version has no trust evidence; resolved to fastq@1.20.1 instead code=WARN_NUB_TRUST_DOWNGRADE_SKIPPED
```

- **Bounded at the refused pick**, in the pick's own direction — downward for a highest-wins pick, upward under `resolution-mode=time-based`. A re-pick can never climb above a `dist-tags.latest` the publisher has moved off.
- **A literal `latest` range widens to `<=` the tag**, the same widening `pick_version` applies to an age-gated `latest`, so `dlx` and an unversioned `add` get the fix too. Other tags name exactly one version and stay un-widened.
- **Known-vulnerable candidates rank second**, matching `prefer_non_vulnerable_pick` — a worse answer than a clean version, a better answer than failing.
- **A missing `time` entry stays fatal.** That is a metadata anomaly, not a refused candidate, and every other candidate would be judged on the same shape.
- **The lockfile-validation pass is unchanged.** A lockfile pins an exact version, so there is no range to walk; `nub ci` and a lockfile carrying a refused version still abort rather than silently install something the lockfile does not name.

Neither gate is weakened. When nothing the range admits clears both, the original refusal stands, and a version the attacker published is never installed either way — the change is that the user is told rather than blocked.

This is a deliberate divergence from pnpm, whose `failIfTrustDowngraded` throws. pnpm implements its age gate as a packument filter (`filterPkgMetadataByPublishDate`) and its trust gate as a post-pick throw, so the same asymmetry is baked into upstream.

## Verification

- Every new test was driven red by removing the guard it covers — the backtrack itself, the range bound, the direction bound, the age filter, and the `latest` widening each fail their own test when reverted.
- End to end against a build of this branch and a build of the merge base: `fast-glob@^3.3.2` goes 23 → 0 and locks `fastq@1.20.1`. `fastq@1.20.2` pinned exactly still exits 23; `~1.20.2` and `>=1.20.2` still exit 23; `fastq@1.20.3` still exits 21 on the age gate.
- Ten real packages (`express`, `typescript`, `vite`, `eslint`, `chokidar`, `webpack`, `@babel/core`, `axios`, `rollup`, `prettier`) produce byte-identical lockfiles on both builds.
- `cargo clippy --all-targets --all-features`, `cargo fmt`, `cargo test -p nub-cli`, and `cargo test --workspace` in `vendor/aube` are green.

Refs #270

### DIFF
--- a/site/content/docs/install/index.mdx
+++ b/site/content/docs/install/index.mdx
@@ -651,7 +651,7 @@ Frozen reinstalls — `nub ci`, `--frozen-lockfile`, a teammate's clone — inhe
 
 Nub also weighs trust *evidence* across a package's release history — OIDC provenance, a trusted publisher, a staged-publish approval. A resolved version that carries weaker evidence than an earlier-published version of the same package is refused, because a maintainer's pipeline that suddenly publishes without the attestation it used to carry is the shape of a token-theft supply-chain attack.
 
-Refusing a version is not the same as failing the install. Nub keeps walking down the versions the range admits and takes the newest one that still carries its evidence, naming what it skipped and why:
+Refusing a version is not the same as failing the install. Nub keeps looking through the versions the range admits and takes the best one that still carries its evidence, naming what it skipped and why:
 
 ```console
 # captured: fast-glob@^3.3.2 → @nodelib/fs.walk → fastq, whose 1.20.2 was hand-published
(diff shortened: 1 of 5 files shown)


colinhacks:
Both docs points and the dedupe nit are fixed in c4e54784bf.

The directional claim was wrong — `pick_lowest` is true for a direct dep under `resolution-mode=time-based`, so the backtrack walks up there, which this PR's own test asserts. The docs now state the invariant that holds in both modes: the substitution never crosses the refused version. The lockfile carve-out is documented in the same section, including why a local `nub install` can succeed while `nub ci` on an older lockfile still fails.

The notice now dedupes on `name@refused -> substitute` rather than the refusal alone, and the doc comment matches the key.

colinhacks:
Fixed in 8635f57262. Measured all four modes against a lockfile pinning `fastq@1.20.2`: `nub install`, `--frozen-lockfile` and `nub ci` all exit 23; `nub install --no-frozen-lockfile` exits 0 and re-resolves to 1.20.1. The paragraph now names all three aborting forms and the remedy.

colinhacks:

PR #822, opened 2026-08-31
build: self-pruning GC for build residue (buckets + worktree targets)

colinhacks:
`scripts/target-gc.sh`, run at most hourly from `rust-build.sh`: collects buckets no live worktree resolves to after 24h (orphanhood via each worktree's own `--print-target`), anything idle 14 days, and — under 100 GiB free — idle state oldest-first, sparing the newest at-risk dir. Never touches the calling build's dir, anything written within 2h, the newest rlib seed, or `nub-dev`'s bucket; deletes by rename-then-remove.

`make target-gc` prints a dry-run audit. 10-scenario harness (ubuntu + macOS), each protection mutation-checked red; the live-host dry run matched the manual audit's orphaned/referenced verdicts exactly.


colinhacks:
Landed directly on main as a247a228f9 (tooling).

PR #825, opened 2026-09-01
ci: probe the Windows node-24 worker test on current main (not for merge)

colinhacks:
Empty commit on main's tip to get a stable full-matrix run: main's run 33447462117 failed `worker_without_inbound_listener_exits_naturally` (worker-hung) on windows-latest/node 24, and every rerun since was cancelled by the next push to main. #812's own head passed the same leg. Green here = flake; red = regression in today's landings.


colinhacks:
Flake confirmed: main run 33453178513 (0111d6ea4c) passed the same leg; this probe's matrix also ran green on it. Not needed.

PR #826, opened 2026-09-01
docs: compile page + v0.9 announcement post

colinhacks:
Docs page for `nub compile`, plus the announcement blog post.

**Do not merge before the v0.9 release.** `nub compile` ships in release binaries but is undocumented on the site; publishing either file announces it. This is the second of the two PRs — [#830](https://github.com/nubjs/nub/pull/830) carries the code changes and merges first.

The post walks the mechanics in sequence — per-OS payload injection, the Node SEA history, assets, native addons, self-extraction with the Deno/Bun prior art, and the jsdom per-package bundling numbers — then charts size (raw and gzip) and warm startup, and closes on `--smol` with `--install-message`. The startup chart shows Bun and Deno ahead and says so.

## Before merging

1. Move the post's `date:` from its placeholder to the actual release date.
2. **Re-take the three benchmark rows on an idle Mac**, and bump both chart captions from `nub 0.8.2` to the shipping version. Compile `console.log("hello, nub")` with `nub compile`, `nub compile --smol`, `bun build --compile` and `deno compile`, run each once to warm the extraction, then `hyperfine --warmup 20 --min-runs 150` over the four binaries plus `node hello.ts`, and record `ls -l` and `gzip -c | wc -c` for the size rows. Check `uptime` first — load above ~3 makes the timings unpublishable.
   - Warm start has never been re-measured: on a contended box Deno and Node reproduced within 2% while Bun came out 2.4× slow, which is indistinguishable from a Bun regression.
   - The on-the-wire sentence under the first chart quotes 29.0 MB gzipped against Bun's 23.9 MB; re-take both with `gzip -9` at the shipping version.
   - The two size rows drift ~2%: `--smol` measures 0.92 MB / 493 KB gzipped today against the published 0.94 MB / 504 KB, which also moves the `31×/67×/80×` labels to `32×/69×/81×`.

## Verified against a build of main (2026-09-04)

Every other number and transcript was re-measured and either confirmed or corrected. Confirmed unchanged: 29.5 MB embedded (at the pinned `engines.node` 26.8.1), 29.6 MB addon build, 2.8 MB jsdom with all its file counts (1,834 → 1,006, jsdom keeps 657, parse5 33→2, whatwg-url 19→3, mdn-data 25→4), 63.4 MB Bun, 75.1 MB Deno, byte-identical repeat builds, one shared extracted Node across binaries, Bun's `--windows-icon` refusal, and every Node SEA claim (19.7.0, Stability 1.1, `--build-sea` in v25.5.0, the two issue dates, the repo's last push).

Corrected: the app-size claim measured Mach-O padding rather than the app (a four-dependency CLI and a one-line program produce byte-identical artifacts on macOS, so the subtraction yielded zero — the real figures are 445,296 B bundled and 105 KB compressed); stock Node is 145 MB, not 139; six console blocks printed a summary format nub no longer produces; the extracted-tree listing named a file nub no longer emits and omitted two it always writes; sharp declares 25 per-platform packages, not 24.


colinhacks:

PR #827, opened 2026-09-01
compile: find the strip tool by its Windows spelling

colinhacks:
The strip probe looked for a tool with `dir.join(name).is_file()`. On Windows the strippers ship as `llvm-strip.exe`, which that never matches, so both candidate lists missed and the compiler took its unstripped early return on every compile run on a Windows host. It fails quietly: a `note:` on stderr, and an artifact roughly 4 MB larger than it should be.

The probe is now PATHEXT-aware, mirroring the helper the launcher already carries for the same reason. It still returns the bare name, since `Command::new` does its own PATH search and only the existence check was wrong.

Found during a production-readiness review of the compile feature, not from a report.

Gates are green on macOS and behavior there is unchanged. The added test can only go red on Windows, so the Windows leg is what actually exercises this.

### DIFF
--- a/crates/nub-cli/src/compile/mod.rs
+++ b/crates/nub-cli/src/compile/mod.rs
@@ -1899,28 +1899,17 @@ impl Drop for FileGuard {
     }
 }
 
-/// Returns the file that was actually found, not the name that was searched for.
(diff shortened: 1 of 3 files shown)


colinhacks:

PR #829, opened 2026-09-01
compile: close the shim-capability gap at both ends, and gate it on every host

colinhacks:
A production-readiness pass over `nub compile`. One real defect class, closed at both ends, plus the test and CI coverage that was missing around it.

## The defect

A `--smol` build that carries `--external` or `--allow-dynamic-import` installs a `module.registerHooks` shim, so the Node it runs on must provide that API. The build-time gate only saw the pin's **floor**, and a floor in `22.15.0..23.0.0` admits the 23.0–23.4 band — which sorts above the floor and satisfies the range, yet predates `registerHooks` on the 23.x line. `--target ">=22.15"` built clean and the artifact died on the user's machine with `registerHooks is not a function`.

It is closed at both ends, because a version bound cannot answer a capability question:

- **Discovery** — the manifest now records that the payload *needs* the API, and `SmolTarget::matches` applies `supports_augmentation` to the candidate actually found, rather than predicting it from a bound.
- **Provisioning** — the same artifact could still bake a provisioning preference with the same defect: `--target ">=22.15 <23.5"` resolves its newest satisfying release to 23.4. `provision_preference_is_usable` drops it at build time. Falling back is safe rather than merely conservative — `check_node_support` has already failed the build when a shim-bearing floor lacks the API, so no preference means the launcher provisions a floor known to run the payload.

The build gate stays, because it fails the *build*, which is where a wrongly-chosen target should be reported and before the ~100 MB download.

## Also here

- A discovered Node refused on capability now says so once, naming the version that gained the API on its own line. It was silent, so a user with a modern-looking Node saw an unexplained download.
- **The erasability harness runs on every native host**, not Linux alone — macOS and both Windows legs. It is the only gate that catches an augmentation the runtime makes and the compiler does not, and that class is not Linux-specific. One portability fix was needed: `--out` is taken verbatim, so it now names the target's exe suffix instead of asking for an extensionless `./bin`.
- `compile` now appears in both help surfaces, feature-gated so a feature-off dev build still advertises nothing it would reject.
- Three bundler tests for flags that had no coverage (`--conditions`, `--sourcemap-exclude-sources`, `--alias`), a brand leak fixed in a user-facing bundler error, and two misleading test names corrected — both cover the harness, not a compiled artifact.
- `--enable-source-maps` withholding on Node 26: the comment asked for a re-verification once 26.8.0 shipped. It has, so I ran it. `TypeError` on 26.7.0, `AssertionError` on 26.8.1 — the predicted boundary holds and the constant is unchanged. The artifact published as v26.8.0 also gives `AssertionError`, but it self-reports `v26.8.0-alpha.0.0.0`, which sorts *below* `26.8.0` and so lands inside the withheld band; the comment now states that rather than rounding it off.
- `package-lock.json` is ignored. `npm install` creates one whenever anyone populates `node_modules` for the compile suites, and being neither tracked nor ignored left it one `git add` from a commit.

## Verification

Every new test was mutation-tested: the fix reverted, the test confirmed **red**, the fix restored. That covers the producing side (`provision_preference_is_usable`), the reading side (the `smol_target` wiring), and the manifest field end-to-end from a produced artifact in both directions.

Local gates green: root `clippy --all-targets --all-features`, `fmt --check` in all three workspaces, launcher `clippy --all-targets` plus 74/74, nub-cli 785 unit tests. The augmentation harness runs **13 ok / 0 failed** on macOS at Node 24, the version this workflow pins.

**Reviewing this needs one flag.** A plain local `nub compile` silently injects a *cached release* launcher rather than your build (`compile/launcher.rs`), so a fix can look absent when it is present. Pass `__NUB_LAUNCHER_TEMPLATE=crates/nub-launcher/target/release/nub-launcher`.

## Known, not changed — two things this pass found and deliberately left

`tests/compile-augmentation/a-sourcemap` fails on Node 26.0–26.7, and that is correct: `source_maps_safe` deliberately withholds `--enable-source-maps` across that band for nodejs/node#63169, and a compiled artifact is a minified bundle so it needs the map that `nub <file>` does not. CI pins Node 24, where the row is green. Teaching the fixture about the withholding band would duplicate that policy, so what the row should assert is left as a call to make.

`compile_artifact_process_bootstrap_topology` fails on a Node 26.4+ host over `--js-defer-import-eval`, which #770 auto-unflags from that version. Its CI gate (`ci.yml`, "Compile artifact process topology") is pinned to `matrix.node == '24'`, so nub's newest supported major is unverified for compiled-artifact process topology and CI never sees the mismatch. Raising the gate to 26 turns the leg red until the Worker-vs-root execArgv difference is resolved, which is flag-injection territory rather than compile-capability, so it is reported rather than folded in here.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -16,3 +16,13 @@ tests/fixtures/import-text-named/*  text eol=lf
 # provisioning ever widens.
 tests/fixtures/data-loaders-require/*       text eol=lf
 tests/fixtures/loader-redirect-both-ways/*  text eol=lf
+
+# Same trap, third form: the compile-augmentation harness reads `*.flags` as
+# ARGV and `*.differs` as a reason string, so a CRLF checkout appends `\r` to a
+# flag's value. On Windows that turned `--sourcemap=inline` into
+# `error: invalid value 'inline' for '--sourcemap[=<MODE>]'` — a message naming a
+# value that IS in the allowed list, because the `\r` making it invalid does not
+# print. It surfaced the moment the erasability gate started running on every
+# native host rather than Linux alone. `**` rather than `*` because the `.d`
+# companion directories nest, and their sources feed the same byte comparisons.
+tests/compile-augmentation/fixtures/**  text eol=lf
--- a/.github/workflows/compile-native.yml
+++ b/.github/workflows/compile-native.yml
@@ -174,32 +174,44 @@ jobs:
       # reference run and the artifact are all the same build. That control is what
       # makes a green row mean something, and the harness prints how many rows were
       # discriminating so a run that quietly stops proving anything is visible.
+      #
+      # Every host, not just Linux. The class this catches — an augmentation the
+      # runtime makes and the compiler does not — is not Linux-specific, so gating
+      # it on one OS left the other three unguarded against exactly the failure
+      # the step exists to find. The harness needed one portability fix to get
+      # here: `--out` is taken verbatim, so it now names the target's exe suffix
+      # rather than asking for an extensionless `./bin` on Windows. The other two
+      # Unix-looking things in it are already portable — the provisioned-Node
+      # lookup falls back to `command -v node`, and XDG_CACHE_HOME outranks the
+      # platform cache roots on every OS (`nub-launcher/src/cache.rs`).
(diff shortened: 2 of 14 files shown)


colinhacks:

PR #830, opened 2026-09-01
compile: pre-v0.9 fixes and flags, plus the Windows script-env casing fix

colinhacks:
Pre-v0.9 fixes for `nub compile` and the runtime it ships. Docs and the announcement post are [#826](https://github.com/nubjs/nub/pull/826), which must not merge until the release.

**Worker `execArgv`.** `nub` hides `--js-defer-import-eval` from `process.execArgv` so tooling forwarding that array does not hand Node back a flag it refuses. The existing guard covered a Worker given an explicit `execArgv`; workers that inherit the real exec argv were unguarded, and forwarding the reappeared flag threw `ERR_WORKER_INVALID_EXEC_ARGV` — the Turbopack break, one level down, in the worker pools most likely to forward `execArgv` at all. The signal now rides `worker_threads` environment data, which survives an explicit `env`, is transitive to nested workers, and does not cross a process boundary. `env: {}` is out of reach: it drops NODE_OPTIONS, so the preload never runs in that thread.

**Windows static CRT.** `nub.exe --version` exited `0xC0000135` on a clean Windows Server 2022 with empty stderr — the loader kills the process before any code can report it. Both it and the compiled-artifact launcher imported `VCRUNTIME140.dll`. The flag is set in the build steps rather than a `.cargo/config.toml`, because cargo concatenates rustflags down the tree and `+crt-static` is sticky, so a root config would reach the N-API addon with no way to opt out — and a static CRT in a cdylib loaded into `node.exe` is heap corruption, not a link error. `nub-native` carries a `compile_error!` guard against exactly that.

**`gypfile: false`.** `nub add better-sqlite3` exited 1 on any machine without Python while `npm install` succeeded in seconds, because the implicit `node-gyp rebuild` fallback was missing npm's `gypfile !== false` clause. Not Windows-only; Windows just guarantees it bites.

**Rust gzip backend.** The static CRT above broke the Windows link: `nub-launcher` failed with `LNK2019: unresolved external symbol __imp__wassert` from `libz-ng-sys`. rustc had the flag (`/defaultlib:libcmt` is on the link line) and every other C dependency — zstd, blake3, lzma, ring — linked clean, which isolates it to one crate: `libz-ng-sys` builds zlib-ng through CMake, which forces the dynamic MSVC CRT and never reads `CARGO_CFG_TARGET_FEATURE`. cc-rs handles that variable; cmake-rs only forwards it when no generator is selected, which is not the case on a Windows runner. `aws-lc-sys` is the control — also CMake, unaffected, because it sets `CMAKE_MSVC_RUNTIME_LIBRARY` itself. So `flate2` moves to `zlib-rs`, the same algorithms in Rust; the launcher workspace now has no CMake dependency at all. `vendor/aube` is included because the intent was already written there and not implemented — its workspace entry keeps library crates pure-Rust "so embedders don't inherit a cmake build dependency", while the binary crate re-enabled `zlib-ng`, and feature unification is not scoped to a binary. Measured, since the C backend was chosen for speed: interleaved A/B over six rounds gives zlib-ng 451 MB/s and zlib-rs 467 MB/s on a 3.7 MB tarball with byte-identical output — within noise. Verified with a cold-cache install extracting 622 files. `crates/nub-phantom` is the repo's fifth tracked lockfile and its only CI step omitted `--locked`, so the swap left it stale with nothing to catch it; that flag is added here.

**`--define` truncation guard.** `--define API=https://example.com` built clean and shipped a binary that died with `ReferenceError: https is not defined`. It is valid JavaScript — an identifier plus a `//` comment — so nothing upstream caught it. The root cause is that oxc's define validation parses the value and discards the result without checking the parse reached the end of input, so anything stopping early is silently truncated. The guard now compares the parsed expression's span against the value instead of pattern-matching a URL, which closes `/* c */https://x` (a leading comment defeats `trim_start`). Truncation alone is not enough: `git+ssh://host/repo` parses entirely, as `git + ssh` plus a comment, and still fails on `git` — so the comment-swallowed scheme is checked too. `2://x` and `http:/single` change from accepted to rejected; both silently substituted something the author did not write, and neither raises `ReferenceError`, so the message branches rather than promising an error that never arrives.

**`--metadata`, the Windows version resource.** A compiled binary carried no version resource, so Explorer's Details tab was blank and `(Get-Item app.exe).VersionInfo` read nothing — the state installers and antivirus heuristics treat as anonymous. Product name, description, version and company now default from the nearest `package.json` and the original filename from `--out`, so the common case needs no flag; `--metadata Key=value` overrides one and `Key=` drops one, the spelling an empty `--install-message` already uses. Refused rather than ignored for a non-Windows target, matching `--icon`, since no other container format carries the fields. The encoder is written here rather than shelled out to `rc.exe`, which would make the field Windows-hosted-only: Bun refuses all five of these off a Windows host. Its three rules fail silently rather than erroring — every node body is 32-bit aligned, `wLength` spans a node including its own padding, and `wValueLength` counts words for a `String` node and bytes everywhere else — so a resource that is wrong is still walkable and Windows reads garbage. The hazard that cost the most is in the resource directory rather than the encoder: libsui emits root-table entries in insertion order while the format requires ascending ids, because `FindResource` binary-searches them, so an `RT_VERSION` (16) inserted before the `RT_GROUP_ICON` (14) that `build` writes last loses **both** resources with no error. The insertion is deferred to `build` for that reason, and the test pairs an icon with the metadata because without one those ids ascend on their own and the hazard cannot fire.

Verified by re-reading the finished file rather than the writer's output. A cross-compiled Windows binary cannot be executed on the build host, so `verify_artifact` — which already re-read the payload and printed `payload verified statically` — now also walks the version resource by the path `GetFileVersionInfo` takes and compares parsed values, which catches a root header whose children were truncated where a byte compare would only catch a change. End to end, a 30 MB cross-compiled artifact built with both `--icon` and `--metadata` reads back a resource root of `3(RT_ICON) 10(RT_RCDATA) 14(RT_GROUP_ICON) 16(RT_VERSION)`, ascending, with the metadata intact under a parser sharing no code with nub.

**`--metafile` and `--drop`.** `--metafile report.json` writes the build report in esbuild's schema, so it opens in `analyzeMetafile`, esbuild-visualizer and bundle-buddy. Per-input bytes come from Rolldown's own `rendered_length`; a Rollup virtual id (`\0name`) is rendered as an esbuild-style `virtual:` path, because a NUL reaching the JSON breaks every consumer that treats an input key as a filename. `--drop console` and `--drop debugger` remove those calls at build time, as a closed set rather than a free string so an unrecognised name cannot be a build that silently drops nothing; combining either with `--no-minify` is refused, since dropping runs inside the minifier. One trap on the way: an unset compress flag falls back to `CompressOptions::smallest()`, which already has `drop_debugger` on, so writing `Some(false)` for the flag the user left out would make `--drop console` start emitting `debugger` statements that plain minification removes.

**Icon decode errors.** `load_icon` proves the container is an ICO before anything is built, so a failure at `set_icon` is one of the images inside it — and the underlying error named a variant rather than an action (`ImageError(Decoding(DecodingError { format: Exact(Ico), underlying: Some(PngNotRgba) }))`). The case that reaches users is a PNG-compressed entry that is RGB, which several converters emit; the message now says so.

**Windows package-script environment casing.** Every `nub run` script body on Windows saw `NPM_PACKAGE_NAME` where npm, pnpm, yarn and bun all deliver `npm_package_name`, so `$npm_package_name` expanded to nothing and `Object.keys(process.env)` returned zero lowercase `npm_` names — a lookup still answered, because Windows' `process.env` is case-insensitive, which is why this went unnoticed. Measured against all four package managers on one fixture: nub was alone in doing it. The cause is busybox-w32's shell, which up-cases every name as it loads the Windows environment; upstream considers that correct and declined the preserve-casing patch (rmyorston/busybox-w32#125), so the restoration belongs here. busybox's own lookup is case-sensitive, so an `export npm_package_name="$NPM_PACKAGE_NAME";` prologue in front of the script body fixes expansion, the casing children inherit, and enumeration at once — for a consumer in any language, not only the Node children nub augments. It re-binds names rather than values, so nothing needs quoting and a value carrying quotes or newlines cannot break the body. The list comes from the command's own environment rather than a hand-kept copy, and skips names that are already uppercase, names a shell cannot export, and names being removed — re-exporting that last class would resurrect a variable the caller had cleared. Only the bundled default shell gets the prologue: a `--script-shell` is the user's own and `cmd` is case-insensitive. Verified on a real windows-latest runner through the existing busybox probe, which gained the two symptoms as cases plus an instrument check that drives the staged shell directly and fails if it ever stops up-casing — without which both cases would pass whether or not the prologue did anything. 31/31, so the 26 pre-existing cases did not regress.

**The metafile import graph.** The first report described every static edge by the importer's module format, which is right for a module that is wholly one thing and wrong for one that is not: an ESM file containing an unbound `require()` has an ESM format and a require edge at the same time. The kind is only ever offered in the resolver hook, so an observer plugin reads it there and replays resolution through `ctx.resolve` with that edge's own kind, making the answer the build's rather than a second opinion. Three corrections followed from getting the position wrong. The observer has to run FIRST, ahead of every resolver, because a plugin that claims an edge returns before the ones behind it are asked — which is how an `--external` `require()` came to be reported as an import statement; that is the opposite of where the module collector belongs, which is last, so the source length it records is what the bundler parsed, and the two are now separate plugins. `imported_ids` is deduplicated by target, so a module that both imports and requires one file reported a single edge where esbuild emits two; the observations supply the multiplicity and `imported_ids` stays the completeness floor. And an external edge was reported under nub's synthetic id, which the struct's own comment calls an internal compiler identity — esbuild reports the specifier the runtime resolves, so the report translates back.

One divergence is left, deliberately, and the note in the source says why so nobody re-derives it: two edges sharing an importer, a target and a kind collapse to one entry where esbuild emits two. Rolldown memoizes resolution on (importer, specifier, kind), measured as one hook call for two identical imports against two for an import beside a require, so the second occurrence never reaches a plugin; `ModuleInfo` exposes only deduplicated ids, `NormalModule`'s import records are empty at `module_parsed` because linking fills them, and that is the only hook handed a `NormalModule`. Closing it means patching Rolldown or re-parsing every module to recount edges the bundler already counted. Targets, kinds and byte attribution are unaffected. A test asserts the current answer on purpose, so a Rolldown that exposes per-edge records fails there and points at the note.

**A resolver bug the report exposed.** `CompilePreamble` re-enters the resolver so its private graph resolves from the runtime directory rather than the application's cwd, and passed no options — which defaults the import kind to `Import`. Every `require()` in that graph was therefore re-resolved as an import, taking the `import` branch of any conditional `exports` map it met, and fabricating an import edge for the report alongside the real one. Found by sweeping a built binary rather than by a failing test; the user-authored half of the graph was correct throughout. Measured to separate the two halves: with the fix the resolved input set is byte-identical, so no package resolves differently today and the `exports` hazard is latent, while the edge count drops 18 to 12 — exactly the six fabricated edges. Every `require()` in that graph is currently a builtin, a relative path or a spliced virtual module, which is why nothing miscompiled.

**`--metadata` end to end on Windows CI.** The encoder round-trips in unit tests and `verify_artifact` refuses a compile whose resource is unreachable, but nothing ran the CLI: the flag, the `package.json` defaults and the PE libsui writes are three layers that only meet in a real build. `tests/compile-metadata/` covers the manifest defaults, an override, a drop, and the negative control that an empty manifest earns no resource at all — without which the other arms prove nothing, since a Windows binary can carry a version resource for reasons of its own. It reads the result back with a parser sharing no code with nub, runs on both Windows legs of `compile-native.yml`, and uses `--smol` so no Node is downloaded. Both controls were exercised before committing: breaking one expected value fails the run with a self-debugging message.

**The Windows runtime-cache flake, and why it was this PR's to fix.** `Windows embedded runtime and compile` failed on two of eight runs here against six of six elsewhere, and the confound was this branch's own static-CRT flag, so the difference could not be read off the history. A probe settled it instead: a verbatim copy of `windows_security.rs` in a standalone crate, walking the real predicate, built in both linkages on the same runner across a twelve-shard matrix. Both linkages behave identically, and roughly one `windows-latest` image in six gives `D:\a\_temp` an effective `Authenticated Users: Modify` ACE. `Modify` contains `DELETE`, which is in the runtime cache's non-leaf dangerous mask, so nub correctly refuses the ancestor and relocates — leaving no `runtime-*` directory where the assertions look for one. Only an ancestor nub creates is repairable, and that one belongs to the runner. So nub's behavior is unchanged: the step's cold cache moves to `$env:USERPROFILE`, which the same probe measured usable on 24 of 24 shard-linkage pairs including the bad one, against a `RUNNER_TEMP` control that was false on two.

The one product change is a diagnostic. The refusal named the path the caller asked for, which cannot distinguish a runner-created ancestor from the leaf nub makes and hardens itself — and that distinction is the whole question of whether there is anything to fix. It now names the failing component and which of the two it is.

**One message.** `Stripped + ad-hoc re-signed the embedded Node` reads as alarming for a routine build step, and is now `Signing embedded Node.js`.


colinhacks:

Issue #745, opened 2026-08-16
global-bin-dir is ignored in the user ~/.npmrc — the exact file `nub config set global-bin-dir` writes to


s-a-c:
### What happens

`global-bin-dir` is honored from a **project** `.npmrc` and from the environment, but **not from the user `~/.npmrc`** — which is precisely where `nub config set global-bin-dir` writes the key. The config writer and the setting reader disagree about the source of truth, so the tool's own documented way of setting the knob produces a silently ignored value.

Measured on nub **0.7.5** (Homebrew, macOS arm64), one source at a time, observing `nub bin --global` and where `nub add -g <pkg>` symlinks the bins:

| Source for `global-bin-dir` | `nub bin --global` | `nub add -g` links bins into | `nub config get global-bin-dir` |
|---|---|---|---|
| user `~/.npmrc` | default (`~/.local/share/pnpm`) | bare `$PNPM_HOME` (not on PATH) | `undefined` |
| project `.npmrc` | override ✔ | override ✔ | override ✔ |
| `npm_config_global_bin_dir` env | override ✔ | override ✔ | override ✔ |
| `pnpm_config_global_bin_dir` / `PNPM_CONFIG_GLOBAL_BIN_DIR` env | override ✔ | — | override ✔ |
| `AUBE_GLOBAL_BIN_DIR` env | default | — | `undefined` |

### Reproduce (user-npmrc case)

```sh
nub config set global-bin-dir /tmp/my-gb        # → "set global-bin-dir=/tmp/my-gb (~/.npmrc)"
cat ~/.npmrc                                     # global-bin-dir=/tmp/my-gb  ← written by nub itself
nub bin --global                                 # → ~/.local/share/pnpm      (override ignored)
nub config get global-bin-dir                    # → undefined
nub add -g command-code                          # → "Linked 4 bins into /Users/me/.local/share/pnpm" (bare, pre-v11 style)
```

Then, same value in a *project* `.npmrc` (cwd with a `package.json`):

```sh
mkdir -p /tmp/probe && cd /tmp/probe && echo '{"name":"t","version":"1.0.0"}' > package.json
echo 'global-bin-dir=/tmp/my-gb' > .npmrc
nub bin --global                                 # → /tmp/my-gb ✔
```

### Why it matters

[...]

Issue #626, opened 2026-07-30
nub config set should write nub.jsonc, not .npmrc


colinhacks:
`nub config set` writes 136 settings into `.npmrc`, including keys nub does not read back. It should write `nub.jsonc` and fall back to `.npmrc` only on reads.

## The problem

`nub config set <key>` routes any key that is not a `nub.jsonc` field to the engine's config path, which appends it to `.npmrc`. That surface is 136 settings wide. Three things are wrong with it:

**It offers to set keys nub ignores.** Several are pnpm-shaped settings nub does not honor, or honors only under a pnpm incumbent. Writing a key into a config file that nub then does not read back is worse than not accepting it.

**It duplicates `nub.jsonc`.** Nine keys set the same behavior two ways, through different files, with different validation — `nodeLinker`, `shamefully-hoist`, `public-hoist-pattern`, `hoist-pattern`, `minimumReleaseAge`, `minimumReleaseAgeExclude`, `nodeOptions`, `verifyDepsBeforeRun`, `modules-dir`. `nub.jsonc` gets typed validation and a comment-preserving CST write; `.npmrc` gets a flat `key=value` append.

**It only makes sense for pnpm 10 anyway.** pnpm 11 moved behavior settings out of `.npmrc` into `pnpm-workspace.yaml`, so the compat argument covers one major. A project still on pnpm has `pnpm config` available and should use it.

## The shape

`nub config set` writes `nub.jsonc`. Reads keep falling back to `.npmrc` so existing projects work unchanged. Nub stops writing a file it does not own.

## What has to be decided

**Which settings earn a field.** Roughly 90 of the 136 are behavior settings and plausible candidates; ~20 are registry and auth; ~10 are machine-level directories. Each field that lands needs a typed representation, a schema entry in both published schemas, validation, a docs section, and a precedence story against the `.npmrc` value it now shadows. The existing 13 fields were curated individually and the new ones deserve the same scrutiny — this is the substance of the work, not the routing change.

**Auth must not move.** `//host/:_authToken`, `//host/:_auth`, `@scope:registry` write credentials. A project `nub.jsonc` is committed; `.npmrc` already has the conventions and gitignore habits for secrets. `nub login`, `nub logout`, `nub token`, and `nub whoami` already exist and are the credential path.

[...]

PR #833, opened 2026-09-02
config: only write a key that something reads back

colinhacks:
Refs #626
Refs #745

`nub config set` wrote every setting it accepted, including keys no reader under nub consults.

- 15 settings whose only reader is a code path nub does not route join the embedder's `unsupported_settings`, each refused with advice naming a replacement.
- `pnpmfilePath` / `globalPnpmfile` have no `.npmrc` spelling, so the writer's fallback wrote the key verbatim. Refused per route, leaving the workspace-yaml home intact.
- `preferFrozenLockfile` is honored instead: the install family passed a constant `None`. An unset project resolves as before.
- `nub pm use nub` warn-drops an unsupported setting rather than migrating it.

This is the ignored-key half of #626. Routing supported settings to `nub.jsonc`, and the nine duplicate homes, are untouched, so the issue stays open.


colinhacks:

PR #835, opened 2026-09-02
feat(config): a nodeExecutable field with command substitution

### LINKED ISSUE #606: nub.jsonc: a nodeExecutable field with command substitution
Supersedes the never-fetch framing this issue was filed with. Refs #564.

Add `nodeExecutable` to project `nub.jsonc` — the semantics `NODE_EXECUTABLE` already has (top precedence, bypasses the pin chain, floor check still applies), plus command substitution:

    { "nodeExecutable": "$(mise which node)" }

A non-zero exit fails the run, by design: a team wanting no silent fallback to a download delegates to its own toolchain, and an unbootstrapped machine stops instead of substituting. Covers mise, asdf, fnm, Volta, a wrapper, or a literal path with no per-tool knowledge.

Seam: `node_executable_from` (`discovery.rs:955`) is already parameterized — pass the config value in directly, no process-env round-trip.

Also extend `engines_disagreement_warning` (`discovery.rs:792`): a `NODE_EXECUTABLE` win is compared against `engines.node` only, and is silent against `devEngines.runtime`, `.node-version`, `.nvmrc`, and `.tool-versions`. Cover the whole chain, non-fatal.

Blocked on #587.

colinhacks:
Adds `nodeExecutable` to `nub.jsonc` with the semantics `NODE_EXECUTABLE` already has — top precedence, bypasses the pin chain, floor check and tier dispatch still apply — plus command substitution:

```jsonc
{ "nodeExecutable": "$(mise which node)" }
```

A non-zero exit fails the run rather than falling back to a download, so an unbootstrapped machine reports the toolchain's own error instead of silently substituting a different Node.

## Notable decisions

**Published once, not threaded through as a parameter.** `node_executable_from` takes the value fine, but ~15 call sites reach discovery and two of them are inside `nub-core` (`nub node ls` / `install`), where no CLI parameter arrives. Publishing the precedence-resolved value gives the field the same reach the environment variable has, rather than a subset that silently differs by route.

**Resolution is lazy.** Running the command at snapshot time would make a broken toolchain fail `nub config set nodeExecutable …` — locking the user out of fixing the value that broke.

**Two anchors, each to where its value came from.** A relative path resolves against the `nub.jsonc` that declared it, so one committed value names one binary wherever you stand. A `$(command)` runs in the working directory, because a toolchain manager answers per directory: anchoring it to the file would report the repo root's Node for a monorepo member, and a global file's command would report the machine's Node for every project on it.

**Only a whole-value `$(…)` substitutes.** `$(brew --prefix)/bin/node` stays a literal path — the exit-status contract needs exactly one command to be about.

## Two fixes that ride along

`engines_disagreement_warning` compared an explicit-binary win against `engines.node` alone. The override is the one winner that bypasses the chain entirely, so it can contradict `devEngines.runtime` and the pin files with nothing said; it is now compared against the chain's winner too, non-fatally.

`nub node which` printed the override's binary and then credited a pin file that had no say in it. Pre-existing for `NODE_EXECUTABLE` and committable now, so fixed. The `nub node` group initializes no config snapshot by design — a malformed `nub.jsonc` in any ancestor must never block `nub node install` — so the field is published there best-effort, leaving that guarantee intact.

## Verification

Whole-workspace `cargo test` green and `clippy --all-targets --all-features` clean, both on a builder VM. Eight new tests: both spellings through `nub node which`, both anchors, the failing-command refusal carrying the tool's stderr, the chain-wide warning with a silent-when-agreeing control, and the value grammar. Every fixture carries an unsatisfiable `999.0.0` pin, so no assertion can pass unless the field actually took effect.

Closes #606


colinhacks:

PR #836, opened 2026-09-02
install: retry dependency builds a previous install could not run

### LINKED ISSUE #764: Install reports success when a dependency's build script never ran
The install freshness check records no evidence that the previous install's lifecycle phase completed, so a materialized `node_modules` whose build scripts never ran cannot be healed by a later install — it reports `Already up to date` and exits 0.

Measured on a Windows VM. With `node_modules` already materialized, the install short-circuits at the frozen `verify_install_layout` check; `NUB_DIAG_FILE` shows the frozen checks and then no `run_script` or spawn entry. On a box where the same package did run its `postinstall`, the `run_script` entry is present. Clearing `side-effects-v1` does not change it.

Originally reported as a second symptom of #657. It is a separate mechanism: #657 is the lockfile drift check in `aube-lockfile/src/drift.rs`, this is the warm-path short-circuit in `crates/aube/src/state.rs`. Fixing one does not touch the other.

## Not the cause

Three candidate routes, each checked on macOS against `main`, all correct today:

| route | behavior |
| --- | --- |
| dependency build denied by the default policy | skipped, with `WARN_NUB_IGNORED_BUILD_SCRIPTS` naming the package |
| `nub install --ignore-scripts`, then a plain `nub install` | heals — the `run_script` span appears on the second run |
| install interrupted between linking and scripts | cannot seal a bad state; the diag span order is `link` → `run_script` → `dep_lifecycle` → `state`, so state is written after scripts |

So the reproduction is not one of the obvious paths, and it did not reproduce off Windows.

## Where it comes from

The freshness predicate at `state.rs::is_install_needed` compares hashes of the lockfile, the manifests, the on-disk layout and the resolved settings. None of those encode whether the lifecycle phase ran. There is no scripts-completed marker in the state at all — `scripts_ran`, `scripts_completed` and `lifecycle_hash` match nothing in the tree.

The build-policy hash is recorded but only tested for emptiness there (`state.rs:432`); the actual comparison happens later, inside `lifecycle_delta_filter` (`install/finalize.rs:111`), which a short-circuited install never reaches.

That is enough for the archetype to be real: any path that leaves a materialized tree with an unbuilt package and a written state file is permanently sealed. What is missing is the specific path that produces it.

## What would settle it

A Windows repro that names the package and the state it lands in — specifically whether the first install reported the build as skipped, ran it and failed, or never spawned it. From there the question is whether the fix belongs in the freshness predicate (record lifecycle completion, heal on mismatch) or in whatever Windows-specific path drops the spawn.

colinhacks:
The install freshness check recorded nothing about whether the previous install's dependency lifecycle phase completed. Every input it compares describes what the tree was built *from* — the lockfile, the manifests, the layout, the settings — so a materialized `node_modules` whose build scripts never ran was indistinguishable from a complete one. No later install could heal it: each reported `Already up to date` and exited 0.

## Reproduction

macOS, isolated store, same package and defaults throughout. The only variable is install history.

```console
$ npm_config_advisory_check=off nub install     # advisory gate unreachable this run
WARN ignored build scripts for 1 package(s): esbuild@0.24.0
$ nub install                                   # defaults; should build esbuild
nub 0.8.2 · ✓ Already up to date
$ file node_modules/esbuild/bin/esbuild
node_modules/esbuild/bin/esbuild: a /usr/bin/env node script text executable
```

A fresh directory with the identical second command builds it (`WARN defaultTrust: running build scripts for esbuild@0.24.0`, and the bin becomes a Mach-O binary). `nub install --force` heals a sealed tree; nothing else does.

## Mechanism

The `defaultTrust` floor, which nub turns on by default, requires advisory vetting: an OSV `MAL-*` gate that covered this install, or a graph inherited from an already-vetted lockfile. That input is **run-scoped**, so two otherwise-identical installs can disagree, and the one that says no skips the whole dependency-lifecycle phase while still writing state. One install that could not reach the advisory gate therefore sealed the tree permanently.

`dep_build_policy_hash` does fold the floor's posture, but `check_needs_install` only tests it for emptiness; the real comparison lives in `lifecycle_delta_filter`, which a short-circuited install never reaches. Comparing that hash in the freshness check is not available as a fix — `check_needs_install_compute` receives only the project dir and the CLI flags, so with no graph and no OSV result there is no current hash to compare against.

## The fix

Record the builds an install wanted to run and could not, and treat a non-empty set as a warm-path miss.

`DefaultTrustFloor::deferred_by_missing_vetting` reports a package that every gate *but* vetting allows. Packages the config denies stay out: that denial is stable, re-running could never change it, so the ordinary pending-approval steady state keeps taking the warm path and does not turn into a full install on every run.

An allowed package whose materialized directory was not on disk when the phase ran was dropped with a `debug!` line, and nothing else spoke for it — no job means no `run_script` span, and it is not unreviewed, so the ignored-builds warning skipped it too. It now warns under `WARN_NUB_BUILD_NOT_ATTEMPTED` and joins the same retry set.

State written before the field existed cannot say what it deferred, so it re-checks once when it also recorded unreviewed builds. That heals trees already sealed by the old behavior; the following install writes the field and the migration never repeats.

## Verification

Built binary against real fixtures, four arms, all passing:

| arm | expectation | result |
| --- | --- | --- |
| sealed tree, plain install | retries the build | esbuild native |
| control, fresh dir | builds | esbuild native |
| sealed tree, third install | warm path returns once nothing is owed | `Already up to date` |
| control, repeat install | warm path unaffected | `Already up to date` |

The causal chain is confirmed on the state file rather than inferred: the sealing install writes `deferred_dep_builds: ["esbuild@0.24.0"]`, the next install's debug miss reason is `a dependency build did not run on the last install (esbuild@0.24.0); retrying`, the floor then builds it, and the field clears to `[]`.

The anti-regression case that the design turns on: after `nub add`, the state records `unreviewed_builds: ["esbuild@0.24.0"]` with `deferred_dep_builds: []`, and installs stay warm.

`clippy -p aube --all-targets` is clean. The `aube` `e2e` target does not build from the root workspace — its dev-dependencies are unavailable there, which is pre-existing and unrelated to this change; `aube-parity` runs those tests from `vendor/aube`.

## Separately found, not addressed here

A routine `nub add` flips the floor off for already-built packages, producing a spurious `WARN_NUB_IGNORED_BUILD_SCRIPTS`. The cooling window fails closed for lockfile-inherited packages that carry no publish time, because `lockfile_vetted` is false whenever *any* package in the graph was newly picked. This is the behavior the issue thread recorded as unexplained. It is cosmetic today and this change does not regress on it, but the fix means widening when the default-trust floor fires, which is a security-posture decision rather than a bug fix.

Closes #764

### DIFF
--- a/crates/nub-cli/tests/install_engine.rs
+++ b/crates/nub-cli/tests/install_engine.rs
@@ -1816,3 +1816,137 @@ fn ci_rejects_lockfile_whose_root_importer_is_empty() {
         );
     }
 }
+
+/// A dependency build the previous install could not run leaves the tree
+/// incomplete, so the next install must not report it up to date
+/// (nubjs/nub#764). Before the fix nothing recorded that a build was owed:
+/// `check_needs_install` compared only inputs describing what the tree was
+/// built FROM, found them all unchanged, printed "Already up to date" and
+/// exited 0 — permanently, on every later install, with no way out but
+/// `--force` or deleting `node_modules`.
+///
+/// The assertion is the install's own verdict rather than the build's output,
+/// and that is not a shortcut. Three earlier drafts asserted on a marker file
+/// and each answered the wrong question. A marker written inside the dependency
+/// is part of a `file:` package's content, so deleting it to set up the retry
+/// changes the very input the install compares — and the side-effects cache
+/// restores it whether the script re-ran or not. A marker written outside the
+/// dependency never survives the build jail, which scrubs the environment and
+/// confines writes. The verdict has neither problem: it is exactly what the
+/// issue reports and exactly what the fix changes.
+///
+/// This lives nub-side rather than in aube's own e2e suite because the warm
+/// short-circuit is not reachable there at all — measured: under aube's
+/// defaults a `file:` dependency takes the full path on every install, with or
+/// without a build script, so only a dependency-free project ever goes warm.
+#[test]
+fn an_owed_dependency_build_stops_the_next_install_reporting_up_to_date() {
+    let dir = pm_tmpdir("owed-build-retry");
+    let dep = dir.join("plainbuild");
+    std::fs::create_dir_all(&dep).unwrap();
+    std::fs::write(
+        dep.join("package.json"),
+        r#"{"name":"plainbuild","version":"1.0.0","scripts":{"postinstall":"node -e \"process.exit(0)\""}}"#,
+    )
+    .unwrap();
+    // The approval is keyed by SOURCE (`name@file:./path`), not bare name — a
+    // bare-name entry never authorizes a source-backed build.
+    std::fs::write(
+        dir.join("package.json"),
+        r#"{"name":"app","version":"1.0.0","private":true,"dependencies":{"plainbuild":"file:./plainbuild"},"allowBuilds":{"plainbuild@file:./plainbuild":true}}"#,
+    )
+    .unwrap();
+    // Dead registry + no retries: the fixture is entirely local, so any
+    // network attempt is a regression and must fail fast rather than hang.
+    std::fs::write(
+        dir.join(".npmrc"),
+        "registry=http://127.0.0.1:1/\nfetch-retries=0\n",
+    )
+    .unwrap();
+
+    // Run an install and report whether it took the warm short-circuit.
+    let up_to_date = || -> bool {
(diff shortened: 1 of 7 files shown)


colinhacks:

PR #837, opened 2026-09-02
pnpmfile: run preResolution on every install and give it pnpm's lockfile shape

### LINKED ISSUE #755: pnpm's preResolution hook working differently in nub
Consider the following **preResolution** hook

```js
 //.pnpmfile.mjs
function preResolution(options, ctx) {
	const lockfile = options.wantedLockfile;
	// lockfile should not be null
	ctx.log(lockfile);
	console.log(lockfile);
	if (!lockfile) return options;
	if (!lockfile.packages) return options;
	const key = Object.keys(lockfile.packages).find(k => k.includes('is-obj'));
	// following should not be null
	ctx.log(lockfile.packages[key].resolution);
	console.log(lockfile.packages[key].resolution);
	return options;
}

export const hooks = {
	preResolution,
};
```

- run `pnpm i`, we can see lockfile and resolution correctly.
- run `nub i`, we can see locfile logged as null

This is [the link](https://github.com/SukeshP1995/test-nub-git-resolver) for the reproduction.

colinhacks:
Four divergences from pnpm's `preResolution` contract, measured against pnpm 10.15.1 and 11.

- The hook only ran when the install re-resolved. pnpm fires it on every install, before any lockfile-freshness decision.
- `wantedLockfile` / `currentLockfile` were `null`. pnpm always passes an object, empty when there is no lockfile.
- The shape was the narrow `afterAllResolved` projection: no `resolution`, internal dep-path keys. It is now pnpm's in-memory `LockfileObject`, built by reusing the lockfile writer's projection.
- A `console.log` in a hook body was swallowed by the shim's result channel.

The second argument is now pnpm's `{info, warn}` logger, and an in-place lockfile edit warns rather than no-opping.

Closes #755

### DIFF
--- a/crates/nub-cli/src/cli.rs
+++ b/crates/nub-cli/src/cli.rs
@@ -1344,6 +1344,20 @@ pub enum Command {
         #[arg(long)]
         ignore_scripts: bool,
 
+        /// Run this pnpmfile instead of the project's own one. Relative
+        /// paths resolve against the project root. Naming a path this way
+        /// is what loads hooks in a project whose incumbent is not pnpm.
+        #[arg(long, value_name = "PATH", conflicts_with = "ignore_pnpmfile")]
+        pnpmfile: Option<PathBuf>,
+
+        /// Run this pnpmfile before the project's own one.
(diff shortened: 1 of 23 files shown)


colinhacks:

PR #838, opened 2026-09-02
ci: accept the dependency-axis flags pnpm's ci inherits

### LINKED ISSUE #747: [Feature Request] Allow --prod for nub ci
Please add the `--prod` flag for `nub ci`. 

In docker files for node projects especially, we need to ignore `devDependencies` in a layer so we can copy only `dependencies`.

colinhacks:
Closes #747

`nub ci` had no CLI channel for the production/dev split, so a Docker build stage could not produce a devDependency-free tree to `COPY --from`.

pnpm 11 builds `ci` as `clean` + `install --frozen-lockfile` and re-exports install's whole option table (`cliOptionsTypes`/`shorthands` in `pnpm/src/cmd/cleanInstall.ts`), so `-P`/`--prod`/`--production` and `-D`/`--dev` are part of its `ci` surface. The flags land on the `Ci` clap variant exactly as they do on `Install`, and OR into the `DepSelection` that `run_ci` already builds from config-derived signals.

Verified against pnpm 11.25.0 on one identical fixture (`is-number` as a dependency, `ms` as a devDependency):

```
pnpm ci     -> is-number ms       nub ci     -> is-number ms
pnpm ci -P  -> is-number          nub ci -P  -> is-number
pnpm ci -D  -> ms                 nub ci -D  -> ms
```

Workspace `ci -r -P` matches pnpm too, and the root `prepare` hook still runs under `--prod` in both tools.

pnpm 10 rejects both flags on `ci`, but that carries no grammar signal: its `ci` accepts no options at all because the handler is a stub that throws `ERR_PNPM_CI_NOT_IMPLEMENTED`. The comment above the grammar test recorded the opposite conclusion from that stub, so it is rewritten.

### DIFF
--- a/crates/nub-cli/src/cli.rs
+++ b/crates/nub-cli/src/cli.rs
@@ -1414,6 +1414,14 @@ pub enum Command {
     /// Clean install for CI: delete node_modules, install strictly from the
     /// lockfile (drift or a missing lockfile is a hard error).
     Ci {
+        /// Skip devDependencies; install only production deps.
+        #[arg(short = 'P', long, visible_alias = "production")]
+        prod: bool,
+
+        /// Install only devDependencies.
+        #[arg(short = 'D', long, conflicts_with = "prod")]
+        dev: bool,
+
         /// Skip all lifecycle scripts (root and dependency).
         #[arg(long)]
         ignore_scripts: bool,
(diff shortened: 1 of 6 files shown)


colinhacks:

PR #839, opened 2026-09-02
launcher: pin the probe directory's mode instead of taking it from the umask

colinhacks:
The probe directory was made with `create_dir_all`, whose mode comes from the umask, while the read path runs a gate that refuses a group- or other-writable directory. Under a umask of 002 every write landed where every read refused: the probe cache wrote and never hit, so each launch re-ran the Node version probe. This module's `create` exists for exactly this, and the write path was its only non-caller.

Two extraction tests failed for the neighbouring reason — fixtures inheriting `0o664`. Production pins `0o600` and is not exposed.

Invisible under a 022 umask, so CI never saw it. On a umask-002 builder the launcher suite goes from 72 passed / 3 failed to 75 passed.


colinhacks:

Issue #569, opened 2026-07-27
[Feature request] Standalone runtime


fiws:
For deployment it would be nice to have a standalone runtime meaning `nub script.ts` without any other feature (a `nub-run` binary for example).
Why? Because of attack surface (main reason) and final image size.

Unless it is not recommended to deploy nub to prod in the first place? then i guess #567 would solve this.

Background: Currently using [distoless node](https://github.com/GoogleContainerTools/distroless/tree/main/nodejs) and wanting to add nub to it, but it kinda defeats the point. Distroless excludes stuff like the package manager on purpose.

joealden:
@colinhacks I believe this can be closed now that `nubr` exists in 0.9.2?

PR #840, opened 2026-09-02
run: rename the standalone loader to @nubjs/runner and give it a nubr command

colinhacks:
Renames `@nubjs/loader` to `@nubjs/runner` (and its 8 platform packages), and adds a `nubr` command that runs a TypeScript file, a `package.json` script, or an installed bin from `node_modules/.bin`. Names resolve most-specific-first; a script wins over a bin of the same name, matching npm.

A file run re-execs Node with `--import`. In-process was ~30ms cheaper but diverged: `Module.runMain` cannot load an ES module below Node 22.15, and worker threads take their preload from the option store, not `process.execArgv`.

**Before releasing:** run `scripts/npm-first-publish-runner.sh` once, or the release preflight fails on the missing names. Deprecate `@nubjs/loader*`.

Refs #569


colinhacks:
All three findings reproduced against npm/node and are fixed in a63a6e8.

Arguments are now shell-quoted per platform. A leading flag hands the whole argv to Node verbatim rather than enumerating which options take a value, so split-value forms work. Script runs get the manifest-derived `npm_package_*` environment.

`tests/run` gains four checks pinned to exact npm/node output; each was confirmed to go red when its fix is reverted.

colinhacks:
Windows escaping fixed in 636c092 by porting npm's `escape.js`, the same source `crates/nub-core/src/workspace/shell_escape.rs` uses for `nub run` — MS backslash/quote rules, the caret pass, and the repeated pass for `.cmd`/`.bat`. Both files now note that divergence from npm is the bug.

Moved to `runtime/nubr-escape.mjs` with `scripts/nubr-escape.test.mjs` pinning it to the Rust port's expectations, since CI's `test:scripts` glob covers `scripts/` and the cmd.exe cases cannot run here. PATHEXT limitation carried over rather than dropped.

Checked against npm on nine argument shapes; all byte-identical.

colinhacks:
Shipped in v0.9.2: https://github.com/nubjs/nub/releases/tag/v0.9.2

PR #842, opened 2026-09-02
config: refuse an .npmrc write that nub.jsonc already answers

### LINKED ISSUE #626: nub config set should write nub.jsonc, not .npmrc

colinhacks:
Closes #626, stacked on #833.

A `nub.jsonc` install field lowers into engine settings at the `project_config` tier, which outranks project `.npmrc`. Once the field is set, an `.npmrc` line for the same setting is read by nothing — `config set` reported success and no resolve consulted the value. Measured in both directions against a real install.

This refuses that write and names the field to set instead. Keyed on what the project actually supplies, not on the key alone: `linker: "hoisted"` decides only the linker, so `hoist-pattern` still routes to `.npmrc` there.

Not a redirect — the two surfaces disagree on value grammar, and moving the write would desync `get` from `set`.

### DIFF
--- a/crates/nub-cli/src/pm_engine/duplicate_home.rs
+++ b/crates/nub-cli/src/pm_engine/duplicate_home.rs
@@ -137,7 +137,7 @@ pub(super) fn shadowed_error(key: &str, field: &str) -> anyhow::Error {
     anyhow!(
         "nub config set {key}: this project sets `{field}` in nub.jsonc, which outranks .npmrc, \
          so the value written here would never be read\n\
-         \x20\x20set `{field}` instead, or remove it from nub.jsonc to configure this from .npmrc"
+         \x20\x20run `nub config set {field} <value>` instead, or remove `{field}` from nub.jsonc"
     )
 }
 
--- a/crates/nub-cli/src/pm_engine/duplicate_home.rs
+++ b/crates/nub-cli/src/pm_engine/duplicate_home.rs
@@ -1,0 +1,194 @@
+//! Engine settings a `nub.jsonc` field ALSO supplies, and the refusal that
+//! keeps a write out of the file that loses.
+//!
+//! Twelve engine settings can be supplied by a `nub.jsonc` field. The lowering
+//! in [`super::lower_native_install_settings`] injects them at the engine's
+//! `project_config` tier, which outranks every file source — `project_npmrc`
+//! included — so once the owning field is set, an `.npmrc` line for the same
+//! setting is read by nothing. `nub config set` reports success, the file gains
+//! a line, and no resolve ever consults it: the same silent no-op
+//! [`super::store_config_family::npmrc_first`] refuses elsewhere, reached from
+//! the other side. There the key was never readable; here it is readable in
+//! general and answered by a different file in THIS project.
+//!
+//! So the refusal is conditional on what the project actually sets, not on the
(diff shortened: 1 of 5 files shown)


colinhacks:

PR #843, opened 2026-09-02
runtime cache: report why a base was refused instead of naming a cause

colinhacks:
The cold path printed `unsafe owner or permissions` for every refused runtime-cache base — a read-only filesystem, a reparse point, and a transient failure of the Windows ACL query all reached the log as the same sentence. The walkers already produce a specific `io::Error`; `.ok()?` discarded it one frame above where it was needed.

Splits `safe_base_or_reason` out of `ensure_safe_base`, which keeps its `Option` signature so its existing callers are untouched.

Behaviour is unchanged: a base is declined either way, and only the log differs.

Found while attributing an intermittent Windows CI failure where the runtime cache silently relocated and the log could not say why.


colinhacks:

PR #846, opened 2026-09-03
compile: cut compiled-artifact warm start roughly in half

colinhacks:
Warm overhead over the same bundle on installed Node: 23.5 ms → 12.0 ms (hello-world, min of 120 runs).

- New `sealed_module_graph` manifest field: a bundle with no `--external` and no retained computed `import()` skips `--js-defer-import-eval` (~6-7 ms; a non-default V8 flag forfeits the snapshot fast path). Legacy manifests decode as false.
- On Unix the launcher `exec`s Node instead of spawn+forward+wait (~5 ms user, ~4 ms sys): native signals/TTY/exit status. Windows keeps the spawn path.

Augmentation sweeps match a pre-change control on 26.7.0/26.8.1/24.18.0 and `--smol`; the shared failure is the documented `--enable-source-maps` withholding (nodejs/node#63169).


colinhacks:
be6d35e5d7 addresses the review:

- Sealing now also requires no verbatim `.js`/`.mjs`/`.cjs` payload file (includes, emitted asset copies, island contents). Verified with a fixture: an included `import defer` module loaded via computed `require()` works before and after; it keeps the flag.
- The doc sentence now names both comparators: 12.0 ms over the neutral bundle, 7.3 ms over the program's own bundle.
- The Unix signal change is intentional: signaling the artifact PID behaves exactly as signaling plain `node` — the compat baseline. The old group-forward was launcher-added behavior; `docker stop` and CI cancellation signal the group or namespace and are unaffected.

colinhacks:
8c61b6f134: sealing now requires no verbatim payload file at all — the CJS loader parses an exact-path `require()` of any unknown or absent extension as JS, so an extension allowlist cannot prove a shipped file is not runtime JavaScript. Verified: a required `.foo` file containing JS keeps the flag and loads; a payload with no verbatim files still seals.

colinhacks:

PR #848, opened 2026-09-03
npm lockfile: emit the link pair for a `link:` directory dependency

colinhacks:
`nub add ./dep` writes `link:./dep`, and the npm lockfile writer recorded only the root importer entry. The freshness check reads a dep's specifier off the `node_modules/<name>` node, so the importer parsed back empty and frozen installs failed with `manifest adds dep@link:./dep`.

npm has no separate `link:` encoding — a `file:` dir dep is written as a link record — so `emit_file_dep_links` already built the right pair and skipped only that variant. The pnpm writer was fine.

Widen the match to `LocalSource::Link`, skipping links whose target is an importer so the workspace pass keeps members.

`file:`, workspace and `../` output unchanged. aube 3006, nub-cli 1090, clippy/fmt clean.

Refs #755.

### DIFF
--- a/vendor/aube/crates/aube-lockfile/src/npm/tests.rs
+++ b/vendor/aube/crates/aube-lockfile/src/npm/tests.rs
@@ -842,6 +842,95 @@ fn test_write_nests_a_conflicting_member_local_dep() {
     );
 }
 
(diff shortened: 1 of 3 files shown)


colinhacks:

PR #849, opened 2026-09-03
pm: give Remix 3 projects a project-local store

colinhacks:
A fresh `npx remix@next new` app fails its first page load under Nub's default install with `IMPORT_OUTSIDE_MOUNTS`: Remix 3's asset server serves npm packages only from mounts inside the app root, by real path, so packages in the machine-global store are rejected. npm, pnpm, and `--node-linker hoisted` layouts pass.

Adds `remix` to the whole-install GVS-off list, version-gated at major 3 like the `expo` gate; the install report names the reason and the docs list Remix 3.

Verified on the RC template: `nub server.ts` and `nub run dev` now serve the page; main reproduces the failure.


colinhacks:

PR #851, opened 2026-09-03
build: dedupe dependencies and compact the registry primer

colinhacks:
linux-x64 release binary (with `compile`, before the runtime blob): 58.6 MB → 52.6 MB stripped.

- One reqwest (0.13), one rustls provider (aws-lc-rs), sha2 0.11, toml 1.1, clx 3, two YAML stacks instead of four.
- toml 1.x: bunfig and data-format loaders parse a `Table`, not a `Value`.
- Primer: raw SHA-512 digests and a 3-year age prune keeping the highest of each `major.minor` line and every dist-tag target (schema s4): 97,292 → 58,806 versions, 10.05 MB → 6.35 MB, same 862/1,120 primer hits on the bench fixture.
- A pruned seed is flagged sparse, and the resolver refetches any pick a dropped version could outrank (`sparse_pick_needs_refetch`): `lodash@<4.17.21` resolves to 4.17.20 again instead of the held 4.16.6. Line-end picks (`^x`, `~x.y`) and held exact pins stay offline.
- Per-crate `opt-level = "z"` measured and rejected: −7.5 MB but +8.7% warm resolve+link; cold crates alone −0.25 MB.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -67,20 +67,11 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c1daa54020e05aa0b163ee10434fff35a0f18d28a1cafa142bd1290e1abe630e"
 dependencies = [
  "astral-reqwest-middleware",
- "reqwest 0.13.4",
+ "reqwest",
  "secrecy",
  "serde",
  "serde_json",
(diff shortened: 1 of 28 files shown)


colinhacks:

PR #853, opened 2026-09-03
compile: gate the --smol probe cache once per pass

colinhacks:
Warm-start work following #846, which landed the exec launcher and the sealed-graph flag skip. One cost is fixed here; the rest of what the investigation found is recorded below rather than changed.

## The namespace gate ran once per lookup

Reading one remembered `--smol` probe verdict walked the ancestor chain of every candidate cache root — roughly six `symlink_metadata` calls each — to read at most one small file. Discovery calls that once per installed Node, so on a host carrying 91 of them a warm start ran the walk about 720 times. The walk, rather than the verdict, was the cost.

The gate's guarantee is per-pass, exactly as `cache::resolve`'s is: that function validates one root and threads it through the whole run rather than re-checking before each use. The probe directories are now resolved once into a `ProbeStore` and threaded the same way, leaving each lookup a stat and a read. Order and fall-through are unchanged.

Not a process-global memo: `candidates()` reads the environment and this crate's own tests mutate `__NUB_COMPILE_CACHE_DIR`, so a `OnceLock` over it would freeze whichever test called first and make the suite order-dependent.

Warm `--smol` pre-spawn **10.92 → 8.10 ms** as a mean over ten interleaved pairs on one host, per lookup **40.3 → 25.9 us**. The embed shape does not probe and is untouched.

## Overlap with #830

The same investigation found duplicate version-manager roots — `NVM_DIR=$HOME/.nvm` colliding with the `~/.nvm` fallback, so all 91 Nodes were enumerated and probed twice. #830 already fixes exactly that as `dedupe_node_dirs`, so that half was dropped from this branch rather than landed twice. The two compose: #830 halves the number of lookups, this halves the cost of each.

One difference worth a look while #830 is open — its dedupe is lexical by design, so a manager reached through a symlinked spelling is still scanned twice. That is a deliberate call in its doc comment (a syscall per root to catch a shape no default layout produces) and is left alone here.

## Verification

- Launcher gates green: `clippy --all-targets -- -D warnings`, `fmt --check`, 75 tests, release build. `cargo metadata --locked` clean — no dependency change.
- Selection is unchanged: an artifact whose target is satisfied by an installed Node picks the same Node before and after (v26.7.0 either way), and `process.execPath` still reports the outer executable.
- Exercised `--smol` at exact, major and range targets. Exit code 42 propagates, SIGTERM gives 143, SIGINT reaches the handler and gives 130, stdout through `head` is clean.
- `new Worker(new URL('./w.mjs', import.meta.url))` and `child_process.fork(url, args)` match plain Node exactly.

## Two findings for the record

**The eight `--experimental-*` flags are free — do not spend a manifest field on them.** Emitting only the flags a bundle needs was on the table for this work. Measured on a real artifact at n=240, adding `--disable-warning=ExperimentalWarning`, the seven `--experimental-*` flags and `--enable-source-maps` to the child costs **+0.01 ms** of CPU. The positive control in the same run, `--js-defer-import-eval`, costs +6.04 ms, which independently confirms #846's figure and shows the instrument works. Those flags are Node-level rather than V8-level, so they do not forfeit the startup snapshot the way the V8 flag did. Recording a needed-flag set would add a payload field and a compile-time analysis, and carry a wrongly-dropped-flag breakage risk, to buy nothing.

**Where the remaining embed overhead sits.** Post-#846, over plain `node hello.js` at n=240: launcher image ~4.2 ms, launcher pre-spawn ~1.2 ms, the `--require` bootstrap ~2.0 ms, the emitted bundle ~4.0 ms, flags ~0.7 ms (noise). Artifact size is not a factor — pre-main cost is 4.15 ms for the 29.5 MB embed binary and 4.06 ms for the 0.9 MB smol one, so shrinking artifacts will not help warm start. The largest addressable item is the bundle: `runtime/polyfills.cjs` ships 47.7 KB minified into every artifact and installs DisposableStack, the Iterator surface, `RegExp.escape`, `Promise.try`, Set methods, `Array.fromAsync` and more, all native on Node 26 and all still parsed and run. The `nub:polyfill:<name>` region mechanism and `strip_native_polyfills` already exist but cover only the five npm-backed polyfills. Extending them needs a per-polyfill native-since floor and has to respect `--smol`'s floor target rather than the probed version, so it is left for a decision rather than taken here.

## The announcement post's startup chart

#826 charts warm startup at 58.1 ms for the embed shape and 60.5 ms for `--smol`. Both #846 and this change move the `--smol` number, and #846 moves the embed number, so that chart needs re-measuring before v0.9 ships. Not edited here.

Refs #846, #830, #826

### DIFF
--- a/crates/nub-launcher/src/cache.rs
+++ b/crates/nub-launcher/src/cache.rs
@@ -271,7 +271,7 @@ const PROBE_DIR: &str = "smol-probe";
 ///
 /// The probe directory is made by `create`, never `create_dir_all`, and the two
 /// are not interchangeable here: `create_dir_all` takes the directory's mode from
-/// the ambient umask, and [`read_node_version`] runs the same leaf gate that
+/// the ambient umask, and [`ProbeStore::resolve`] runs the same leaf gate that
 /// rejects any group- or other-writable directory. Under a umask of 002 that
 /// `create_dir_all` yields 0o775, so every write lands in a directory every read
 /// then refuses — the cache writes and never hits, silently, exactly the failure
--- a/crates/nub-launcher/src/cache.rs
+++ b/crates/nub-launcher/src/cache.rs
@@ -266,46 +266,6 @@ fn resolve_from<T>(
 /// verdict then costs one launch on a mislabelled Node, not a safety property.
 const PROBE_DIR: &str = "smol-probe";
 
-/// A remembered probe verdict for `node_bin`, if one exists and still applies.
-///
-/// Returns `None` on any doubt — a missing, untrusted, malformed, or stale entry
-/// all mean "probe it again", which is always correct and merely slower.
-pub fn read_node_version(node_bin: &Path, stamp: &str) -> Option<String> {
-    for candidate in candidates() {
-        let Some(base) = candidate.path else { continue };
-        let path = probe_path(&base, node_bin);
-        // The namespace gate walks DIRECTORIES, so it is applied to the directory
-        // holding the entry; the entry itself is then checked directly. Handing it
-        // the file made every read miss, silently — the cache wrote and never hit.
-        let Some(parent) = path.parent() else {
-            continue;
-        };
-        if inspect_existing(parent).is_err() {
-            continue;
-        }
-        let Ok(metadata) = fs::symlink_metadata(&path) else {
-            continue;
-        };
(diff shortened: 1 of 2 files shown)


colinhacks:

PR #854, opened 2026-09-03
pm: degrade gracefully inside coding-agent sandboxes

colinhacks:
Codex CLI and Claude Code run shell commands inside deny-by-default sandboxes: writes limited to the workspace and temp dirs, network off (Codex) or allowlisted through a local proxy (Claude Code). Both announce themselves through env vars (`CODEX_SANDBOX`, `CODEX_SANDBOX_NETWORK_DISABLED`, `SANDBOX_RUNTIME`). This makes `nub install` and friends work there instead of failing on the first write to `~/.local/share/nub/store`, sitting through a 90-second retry ladder against a denied network, or — with the registry allowlisted in Claude Code — failing every TLS handshake.

## Changes

- **Project-local store fallback with read-through.** When the default store is not writable, new packages go to `node_modules/.nub-store` with a `WARN_NUB_STORE_FALLBACK` warning, and the store reads through to the unwritable global store for everything it already holds: CAS files, package indexes, no-integrity URL→sha512 bindings, and nub's phantom-scan sidecars. Writes, index saves, and stale-entry invalidation always target the writable store; content the global store already holds is shared from there, never re-imported. A `storeDir` that merely restates the embedder profile's default counts as the default; a user-set one is used verbatim. nub's phantom producer and consumer (`dynamic_phantom`, `phantom_closure`) resolve the same layered store as the install itself.
- **Fail fast on a hard network deny.** Every transport error is classified once (`aube_registry::Error::NetworkDenied`): `EPERM`/`EACCES`, unreachable network/host, or — inside a detected agent sandbox — a refused connection or proxy tunnel. Those are not retried, and the OSV/downloads probes are skipped when Codex declares network off.
- **Sandbox-aware failure text, only on evidence.** The packument, tarball, and fetch failures carry help that names the sandbox and the two remedies (rerun outside it, or `--offline`) only when the failure is a classified network deny; an auth, integrity, or 5xx failure inside a sandbox keeps its ordinary help. Registry errors now render their source chain (`… client error (Connect): tunnel error: unsuccessful`) instead of a bare "error sending request".
- **TLS inside Claude Code's sandbox.** `srt` on macOS seals off the trust daemon the platform verifier needs, so every handshake failed with `OSStatus -26276` even for an allowlisted registry. Under that sandbox the registry client verifies against the bundled webpki roots (extra roots from `NODE_EXTRA_CA_CERTS` / `ca` / `cafile` still merge in).
- **Actionable store-write error.** A permission denial on a store path says the store is not writable and names `store-dir` (and the sandbox) instead of a bare I/O error.

## Verification

- `tests/agent-sandbox/` (landed in 0a017060bb) regenerated against this build inside the real `codex sandbox` and `srt` default profiles, plus `srt` with `registry.npmjs.org` allowlisted. Control 9/9; registry-allowlisted Claude Code 9/9, including a cold install that ejects `@firebase/database@1.0.8` from the project-local store (the probe reports which path was materialized); Codex and no-network Claude Code pass everything except the two flows that need the registry, which now fail in one attempt with the sandbox named. `node run.mjs --check` is clean.
- Unit tests: `aube-store` (fallback sharing never writes toward the fallback, stale fallback index left in place and shadowed), `aube-resolver` (help fires only for a classified deny), `aube` (binding read-dir layering), `aube-util`, `nub-cli` (sidecar read-through).
- Remote gates on the final commit: root `clippy --all-targets --all-features`, root `cargo test`, and `cargo test --workspace` in `vendor/aube`.


colinhacks:
Pushed fixes for all four findings, plus one the new harness cell exposed.

- Reads resolve through to the fallback; CAS/index writes, marker creation, and invalidation target the primary only. Content the fallback holds is shared, never re-imported.
- No-integrity bindings read the local dir then the global one, and write local.
- The blocked-network help fires only for a classified deny (`Error::NetworkDenied`); auth, integrity, and 5xx failures keep their own help. The tarball-fetch surface carries the same help.
- The phantom producer and consumer use the engine's layered store (`resolved_project_store_v1_dirs`), with sidecar read-through.
- New: under Claude Code's `srt` on macOS, TLS verification failed for an allowlisted registry (`OSStatus -26276`, trust daemon unreachable); the client now verifies against the bundled roots there.

Harness: control 9/9; `srt` with the registry allowlisted 9/9, including a cold install that ejects `@firebase/database@1.0.8` from the project-local store; Codex and no-network `srt` fail only the two registry-dependent flows, in one attempt, with the sandbox named. `--check` clean.

colinhacks:

PR #855, opened 2026-09-03
pm: fix npm workspace importer detection, and give afterAllResolved pnpm's lockfile shape

colinhacks:
Two fixes in the npm/pnpmfile area, both surfaced while investigating #755.

**npm lockfile reader.** npm records a workspace member and a `file:` local directory dependency identically — a bare path key plus a `link: true` record — so the reader registered targets like `vendor/local` as workspace importers. The freshness check then rejected the lockfile with `workspace importer vendor/local is in the lockfile but not in the workspace`, failing every `--frozen-lockfile` for a workspace with an out-of-tree local dep, including against lockfiles npm itself wrote. Verified against a real npm 11.19.0 lockfile: exit 16 to exit 0.

Link targets are now matched against the workspace patterns, which is the only thing that can tell the two apart. That matcher has to agree with `find_workspace_packages`, and getting it wrong in the exclusive direction costs a real member its importer entry — so it now follows the discovery branch each pattern actually selects (`**` present or not), is explicitly case-sensitive, normalizes `./`, interior `.`, trailing `/` and an in-root `..`, and skips `node_modules`. A ten-case differential asks the walk and the matcher the same question so the two cannot drift.

**`afterAllResolved` hook shape.** `preResolution` was fixed to hand hooks pnpm's in-memory `LockfileObject`; `afterAllResolved` still got a narrow aube-shaped one, so a hook reading `lockfile.packages[k].resolution.integrity` threw. It round-trips, so the writer now also returns the snapshot-key to graph dep_path map, and edits apply as a delta — which is what stops optional dependencies being silently deleted. Measured on a builder VM: nub's projection now matches pnpm 10.15.1 field-for-field.

Refs #755

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -323,6 +323,7 @@ dependencies = [
  "aube-manifest",
  "aube-settings",
  "aube-util",
+ "aube-workspace",
  "base64",
  "blake3",
  "glob",
--- a/vendor/aube/Cargo.lock
+++ b/vendor/aube/Cargo.lock
@@ -356,6 +356,7 @@ dependencies = [
  "aube-manifest",
  "aube-settings",
  "aube-util",
+ "aube-workspace",
  "base64",
  "blake3",
  "criterion",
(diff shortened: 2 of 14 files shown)


colinhacks:

PR #858, opened 2026-09-03
compile: add --hide-console for Windows GUI apps

colinhacks:
Bun ships `--windows-hide-console` and Deno ships `--no-terminal`. Nub had neither the flag nor the mechanism, so anything with a GUI, a tray icon or a file-association handler could not ship on it.

## Why the subsystem flip is only half of it

Bun and Deno **are** the process they hide, so setting the PE subsystem to GUI is their entire fix. Nub's launcher spawns Node, and Windows allocates a console for a console-subsystem child whose parent has none — so the flip alone just moves the flash from the launcher to Node, and on a `--smol` first run to `curl`/`wget` and the version probe as well.

So the artifact carries the decision. `Manifest::hide_console` is `#[serde(default)]`, so older payloads still load, and the launcher passes `CREATE_NO_WINDOW` on every spawn it makes. That is process-wide state rather than a threaded argument because the spawn sites are spread across several call paths, and it is armed once from the payload before the first of them runs.

**Suppression is conditional on no console being attached already.** A GUI process started from `cmd.exe` inherits that console, and suppressing the child's would throw away output the user is watching; started from Explorer there is nothing to inherit. Bun gets that split for free by being one process; this reproduces it rather than picking a side.

## Where the flip is applied

To the finished image, not the template: libsui rebuilds the PE headers while rewriting the resource directory, so a subsystem set on the way in is discarded — the same reason the icon and version resource are set inside that builder chain. `verify_artifact` reads the field back, since a cross-compiled Windows binary is never executed on the build host.

`Subsystem` sits at optional-header offset 68 in both PE32 and PE32+, so the position is `pe + 24 + 68` with no magic-dependent arithmetic. Deno computes it as `standard_fields_size + 68`, which is correct only because it ships 64-bit alone; on PE32 that lands in `DllCharacteristics`.

## Verification

`tests/compile-hide-console/run.sh` runs on the Windows legs of `compile-native.yml`. It cannot see a window, so it asserts what a headless runner can: the subsystem byte with a console-subsystem control, that the artifact still runs and delivers output through a redirect, and — from the timing trace — that the launcher actually took the suppressing path, so a runner that owns a console reports itself inconclusive rather than passing for the wrong reason.

`observe-console.ps1` asks the question directly on a real desktop: launch through ShellExecute as Explorer does, count consoles allocated, with a non-hidden build as the control and a marker file proving the program ran despite its invalid standard handles. Deliberately not wired into CI, which has no interactive desktop.

Also repairs a doc comment #830 split across two tests in `compile/mod.rs`.

Refs #830

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -26,3 +26,6 @@ tests/fixtures/loader-redirect-both-ways/*  text eol=lf
 # native host rather than Linux alone. `**` rather than `*` because the `.d`
 # companion directories nest, and their sources feed the same byte comparisons.
 tests/compile-augmentation/fixtures/**  text eol=lf
+# PowerShell 5.1 reads a BOM-less script in the ANSI codepage and is happiest with
+# CRLF; keep this probe byte-identical on every checkout.
+tests/compile-hide-console/*.ps1 text eol=crlf
--- a/.github/workflows/compile-native.yml
+++ b/.github/workflows/compile-native.yml
@@ -20,6 +20,7 @@ on:
       - 'tests/compile-augmentation/**'
       - 'tests/compile-corpus/**'
       - 'tests/compile-metadata/**'
+      - 'tests/compile-hide-console/**'
       - '.github/workflows/compile-native.yml'
   pull_request:
     branches: [main]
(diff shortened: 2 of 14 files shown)


colinhacks:

PR #859, opened 2026-09-03
compile: scope --allow-dynamic-import with a glob

colinhacks:
`--allow-dynamic-import` was all-or-nothing: refuse every computed `import()`, or excuse every one of them. A plugin loader in one directory meant excusing the whole program.

It now takes an optional, repeatable glob. The default is unchanged — absent still refuses everything, which stays safer than Bun's allow-everything.

```bash
nub compile app.ts --allow-dynamic-import 'src/plugins/**'
```

## Scoped by importer, not by specifier

This is where it parts company with Bun, deliberately. Bun matches its glob against the specifier's extracted template shape and needs an `'<empty>'` sentinel for the opaque ones — but an opaque specifier is precisely the case the flag exists for, so that matcher can only ever describe the imports that did not need it.

`DynamicSite.module` is known for every site and is the thing an author actually controls, so the glob selects the file the import is **written in**.

## Anchoring

The glob is matched against the module path relative to the working directory, which is the anchor `--include` already resolves its patterns against (`Token::parse(cwd, token)`). A module outside it — a dependency above the project — can be named by no relative pattern and stays refused. That is the intended reading rather than a gap: a scoped escape hatch names the code you are vouching for, and you cannot vouch for a path you cannot write down.

## One deliberate divergence from house style

The flag uses `num_args = 0..=1` **without** `require_equals`, unlike `--sourcemap`. `require_equals` would reject the spaced form `--allow-dynamic-import 'src/plugins/**'`, which is how the flag reads naturally and how it is documented.

Refs #830

### DIFF
--- a/crates/nub-cli/src/cli.rs
+++ b/crates/nub-cli/src/cli.rs
@@ -1256,8 +1256,20 @@ pub enum Command {
         /// time — a plugin loader, a config module. Such an import is refused by
         /// default: the binary resolves it from the directory it is run in, so
         /// what it loads depends on the machine you ship to.
-        #[arg(long = "allow-dynamic-import", help_heading = COMPILE_ADVANCED)]
-        allow_dynamic_import: bool,
+        ///
+        /// Takes an optional glob scoping the permission to the files the imports
+        /// are written in, so one plugin directory can be excused without excusing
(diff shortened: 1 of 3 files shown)


colinhacks:

PR #860, opened 2026-09-03
compile: an error tier for the readout, and four defects the first embed build showed

colinhacks:
Follow-on to #830, which shipped the live line and the block but only exercised `--smol`.

A failed compile escaped to `Termination`'s unstyled `Error:`. It now wears a red `error` label sharing one renderer with `warn`. Its body is reproduced byte for byte; re-indenting flattened the nested `file:line`.

Four defects, found by running an embed build:

- `app 0.0 MB` for a region under 50 KB. Now KB.
- `Signing embedded Node.js` above the block: a step, not a fact. Now a phase.
- One SGR span per word — eleven for one aside. Now one per ink run.
- Diagnostics did not wrap; `unknown --platform` is 155 columns.

834 passed, 0 failed under `--features compile`.

### DIFF
--- a/crates/nub-cli/src/cli.rs
+++ b/crates/nub-cli/src/cli.rs
@@ -3257,7 +3257,11 @@ fn dispatch_subcommand(rest: Vec<String>) -> Result<i32> {
                 // target Node; the CLI layer does not know it yet.
                 target_node: None,
             },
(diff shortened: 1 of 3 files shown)


colinhacks:

PR #861, opened 2026-09-03
pnpmfile: give afterAllResolved pnpm's lockfile shape

colinhacks:
`preResolution` was fixed to hand hooks pnpm's in-memory `LockfileObject`; `afterAllResolved` still got a narrow aube-shaped one. Measured against pnpm 10.15.1 on one fixture: 2 top-level keys vs 10, `packages[k]` as `{name, version, dependencies, peerDependencies}` rather than `{resolution, engines}`, and `importers["."]` an ARRAY where pnpm's is a map with `specifiers`. A hook reading `lockfile.packages[k].resolution.integrity` — the shape #755 was filed about — threw. The source calls this the hook most pnpmfiles use.

This hook round-trips, so the projection alone is not enough: keys and dependency values both arrive in pnpm's spelling. `pnpm::write::build` now also returns the snapshot-key → graph dep_path map, collected where both spellings are already in hand. Edits apply as a delta, so a name the projection never showed the hook is left alone — which is what stops optional dependencies being silently deleted, since the writer strips those from a snapshot's `dependencies` block.

Verified on a GCE builder: nub's projection now matches pnpm 10.15.1 field-for-field, and clippy is clean.

Refs #755

### DIFF
--- a/vendor/aube/crates/aube-codes/src/warnings.rs
+++ b/vendor/aube/crates/aube-codes/src/warnings.rs
@@ -194,7 +194,7 @@ pub const ALL: &[CodeMeta] = &[
     CodeMeta {
         name: WARN_AUBE_HOOK_IDENTITY_REWRITTEN,
         category: category::PNPMFILE_HOOKS,
-        description: "A pnpmfile hook rewrote a package's `(name, version)` identity; aube reverted the edit.",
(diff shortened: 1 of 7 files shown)


colinhacks:
Folded into #855 — one PR for the whole change.

PR #862, opened 2026-09-04
compile: key the Node flag cache per binary instead of one flat map

colinhacks:
Warm-start work on the compiled artifact's launcher, and on the `nub` spawn path it shares.

## The premise moved before the work started

Every figure in the brief for this came from released v0.8.2, which spawned Node and waited. `e3de38c4e9` replaced that with `exec` on Unix. Re-measured on `main` first — linux-x64, hello-world artifact, warm, min of 25 interleaved runs, clean cache, one Node on PATH, no version manager installed:

| | v0.8.2 (darwin, contended host) | `main` (linux-x64) |
|---|---|---|
| embed — launcher's own pre-spawn work | 2.09 ms | **0.56 ms** |
| smol — launcher's own pre-spawn work | 6.80 ms | **0.66 ms** |

End to end: `node -e 0` 27.8 ms, embed 33.8 ms, smol 36.5 ms. The launcher's own work is ~2% of the artifact's start; most of the rest of the gap is Node itself starting from a different binary.

So there were no fixed milliseconds left to remove. Both leads turned out to be **unbounded per-Node terms**.

## What the phase split found

Six finer `__NUB_LAUNCHER_TIMING` phases put `cache revalidated → about to spawn node` almost entirely inside one call, `discovery::accepted_env_flags` — 0.07 ms of 0.09 on smol, with `compute_inject_flags` and `argv_inject_flags` ~0.01 each. Sweeping the flat map's entry count:

| entries in `node-env-flags.json` | file size | flag window |
|---|---|---|
| 1 | 8.9 KB | 0.08 ms |
| 10 | 89 KB | 0.45 ms |
| 50 | 447 KB | 2.09 ms |
| 200 | 1.79 MB | 8.45 ms |

Dead linear: **0.042 ms of `serde_json::Value` parse per remembered Node, per launch**, against ~0.6 ms for the entire rest of the launcher's warm work. At 46 entries — one real developer machine — that is ~1.9 ms, which matches the brief's unattributed figure exactly.

The second lead, the `--smol` version-manager scan, is the same shape and 3.7x shallower: 0.09 ms at 0 installs, 0.17 at 1, 0.28 at 10, 0.73 at 50 → **~0.011 ms per installed Node**.

## What this changes

**One entry file per binary** under `node-flags/`, replacing the two flat `{ "<path>": … }` maps. Both verdicts share the `(path, mtime)` key and are read on the same spawn, so they share a file and a launch reads one — the same shape, for the same reason, as the launcher's own `cache::ProbeStore`. Not launcher-only: `accepted_env_flags` is on `node::spawn`'s path and two `cli.rs` sites, so every non-compat `nub run` / `nub <file>` / `nubx` paid this too.

It also stops one bad entry poisoning every other. A single malformed byte in a shared map made `from_str` fail for **every** binary — a `node` spawn per launch, indefinitely — and the next write reset the map, discarding every other binary's verdicts. Three tests pin the per-binary keying, the shared invalidation, and the collision guard.

`blake3` becomes a non-optional `nub-core` dependency, since it names the entry files. That costs the shipped artifacts nothing — `nub-launcher` already depends on it directly and `nub-cli`'s release build turns `embed-runtime` on — so only a feature-off dev build gains it. `crates/nub-launcher/Cargo.lock` moves with it.

**One orphan-sweep pass over the node store instead of two.** That GC runs on every launch including a fully warm one, and each `cleanup_orphan_entries` call costs a `cache::revalidate` ancestor walk plus a `read_dir` before it can reject a single name. Same entries removed, same ages, half the walks.

**The `probe_node_version` doc comment is corrected.** It claimed the version exec "accounts for the whole warm gap between the two shapes"; a warm launch takes the cache-hit branch and execs nothing, so it cannot. Replaced with the measured split.

## Measured and deliberately left alone

- **The `--smol` rescan.** ~0.011 ms per installed Node, redone every launch with nothing remembering which Node won. A fix means new persisted state with its own invalidation rules — worth its own change and review, and it is the smaller of the two terms.
- **The orphan GC's remaining ~0.06 ms.** Gating it on cold launches would remove it, since garbage is only produced by a launch that extracts — but a warm-forever artifact would then stop reclaiming another payload's 30-day-old tree. A real behaviour change for ~9% of a 0.6 ms budget.
- **`app_cache_is_ready`, ~0.29 ms — the largest single fixed cost, ~45% of a warm launch.** It zstd-decompresses every app payload file and byte-compares it against the on-disk copy, so it is O(bundle size) per start; hello-world hides that. Whether it can be cheapened is a cache-trust question rather than a perf one.

## Verification

`clippy --all-targets --all-features` (all four legs, including `nub-launcher`) and the full `cargo test` are green on a remote builder; the three new tests run and pass.

The direct proof is sweep 1 re-run against the fix. The pre-spawn flag window, same host, same script, same min-of-25 interleaved method:

| sibling entries | before — one flat map | after — one file per binary |
|---|---|---|
| 1 | 0.08 ms | **0.05 ms** |
| 10 | 0.45 ms | **0.05 ms** |
| 50 | 2.09 ms | **0.05 ms** |
| 200 | 8.45 ms | **0.05 ms** |

Flat, and the launcher's whole warm path holds at 0.66 ms at every entry count — the per-Node term is gone rather than reduced. The version-manager sweep is unchanged, as expected, since this does not touch that scan.

One piece of fallout not fixed here, because this branch is under a no-docs constraint: `wiki/research/node-experimental-flag-lifecycle.md:40` still names `~/.cache/nub/node-env-flags.json` as where the verdict is cached. One-line follow-up.

### DIFF
--- a/crates/nub-core/Cargo.toml
+++ b/crates/nub-core/Cargo.toml
@@ -33,7 +33,7 @@ repository = "https://github.com/nubjs/nub"
 # it on via `nub-cli/embed-runtime`. The optional deps are pulled in ONLY here:
 # `ruzstd` (pure-Rust decode at runtime) ships in the binary; `zstd`/`tar`/`sha2`
 # (encode + key-hash at build time) are build-deps that never reach the binary.
-embed-runtime = ["dep:ruzstd", "dep:zstd", "dep:tar", "dep:sha2", "dep:blake3"]
+embed-runtime = ["dep:ruzstd", "dep:zstd", "dep:tar", "dep:sha2"]
 
 [dependencies]
 anyhow = "1"
@@ -98,7 +98,14 @@ ruzstd = { version = "0.8", optional = true }
 # win-arm64/musl), vs ~6 ms for BLAKE3's portable SIMD. BLAKE3 is collision/preimage
 # resistant — fully adequate for integrity — and builds identically on every target.
 # (`sha2` stays for the 32-bit cache KEY, which is not on the hot path.)
-blake3 = { version = "1", optional = true }
+#
+# NOT optional, though the R2 digest use above is `embed-runtime`-only: the
+# per-binary Node flag cache (`node::discovery`, `FLAG_ENTRY_DIR`) names its entry
+# files by hashing the binary's path, and that cache is read on every spawn in
+# every build. Unconditional costs the shipped artifacts nothing — nub-launcher
+# already depends on blake3 directly and nub-cli's release build turns
+# `embed-runtime` on — so only a feature-off dev build gains it.
+blake3 = "1"
 
 [target.'cfg(windows)'.dependencies]
 windows-sys = { version = "0.61", features = [
--- a/crates/nub-core/Cargo.toml
+++ b/crates/nub-core/Cargo.toml
@@ -33,7 +33,7 @@ repository = "https://github.com/nubjs/nub"
 # it on via `nub-cli/embed-runtime`. The optional deps are pulled in ONLY here:
 # `ruzstd` (pure-Rust decode at runtime) ships in the binary; `zstd`/`tar`/`sha2`
 # (encode + key-hash at build time) are build-deps that never reach the binary.
-embed-runtime = ["dep:ruzstd", "dep:zstd", "dep:tar", "dep:sha2"]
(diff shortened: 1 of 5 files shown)


colinhacks:

PR #863, opened 2026-09-04
pm: rename the neutral build allowlist to allowScripts

colinhacks:
npm 12 gates install scripts on a top-level `allowScripts` map in package.json (RFC npm/rfcs#868) — the same slot Nub wrote `allowBuilds` to, same shape, same deny-wins fold. Nub's neutral field becomes `allowScripts` on read and write, so one map serves both tools.

pnpm's `allowBuilds` (pnpm-workspace.yaml, `package.json#pnpm`) is unchanged; `nub pm use pnpm` renames the map back.

A stale top-level `allowBuilds` errors with `ERR_NUB_ALLOW_BUILDS_RENAMED` — ignoring it would drop its `false` denials as well as its approvals.

Also: `INSTALL_SHAPE_FIELDS` follows the key, so an approval still invalidates the install, and `nub remove` now prunes the root entry.

Breaking: rename the key.

### DIFF
--- a/crates/nub-cli/src/cli.rs
+++ b/crates/nub-cli/src/cli.rs
@@ -10294,7 +10294,7 @@ fn run_pm_use(name: &str, spec: &str, cwd: &Path) -> Result<i32> {
 
     // `use pnpm` regenerates pnpm-workspace.yaml from the nub-mode package.json
     // homes (workspaces + catalogs, top-level overrides/patchedDependencies/
-    // allowBuilds/auditConfig) — the exact reverse of `use nub`'s migration.
+    // allowScripts/auditConfig) — the exact reverse of `use nub`'s migration.
     // No-op on a project that never carried them.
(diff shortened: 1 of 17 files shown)


colinhacks:
Pushed four review fixes.

- `nub remove` now prunes pinned `name@spec` allowlist keys, not only the bare name. npm's `approve-scripts` writes the pinned form by default, so the common case survived removal.
- `nub pm use pnpm` accepts a pre-cutover root `allowBuilds` map. That path builds no install session, so the rename refusal never reaches it, and refusing would strand the build policy.
- The `ROOT_ALLOW_SCRIPTS_KEY` doc now says interchangeable with npm for registry keys only.
- Docs: an explicit `false` wins within the map, not across sources.

Merge order across sources is pre-existing and deliberate — the root map is read last so it can override a `pnpm.*` entry.

colinhacks:

PR #864, opened 2026-09-04
compile: add --icu to trim locales from the embedded Node

colinhacks:
Adds an opt-in `--icu` flag to `nub compile` that trims unused locales out of the embedded Node's ICU data.

An official Node is built `--with-intl=full-icu`, so one linked-in ICU data package covering ~700 locales accounts for 31.6 MiB of the ~102 MiB stripped binary — 31.7% of the compressed blob a default-shape artifact ships. Naming the languages an app actually formats for reclaims most of it.

```
nub compile app.ts                # 29,456,306 B — every locale (unchanged default)
nub compile app.ts --icu=en       # 24,898,994 B — 1000 of 4305 ICU resources
nub compile app.ts --icu=en,de    # 24,898,994 B — 1027 of 4305
```

Measured end to end on Node 26.8.1 for darwin-arm64: **−4,557,312 bytes, or 15.5%**. The two trimmed artifacts match in total because their Node blobs differ by only 35 KB and the injection's alignment padding absorbs it; they are not identical, and they format differently.

Locale data is cheap once the full set is gone — twenty languages cost 4 MiB raw more than English alone — so the value list is where the saving sits rather than an on/off switch.

## Flag grammar

Written `--icu=<LOCALES>`; a bare `--icu` means `full`. The `require_equals` form is required, not stylistic — an optional-value flag beside a positional would otherwise consume the entry, and `nub compile --icu en,de app.ts` would fail with `unexpected argument`. This matches `--sourcemap` in the same subcommand.

Matching is on the language subtag, so `--icu=zh` retains `zh_Hans_CN` and `--icu=en-US` retains `en`. The `root` locale is always kept.

## How the rewrite works

ICU reaches its data package through a bare pointer and navigates it by its own table of contents, and nothing records the package's length. A smaller valid package written at the same offset is therefore served normally, and the zero padding behind it survives zstd at under a kilobyte.

Locating the package needs no Mach-O/ELF/PE parser and no symbol table: it announces itself with a two-byte magic plus a `CmnD` format tag, and that pair occurs exactly once per binary. Verified on Node 26.8.1 for darwin-arm64, linux-x64 and win32-x64 — one header each, identical 4305-item table of contents. The symbol-based alternative does not survive `strip` on ELF.

Only locale-shaped resources are dropped. The charset converters, break iterators, normalization tables and every supplemental resource stay, so no API changes behavior: `Intl.Segmenter`, `TextDecoder('shift_jis')`, `String.prototype.normalize` and time-zone arithmetic all work unchanged. A dropped language falls back through ICU's normal chain to `root`.

## Why the default is unchanged

That fallback is silent. `Intl.NumberFormat('de-DE').format(1234.5)` returns `1,234.5` on a trimmed build, with no error, and the publisher choosing the default is not the person who reads the wrong output. `vercel/pkg` shipped small-icu base binaries and collected years of exit-0-wrong-output reports ([#539](https://github.com/vercel/pkg/issues/539), [#1563](https://github.com/vercel/pkg/issues/1563), [#1731](https://github.com/vercel/pkg/issues/1731)). A build with no `--icu` keeps every locale and formats exactly like the user's own `node`.

## Three safety properties

**A trimmed payload never dedups against an official store Node.** The launcher's dedup rests on the embedded and provisioned binaries running identically, which a trim falsifies. Without the gate, one artifact would format through full ICU on a machine that has Node in nub's store and through trimmed data on a machine that does not — a difference the publisher could not reproduce locally. `a_trimmed_payload_refuses_an_official_store_node` covers it, and was confirmed to fail with the gate removed.

**A trim requires `codesign` on a Mach-O target**, rather than degrading to an unsigned binary macOS will not launch. Every other fallback in `prepare_node_bytes` ships the original Node, which is correct but larger; a requested trim turns each of those into an error instead, because silently shipping 700 locales when two were asked for is the build misreporting what it produced.

**A host-target trim is verified by formatting, not by launching.** A package whose per-tree indexes no longer resolve answers `--version` correctly and then aborts inside the first `Intl.DateTimeFormat`, so the existing `node_runs` probe cannot see it. `node_formats_dates` formats a date and segments a string instead.

Refused rather than ignored under `--smol`, which embeds no Node to trim, matching how `--icon` and `--metadata` handle a target that cannot carry them.

## Verification

- Unit tests over the locale-shape rule, the flag parser and the package rewrite against a synthetic package.
- The launcher regression test above, plus its negative control.
- End-to-end: one fixture compiled three ways against the same Node, run, and diffed, with an official Node deliberately left in the store so the dedup gate is under test.

| Probe | default | `--icu=en` | `--icu=en,de` |
| --- | --- | --- | --- |
| `Intl.NumberFormat('de-DE')` | `1.234,5` | `1,234.5` | `1.234,5` |
| `Intl.NumberFormat('fr-FR')` | `1 234,5` | `1,234.5` | `1,234.5` |
| `Intl.DateTimeFormat('de-DE')` | `Mittwoch, 31. Dezember 1969` | `Wednesday, December 31, 1969` | `Mittwoch, 31. Dezember 1969` |
| `Intl.ListFormat('de')` | `a, b und c` | `a, b, and c` | `a, b und c` |
| `resolvedOptions().locale` for `de-DE` | `de-DE` | `en-US` | `de-DE` |
| `Intl.Segmenter` word count | 3 | 3 | 3 |
| `TextDecoder('shift_jis')` | `あ` | `あ` | `あ` |
| `'i'.toLocaleUpperCase('tr')` | `İ` | `İ` | `İ` |
| time zone `Asia/Tokyo` | `9:00:00 AM GMT+9` | `9:00:00 AM GMT+9` | `9:00:00 AM GMT+9` |

German is correct under `--icu=en,de` and falls back under `--icu=en`; French falls back under both, since neither names it. Every API-level probe is unchanged across all three.
- Root and `nub-launcher` clippy at `--all-targets`, `cargo fmt`, and the test suite.

### DIFF
--- a/crates/nub-cli/src/cli.rs
+++ b/crates/nub-cli/src/cli.rs
@@ -1105,6 +1105,21 @@ pub enum Command {
         #[arg(long = "no-minify")]
         no_minify: bool,
 
+        /// Languages to keep in the embedded Node's ICU data, comma-separated:
+        /// `--icu=en,de,fr`. Default is every locale, which formats exactly like
+        /// the user's own `node`. Written `--icu=<LOCALES>`; bare `--icu` is
+        /// `full`. A region is accepted and narrows nothing (`en-US` keeps `en`),
+        /// and `root` is always retained. Dropped languages fall back silently,
+        /// so name every one the app formats for. No effect under `--smol`.
+        #[arg(
+            long,
+            value_name = "LOCALES",
+            num_args = 0..=1,
+            require_equals = true,
+            default_missing_value = "full"
+        )]
+        icu: Option<String>,
+
         /// Where the source map goes: `none` (default), `linked`, `inline`, or
         /// `external`. Written `--sourcemap=<MODE>`; bare `--sourcemap` is inline.
         #[arg(
@@ -3197,6 +3212,7 @@ fn dispatch_subcommand(rest: Vec<String>) -> Result<i32> {
             target,
             platform,
             no_minify,
+            icu,
             sourcemap,
             define,
             define_file,
@@ -3235,6 +3251,10 @@ fn dispatch_subcommand(rest: Vec<String>) -> Result<i32> {
             icon,
             metadata,
             hide_console,
+            icu: match icu.as_deref() {
+                Some(value) => crate::compile::parse_icu_locales(value)?,
+                None => None,
+            },
             metafile: metafile.as_deref().map(PathBuf::from),
             bundle: crate::compile::BundleOptions {
                 minify: !no_minify,
--- a/crates/nub-cli/src/compile/icu.rs
+++ b/crates/nub-cli/src/compile/icu.rs
@@ -190,7 +190,7 @@ pub fn trim(bytes: &mut [u8], locales: &[String]) -> Result<TrimReport> {
     // corruption of exactly the kind this whole rewrite has to be trusted not to
     // produce. Not observed in any Node measured; checked because the cost of being
     // wrong is unbounded and the cost of the check is one pass.
-    if sizes.iter().any(|&size| size == 0) {
+    if sizes.contains(&0) {
         bail!("the ICU data package aliases two resources to one offset; refusing to rewrite it");
     }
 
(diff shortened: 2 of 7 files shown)


colinhacks:

PR #865, opened 2026-09-04
compile: make an extracted app's layout reachable from the code inside it

colinhacks:
Two defects, reproduced on `main` on macOS and Linux with separately built binaries.

A bundled module's `__dirname` was the chunk's directory, while `--include` extracts an asset at its path relative to the anchor. A module in `sub/` reading `path.join(__dirname, "data/x")` got the app root's copy at exit 0, or `ENOENT`, while its own asset sat unreachable at `sub/data/x`. Fixed for CommonJS and ESM `import.meta.dirname`; the entry keeps the value it had, and `import.meta.url` and `node_modules` modules are unchanged.

A pure bundle's extraction dir had no `package.json`, so a `getRoot` walk-up escaped it. A synthesized root manifest stops it, with no `"type"` field.

Refs #830


colinhacks:

PR #866, opened 2026-09-04
compile: run a pure-JavaScript artifact straight from the executable

colinhacks:
A compiled binary extracts its app files to `~/.cache/nub/compile-app/<hash>/` on first run, because Node has to be handed a `--require` preload and an entry by path. A payload that reduces to generated JavaScript needs neither: the launcher passes the bootstrap as `-e` and the bootstrap serves each chunk to `import()` as a `data:` URL read out of the executable's own bytes.

A `console.log("hello world")` artifact wrote 7 files before this change and writes zero after it, `embed` and `--smol` alike. A `--smol` build whose Node is already on the machine now resolves no cache directory at all, so it starts under a read-only `HOME`/`TMPDIR` where it previously refused with `nub: no writable, executable cache directory was found.`

## Qualification — is `sealed_module_graph` the right test?

Necessary, not sufficient. A statically traced worker chunk and a `--sourcemap=linked` map are both files the bundler itself parsed, so `sealed_module_graph` stays true with either present, and neither can be reached from a `data:` URL (`new Worker` does not take one; a linked map cannot be fetched from one). The operative predicate is `sealed_module_graph` AND no worker roots or wrappers AND no source map requested AND every payload file is the bootstrap or a generated `.mjs` chunk AND the chunk graph is acyclic. Anything else extracts exactly as it always has.

Automatic-only, no `--no-extract` flag: a flag promising "no extraction" would be misleading, because an `embed` artifact still unpacks its Node once and a `--smol` one still provisions when none is found. The build summary's new `app` row reports the state and the decline reason on every build.

## Three departures from the obvious shape

- The app region is located by a trailer, not by an offset the launcher passes. `libsui::find_section` returns a pointer into the mapped image on all three platforms (`getsectdata`, `dl_iterate_phdr`, `LockResource`), so no file offset exists on that path. `encode_with_license` appends a 32-byte locator past the container's last declared region, which `decode` already tolerates; the bootstrap scans a bounded tail window backwards for it.
- Inline chunks are brotli, not the zstd beside them. The decompressor is the artifact's own JavaScript, and `zlib.zstdDecompressSync` is absent on Node 23.5 and 23.6 while brotli is on every supported Node. The bootstrap file alone is stored verbatim, so the launcher needs no decoder.
- The chunk set is unchanged and the chunks are linked at run time. Collapsing to one chunk would put `function require(id)` at module scope in an authored-ESM chunk and break `typeof require !== "undefined"` detection, which three existing `bundle.rs` tests assert against. Each chunk becomes its own `data:` URL instead, with cross-chunk specifiers rewritten at build time to `"nub-inline:<name>"` placeholders the loader substitutes — so a payload that does not qualify costs no second bundle.

Two `-e` behaviours the bootstrap undoes, both measured on 26.7.0: Node publishes the CommonJS wrapper's five names as globals, and it puts the whole script into `process.execArgv`. Each chunk deletes the globals from module scope (a delete from the `-e` script itself is undone, because Node republishes a lazy `module` accessor while bringing the ESM loader up), and the launcher publishes the real flag list in `__NUB_COMPILED_EXEC_ARGV`. `--` before the user arguments is mandatory: `node -e "…" --port 3000` otherwise dies with `node: bad option: --port`.

The compile preamble now bundles `worker-blob-url.cjs` instead of resolving it beside the emitted chunks. That was the last support file forced onto disk, and removing it drops a file from every compile build, extracting or not.

## Verification

A 21-fixture differential sweep, each fixture compiled and run under both this branch and a build of the merge-base `9b464fe00d`, on macOS arm64: 47 checks pass, 2 fail on the harness's own path filter (the `sourcemap` fixture's `at …` frames carry a per-build cache-hash directory the filter only strips at line start; both arms print the identical mapped `app.ts:1` header).

Covered: CJS/ESM feature detection against plain `node`, dynamic `import()`, `--external`, `--include`d assets, static and blob workers, TypeScript, JSON imports, `child_process.fork`, baked node flags, both source-map modes, minified and unminified, the `embed` shape, the polyfill-stripping targets 22.15/24.19/26.7, the tier matrix 18.20.8 through 26.7.0 including the zstd-absent 23.5/23.6 band, and the read-only `HOME` case. Every non-qualifying fixture declines to extraction and matches the merge-base.

Two fixes fall out of the shape: `process.argv[1]` is the artifact's own path rather than the leaked `~/.cache/nub/compile-app/<hash>/entry.mjs`, and `process.execPath === process.argv[1]` is true where it was false.

## Known, not introduced here

A blob-URL worker (`new Worker(URL.createObjectURL(blob))`) inside a compiled artifact fails with `ReferenceError: self is not defined` on both arms — an `{eval:true}` worker never receives the bundled preamble, so the worker-side scope is never installed. Pre-existing and independent of this change.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2926,6 +2926,7 @@ dependencies = [
  "aube-workspace",
  "base64",
  "blake3",
+ "brotli",
  "clap",
  "clx",
  "console",
--- a/crates/nub-cli/Cargo.toml
+++ b/crates/nub-cli/Cargo.toml
@@ -34,6 +34,7 @@ compile = [
     "dep:rolldown_utils",
     "dep:libsui",
     "dep:zstd",
+    "dep:brotli",
     "dep:oxc_allocator",
     "dep:oxc_ast",
     "dep:oxc_ast_visit",
@@ -94,6 +95,10 @@ libsui = { version = "0.16", optional = true }
 # zstd-19 compression of the embedded Node (encode side; the launcher decodes with
 # pure-Rust ruzstd).
 zstd = { version = "0.13", optional = true }
+# The inline (`no-extract`) payload's codec. Brotli rather than the zstd beside it
+# because the artifact's own JavaScript is what decompresses those bytes, and
+# `zlib.zstdDecompressSync` is absent on Node 23.5 and 23.6.
+brotli = { version = "8", optional = true }
 # `--include`/`--exclude` matching. Already in the tree at this version as a
 # nub-core dependency, so the compile feature adds no build cost — declared here
 # rather than reached through nub-core, which does not re-export it.
--- a/crates/nub-cli/src/compile/bundle.rs
+++ b/crates/nub-cli/src/compile/bundle.rs
@@ -1616,13 +1616,6 @@ impl CompilePreamble {
         let source = std::fs::read_to_string(&prelude)
             .with_context(|| format!("reading the compile prelude at {}", prelude.display()))?;
         let source = strip_native_polyfills(&source, target_node);
-        let worker_blob = runtime_dir.join("worker-blob-url.cjs");
-        let worker_blob_bytes = std::fs::read(&worker_blob).with_context(|| {
-            format!(
-                "reading compile runtime support at {}",
-                worker_blob.display()
-            )
-        })?;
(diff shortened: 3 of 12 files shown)


colinhacks:

PR #867, opened 2026-09-04
ci: publish the macOS release binary the PR round-trip already builds

colinhacks:
`macos-embed-roundtrip` already runs a full release build of `nub-cli` and the addon on every PR, then discards the binary. Upload it instead, so a PR can be downloaded and run locally. `embed-runtime` carries the addon and runtime inside the binary, so the artifact is one standalone file.

Two fixes make it fit to run:

- The job never staged `runtime/node_modules` and set `NUB_ALLOW_INCOMPLETE_RUNTIME` to compensate. It now stages the vendored helpers as the Windows twin does, and the flag is gone. The round-trip covers the transpile path, not just extract/dlopen.
- Without the typosquat corpus, `aube-resolver`'s `build.rs` ships an empty primer and only warns, leaving a binary that falls back to network packument fetches at exit 0. The corpus is generated and `AUBE_REQUIRE_PRIMER` makes that a build failure.

`actionlint` is clean. The job is `pull_request`-gated, so this PR is what proves the build passes without the flag.

Linux legs build `dev`/`fast` only, so an artifact there would be a new build, not a reused one. Windows builds a release `nub.exe` per PR and could get the same treatment; not in this PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -771,16 +771,25 @@ jobs:
           node-version: "24"
       - run: npm install --no-audit --no-fund --loglevel=error
       # Stage the vendored helpers into runtime/ so the embedded blob can transpile
-      # anything, not just the erasable-syntax fixture below. `jsbi` is transitive
-      # (via @js-temporal/polyfill) and arrives hoisted at the node_modules root.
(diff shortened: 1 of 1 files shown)


colinhacks:

PR #868, opened 2026-09-04
compile: accept a warm app cache by recorded length instead of by bytes

colinhacks:
A compiled artifact's warm start scaled with the size of its payload. `app_cache_is_ready` materialized every app file to compare it against the extracted copy, so each start paid a zstd decode of the whole app region plus a read-back of the tree — O(payload bytes) rather than O(files). Noted but left open in #862.

Launcher pre-spawn wall (`process entry` → `about to spawn node`) under `__NUB_LAUNCHER_TIMING=1`, warm, darwin-arm64:

| artifact | before | after |
| --- | --- | --- |
| hello world | 0.83 ms | 0.72 ms |
| hello world plus a 20 MB `--include` | 40.1 ms | 0.62 ms |
| a `sharp` app — 36.5 MB, ~130 extracted files | 36.9 ms | 1.77 ms |

End to end, all six binaries interleaved in one `hyperfine -N --warmup 5 --runs 100`:

| artifact | wall before → after | user+sys before → after |
| --- | --- | --- |
| hello world | 42.5 → 41.8 ms | 40.2 → 39.6 ms |
| hello world plus a 20 MB `--include` | 81.5 → 42.5 ms | 78.8 → 40.0 ms |
| a `sharp` app | 103.5 → 69.1 ms | 107.4 → 73.3 ms |

Both sides were compiled by the same `nub` against launchers differing only in this policy. Afterwards the 20 MB artifact starts in the same time as the one carrying nothing, so the per-megabyte term is gone rather than reduced. What remains scales with file COUNT: the `sharp` app's 1.77 ms is the `read_dir` walk over ~130 entries.

## What changed

Payload V3 records each file's extracted length beside its stored bytes, on the physical data record, since identical bytes cannot extract to different lengths. The warm check compares lengths.

That is the policy the embedded Node beside it already uses, and `Manifest::node_size` carries the argument in full: the extraction lives at `compile-app/<short_key(app_sha256)>`, so its content hash is already the path and re-reading the bytes established no identity the directory name did not. The tree is owner-only and non-group-writable, and a principal who can write into it can rewrite the artifact binary itself. What a length does catch is the failure the check exists for — an OS or antivirus sweep truncating or clearing a purgeable cache.

Still enforced: the exact name set, no extra file or directory, no symlink or special entry, trusted ownership and mode on every entry, the executable bits the payload marks, and the completion marker. No longer enforced: a same-length substitution by the owning uid, which was already accepted for the embedded Node.

V2 and V1 payloads record no length and keep comparing bytes. A launcher only ever reads the payload it was compiled with, so that arm is compatibility, not a live path. An inline payload never extracts and never reads any of this.

## The second commit

The Windows-only `node_spawn_error` test still passed `&Path` where #866 changed the parameter to `Option<&Path>`, so `cargo test` inside `crates/nub-launcher` did not compile on Windows and no other leg could see it. Independent of the rest of this PR.

## Verification

- A launcher test pins that a warm cache is accepted from recorded lengths alone: it hands the check a payload whose compressed bodies no decoder will accept. It fails when the length arm is removed.
- Truncating and deleting files in a real extracted tree both re-extract and then run correctly; a same-length substitution runs, as designed.
- Green: `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, `cargo build` and `cargo test` inside `crates/nub-launcher`; root `cargo fmt --check` and `cargo clippy --all-targets --all-features -- -D warnings`; `cargo test -p nub-cli --features compile --bins` (853 tests); `nub-core` compile tests; `lat check` and the wiki public lint.


colinhacks:

PR #869, opened 2026-09-04
compile: give the report tests the app_extracts field and its row

colinhacks:
`d6a114027e` ([#865](https://github.com/nubjs/nub/pull/865)) added `app_extracts` to `BuildFacts` and made the `app` row unconditional, but left the report tests behind. Two initializers omit the field, so `--all-targets` fails with `E0063` before any test runs — red on `main`, inherited by every PR on Clippy and `Test (ubuntu-latest, node 24)`.

Behind that sit three stale assertions: the `app` row is always pushed, so each expected-row vector was short one entry.

`full_facts()` takes `Some(UnsealedGraph)`, matching the `--external` entry and deferred imports it carries. The bare build takes `None`.

333 passed, 0 failed.

Refs #852

### DIFF
--- a/crates/nub-cli/src/compile/mod.rs
+++ b/crates/nub-cli/src/compile/mod.rs
@@ -4453,6 +4453,10 @@ mod tests {
                 ("sharp".to_string(), "--external"),
             ],
             deferred: 3,
+            // Consistent with the two rows above rather than picked for variety:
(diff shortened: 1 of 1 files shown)


colinhacks:
Superseded by 5649aed6d6, which landed the same fix on main.

PR #870, opened 2026-09-04
Unbreak main: the build-summary fixtures and the remote-build job script

colinhacks:
Main is red on two gates; every open PR inherits both.

`Clippy` and the compile-feature test leg fail to compile: #865 added `BuildFacts::app_extracts` and an always-shown `app` row without moving the four build-summary fixtures with it. Adding the field alone compiles and then fails, so the three row assertions gain the row too.

`Docs links & anchors` fails in `test:scripts`. One assertion pinned `cargo test` as the script's last command, which ea64995f changed. The other was reporting a real gap: the bake never warmed the compile-feature leg, so every remote test builder cold-compiled the bundler graph.

Verified: clippy green, 333 compile tests green, 24/24 remote-build tests.


colinhacks:
Superseded by 5649aed and e3a5f71, which landed the same two fixes on main directly. Closing.

The one piece not covered there is the remote-build skill's warm-command inventory, which claims to name every cargo invocation the bake warms and now omits both the compile-feature leg and the nub-launcher leg. Landing that separately.

PR #871, opened 2026-09-04
pm: drop the three nub-only root install fields

colinhacks:
`auditConfig`, `allowUnusedPatches` and `allowNonAppliedPatches` were read from the `package.json` ROOT. No package manager reads any of the three there, so they were un-namespaced names held on the strength of nobody having claimed them yet — the position `allowBuilds` was in when npm 12 shipped `allowScripts` into the same slot (#863).

Checked against each tool's source rather than assumed:

| tool | `allowUnusedPatches` | audit ignore |
| --- | --- | --- |
| npm 12 | `--allow-unused-patches`, **CLI-only by design** — ignored in `.npmrc`/env, rejected by `npm ci` | none at all |
| pnpm 11 | `pnpm.*` / workspace yaml | `auditConfig.ignoreGhsas`, same homes |
| bun | — | `bun audit --ignore <CVE>` |
| yarn berry | — | `npmAuditExcludePackages` + `--exclude` |

npm's `lib/utils/cli-only-flag.js` is explicit: "Values from `.npmrc`, env, or defaults resolve to undefined, so the flag cannot be set as project policy." A nub-native config home for it would be the design npm rejected. And `nub audit --ignore <ID>` already exists — repeatable, matching advisory numbers, GHSA ids and CVE ids together — so a config key would be a second spelling of a flag we already ship.

So the reads go and nothing replaces them. pnpm compatibility is untouched: the branded `pnpm.*` spellings, the workspace-yaml homes and the `auditConfig.*` settings read exactly as before. Only the manifest-root surface goes.

A project still carrying one of the keys gets a warning naming what replaces it — not the hard error `allowBuilds` gets, because both fail safe when unread: advisories the project had muted come back, and an unmatched patch fails an install that used to warn.

Two integration tests guard it through the real binary, offline; the aube-side unit tests structurally cannot reach the removed branch. Both were falsified — restoring either half turns them red.

`nub pm use nub` migrated pnpm's `auditConfig` to the root key. With the read gone that write would report a successful conversion while silently dropping the project's triaged advisories, and the next install would warn about the file the conversion had just written. It now reports the key as unsupported and names `nub audit --ignore`, alongside the other engine-unsupported keys. `nub pm use pnpm` still reads a root `auditConfig`: a project converted before this change carries a dead one, and pnpm's own home is the single place it becomes live again.

### DIFF
--- a/crates/nub-cli/src/cli.rs
+++ b/crates/nub-cli/src/cli.rs
@@ -10294,8 +10294,9 @@ fn run_pm_use(name: &str, spec: &str, cwd: &Path) -> Result<i32> {
 
     // `use pnpm` regenerates pnpm-workspace.yaml from the nub-mode package.json
     // homes (workspaces + catalogs, top-level overrides/patchedDependencies/
-    // allowScripts/auditConfig) — the exact reverse of `use nub`'s migration.
-    // No-op on a project that never carried them.
+    // allowScripts) — the reverse of `use nub`'s migration, plus a legacy
+    // `auditConfig` a pre-drop conversion may have left behind. No-op on a
+    // project that never carried them.
     if name == "pnpm" {
         for line in crate::pm_engine::use_nub::regenerate_workspace_yaml(&root)? {
             println!("  {line}");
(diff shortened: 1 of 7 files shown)


colinhacks:

PR #872, opened 2026-09-04
compile: let an inline payload past the synthesized manifest, and read V3 records

colinhacks:
The no-extract path has never run. Two defects, the second masked by the first.

`assemble_app` appends a root `package.json` and `classify` declined any non-chunk file, so `NonChunkFile` fired on every build. The manifest is now exempt there and dropped in `rewrite`: it serves a walk-up through the extracted tree, and an inline payload has none.

The encoder writes `[u64 data_len][u64 plain_len][data]`; the JS reader read two fields and walked 8 bytes off at the first record. It now skips `plain_len`.

Either fix alone crashes at startup — the brotli call is not wrapped.

The summary now reads `run from the executable`, and the artifact runs. 335 tests pass, up from 333.


colinhacks:

PR #873, opened 2026-09-04
compile: sign darwin-x64 output with a deterministic identifier

colinhacks:
`nub compile --platform darwin-x64` did not produce the same bytes twice. Seven of eight targets were reproducible; darwin-x64 differed by the ASCII digits of the compiling process's PID.

libsui's x86_64 signing leg stages the image at `$TMPDIR/sui_sign_<pid>` and shells out to `codesign`. Without `-i`, `codesign` derives the CodeDirectory identifier from the file's basename, so the PID landed in the signed image:

```
run 1  Identifier=sui_sign_30145-5555494407935bcff598383ca6638cdd9460491e
run 2  Identifier=sui_sign_31424-5555494407935bcff598383ca6638cdd9460491e
```

arm64 was unaffected — it is signed in-process by `apple_codesign::MachoSigner`, which writes a fixed `a.out`. That identifier is now a shared constant, passed to `codesign -i` for x86_64, matching what `prepare_node_bytes` already does with `-i node` for the embedded Node re-sign.

After, on the same fixture:

| target | run 1 | run 2 | identifier |
| --- | --- | --- | --- |
| darwin-x64 | `603006b87717db55` | `603006b87717db55` | `a.out` |
| darwin-arm64 | `5be5bae6db61ffa3` | `5be5bae6db61ffa3` | `a.out` |

`codesign --verify --strict` passes on all four artifacts, and the arm64 one runs.

## Test

The x86_64 signing moved into `codesign_adhoc`, which takes the staged path as a parameter so a test can vary it: signing one input under two names must yield one result. It fails without the fix.

libsui's unit tests ran from nowhere — it is a member of no workspace, so `cargo test -p libsui` fails at the repo root and `cd vendor/libsui && cargo test` fails too. The nub-launcher workspace patches it in as a path dependency, so `-p libsui --lib` resolves there; wired into that workspace's existing per-OS CI step, which also picks up the Mach-O section test added earlier.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -820,6 +820,16 @@ jobs:
         env:
           AUBE_REQUIRE_PRIMER: "1"
         run: cargo build -p nub-cli --release --features embed-runtime
+      # The libsui signing-determinism test is `cfg(target_vendor = "apple")`, and the
+      # per-OS launcher step in the test job reaches macOS only on push:main and
+      # nightly. This is the one macOS leg every pull request runs, so a
+      # filename-dependent signature regresses here rather than on the next main run.
+      - name: libsui units (macOS, pull requests)
+        shell: bash
+        run: |
+          set -euo pipefail
+          cd crates/nub-launcher
+          cargo test --locked -p libsui --lib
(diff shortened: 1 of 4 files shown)


colinhacks:

PR #874, opened 2026-09-04
compile: point the addon-mismatch advice at nub install's platform flags

colinhacks:
The two cross-compile refusals for a native addon named only pnpm and yarn's `supportedArchitectures`. The advice now leads with `nub install --os <os> --cpu <cpu> --libc <libc>`, which works in every project, then the setting, then the target-platform fallback. Re-captured from the built binary against a `sharp` fixture.


colinhacks:

PR #876, opened 2026-09-04
ci: give two dead rust-cache keys a writer and path-gate the docs-only builds

colinhacks:
`ci-win-embed-roundtrip` and `ci-mac-embed-roundtrip` were `save-if: false` with no writer in the repo. Both logged `No cache found.` and cold-built a release binary: 24 and 12 min (runs 33890307588, 33888414113). Trunk now writes them.

Thirteen rust-cache steps set no `save-if`, saving from PR refs. Such a cache is readable only by that PR but counts against the 10 GB budget: 10.27 GB total, 4.57 GB on one PR ref. Now main-only; verb-dispatch gains push:main so it has a writer.

macos-embed-roundtrip, wpt-worker and node-matrix built Rust on docs-only PRs — 17, 13 and 4 min, the whole wall time of #826. Now gated on `run_rust` / `paths-ignore`.


colinhacks:

PR #877, opened 2026-09-04
ci: key concurrency groups on the event name so a push cannot cancel the nightly

colinhacks:
rust-cache saves in a post-step that a cancelled job never reaches, and 34 of the last 39 push:main ci.yml runs were cancelled by the next push. With the event name in the concurrency group, the scheduled run is the one guaranteed to refresh every trunk-scoped cache key. Same change in aube-parity, lockfile-roundtrip and node-matrix, the other workflows with a Rust cache and a schedule. Follow-up to #876.


colinhacks:

PR #878, opened 2026-09-04
compile: skip eager builtin loads a payload never needs

colinhacks:
Loading `node:child_process` costs ~1.9 ms and `node:worker_threads` ~1.4 ms on every run of a compiled artifact (~2.3 ms together — they share a subgraph, so the costs are not additive). Every artifact paid both, including a payload that reaches neither.

## Measured

End-to-end on a quiet CI runner, with duplicate baselines interleaved in the same `hyperfine` invocation. The instrument prints the artifact's strip state beside the timings, so the number is bound to the artifact it came from rather than assumed.

| | before | after |
| --- | --- | --- |
| bootstrap alone, over stock Node | +3.55 ms | **+1.92 ms** |
| the artifact, over stock Node | +11.58 ms | **+8.97 ms** |

A 2.61 ms improvement against a 0.90 ms error bar, and it agrees with the 2.29 ms the per-builtin measurements predicted for both loads together. The bootstrap's own 1.63 ms drop matches the `child_process` load within the bar, so the split is consistent too, not just the total.

The dev Mac cannot measure any of this: two identical commands there differ by ~1.4 ms over 500 runs, which is larger than most of the effects involved.

## Why both loads had to move together

`child_process` was loaded eagerly in two places — the bootstrap's fork-identity fix-up, and `installCompiledChildProcess` called from the preamble. Removing either alone saves nothing, because the other still pays the load. An ablation confirms it: dropping the preamble's call by itself saves 0.16 ms, inside the noise floor.

## How one build-time decision drives both

The bootstrap is not bundled. It rides `root_support_files` and is materialised into the bundle result after the chunks exist, so it can be rewritten against the finished module graph. The preamble cannot — it is an *input* to bundling — so it reads the decision off the bootstrap's frozen record instead. No second bundle pass.

The scan covers application modules only, excluding both nub's runtime tree and the compiler's own virtual modules. That exclusion is what makes it work at all: the preamble bundles `worker-polyfill.mjs` (which declares `class Worker`) and `preload-common.cjs` (which names `child_process`), so a bare `console.log("hello")` artifact contains every marker. A scan over emitted chunks would never strip anything.

Native addon islands are verbatim package copies that never reach the bundler, so they are folded into the same decision separately.

## What counts as unreadable

Keeping a region costs startup; dropping one that was needed disables the fork policy with no error raised, so every uncertain case keeps the load. A module counts as reaching everything when it names `createRequire`, `getBuiltinModule` or `process.binding`, or when a `require(`/`import(` argument is not one complete static string.

Opening with a quote is not enough on its own: `require("child" + "_process")` starts like a literal and names no contiguous marker, so accepting it on its first byte would strip the fork patch for a payload that genuinely forks. The string must close and be followed by the call's own `)` or a second argument.

Empty parens are the deliberate exception — `import()` takes no specifier and is a syntax error, so it reaches nothing. It is worth excluding because the sequence appears in prose: one comment in zod reading "an inline `import()` of an ESM path" was enough to keep both loads for every artifact depending on zod.

`globalThis.Worker` is exempt from gating entirely. It belongs to every compiled artifact's global surface regardless of what the source spells, and computed access such as `globalThis["Work" + "er"]` must still find it. Global-surface parity is compared by own-property name, which an accessor satisfies, so the name is always present and only the load behind it defers to first read. The getter deletes itself before installing, because `installWorkerPolyfill` opens by reading `typeof globalThis.Worker` and would otherwise re-enter forever.

## Verification

Each region is independently removable and leaves the `false` initializer behind, so a removal cannot produce a syntax error or an undefined binding.

End-to-end against built artifacts, reading the bootstrap out of the extracted app tree — not `strings` on the binary, which reports "stripped" for every extracted artifact because the payload is compressed inside the executable:

| payload | `child_process` region | `Worker` region | result |
| --- | --- | --- | --- |
| names neither | dropped | dropped | exit 0 |
| imports `child_process` | kept | dropped | exit 0 |
| uses `Worker` | dropped | kept | exit 0 |
| reaches the name only by computed access | dropped | dropped, accessor present | exit 0 |
| computed specifier | kept | kept | exit 0 |

Unit tests cover the stripper, the scan, and the specifier reader, each with a positive control — the negative assertions would otherwise pass equally against a scan that records nothing at all.

## Also here

`.worktreeinclude` now brings `runtime/node_modules` into a new worktree. Without it 49 compile bundle tests fail there with `Cannot find module 'urlpattern-polyfill/urlpattern'`, which reads as a code defect rather than a missing dependency. CI never sees it, because a Node 26 target strips those polyfill regions before anything resolves them.

## Not addressed

Emptying the preamble entirely saves 6.20 ms, while its three install calls together account for only ~1.9 ms. The rest is not reachable by call-gating, and a follow-up measurement has since located it more precisely than this branch could.

The preamble's whole statically imported graph evaluates in ~1.84 ms of child CPU, so it is not where the remainder sits. Two things carry it, and they split by platform. On Linux the payload term is ~7.1 ms, of which roughly 5.2 ms is beyond the preamble graph — the emitted chunks, the entry, and the loader work for four files rather than two. On macOS the launcher process itself costs several milliseconds that Linux does not pay.

A second bundle pass over the preamble's unused imports was the obvious next move and it is worth recording that it gains nothing: five readings across both platforms, every one inside its error bar. `worker-polyfill.mjs` and `worker-blob-url.cjs` are pure declarations, and dropping them changes `process.moduleLoadList` by zero entries. Evaluation cost tracks top-level work, not bytes — `polyfills.cjs` is 86 KB and evaluates in 0.31 ms because its top level is four statements.

Single-file emission is the largest remaining item.

### DIFF
--- a/.worktreeinclude
+++ b/.worktreeinclude
@@ -51,3 +51,15 @@ symlink .fray
 # clobbering it for every sibling. A missing source is skipped, so a tree that
 # has never built the addon still creates worktrees fine.
 copy runtime/addons/nub-native.node
+
+# The vendored polyfill packages the compile preamble imports (urlpattern,
+# Temporal, Float16Array). Without them 49 compile bundle tests fail in a fresh
+# worktree with `Cannot find module 'urlpattern-polyfill/urlpattern'` — a
+# confusing failure that looks like a code defect. CI never hits it because a
+# Node 26 target strips those polyfill regions before they are resolved.
+# COPIED, not symlinked, and 5 MB: a symlink resolves into the SHARED tree, which
+# puts the packages outside the worktree's own runtime directory. The compile
+# preamble's builtin-usage scan classifies anything outside that directory as
+# APPLICATION code, so a symlink would make it read nub's own polyfills as app
+# usage and keep eager builtin loads no payload asked for.
+copy runtime/node_modules
--- a/crates/nub-cli/src/compile/bundle.rs
+++ b/crates/nub-cli/src/compile/bundle.rs
@@ -1949,6 +1949,44 @@ fn strip_regions(source: &str, prefix: &str, drop: &[&str]) -> String {
     out
 }
 
+/// Can this module reach a builtin by a name the substring scan cannot read?
+///
+/// Compiled CommonJS deliberately PRESERVES a non-static `require(expr)` rather
+/// than failing the build (see [`Requires::classify_require`], which declines to
+/// flag them because every real instance measured was a guarded optional loader).
+/// That is correct for the bundler and fatal for a literal scan:
+/// `require(["child", "process"].join("_")).fork(...)` reaches the builtin naming
+/// neither marker, and stripping the fork identity patch there would let `fork()`
+/// silently re-run the artifact instead of real Node.
+///
+/// So a module that can compute a specifier counts as using EVERYTHING. The same
+/// applies to the indirect accessors, which take a specifier this scan never sees.
+/// A template literal counts as computed even when its body is constant — the
+/// distinction is not worth reading, and the cheap answer is the safe one.
+fn has_computed_module_access(code: &str) -> bool {
+    for accessor in ["createRequire", "getBuiltinModule", "process.binding"] {
+        if code.contains(accessor) {
+            return true;
+        }
+    }
+    for call in ["require(", "import("] {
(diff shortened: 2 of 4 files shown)


colinhacks:

PR #718, opened 2026-08-11
Accept duplicate package manifest fields

colinhacks:
## Summary

- Accept duplicate object keys in published `package.json` files with last-value semantics.
- Keep the fast typed parser and preserve the original diagnostic for manifests that remain invalid.
- Cover duplicate scalar, script, and dependency fields.

## Verification

- `XDG_CONFIG_HOME=<empty> make verify`
- `cargo test -p aube-manifest`
- `cargo clippy -p aube-manifest --all-targets -- -D warnings`
- `lzma-native@0.0.5` now reaches its `node-gyp` install step instead of failing with `ERR_NUB_MANIFEST_PARSE`.

PR #879, opened 2026-09-04
aube: sync upstream v2.2.9 and move nub-cli to usage-rs

colinhacks:
Vendors aube v2.2.9 (upstream `aubepkg/aube`, 237 commits since v1.35.0) and moves `nub-cli` from clap to `usage-rs`, the parser upstream switched to in v2.0.1, so the binary links one parser and the engine verbs compose aube's `usage_rs::Args` types directly.

**Vendored engine.** `vendor/aube/UPSTREAM` now records `79906840d9` / `v2.2.9`. The nub delta (workspace markers, `nub.lock` naming, `.store` virtual-store subdir, branded-config gating, `EngineContext`, env overlay, cold-path settings) is intact. Upstream landed its own copy of `EngineContext` with a `bundled_package_extensions` field, kept as ours plus that field. The project registry converges on upstream's JSON registry under the global virtual store: every project registers there on every linker mode, and nub's `store prune` sweep of the extracted-tree tier now reads that registry instead of the removed store-side one. Upstream's new CAS prune planner stats only unreferenced files (it stat'ed every store file; 98s → on par with the previous 17s on a 900k-file store). Upstream's new PATH composer for dlx, exec and run children keeps the embedder-provisioned node dir as its fallback, which is what lets a fetched bin's `#!/usr/bin/env node` re-enter nub; without it dlx-fetched bins ran unaugmented.

**Kept on nub's side of the merge**, for review: hoisted version-preference policy (upstream #1296 not adopted); global install layout (`~/.local/bin`, `<data>/global`); the peer-context bound and the `latest` → `*` widening are refused; `EmbedderInstallOverrides.cache_dir` outranks `NUB_CACHE_DIR`; default build trust stays on the `defaultTrust` floor rather than upstream #1199's seeded allowlist (a built-in `Allow` answered ahead of the floor and skipped its provenance, advisory and cooling gates and the disclosure line), with pnpm's list folded into the floor's candidates instead, which at this snapshot adds exactly one name to Bun's, `@parcel/watcher`; patch drift on a pnpm lockfile is the hash-against-hash check (upstream's new path-map check saw every recorded pnpm patch as missing, since nub's reader keeps the content hash and no path), and a hashless `{ path }` entry is judged on its path.

**Adopted from upstream:** the lockfile as the trust boundary (#1418, #1403): `trustPolicy=no-downgrade` runs when a version is newly resolved, and an install that reuses an undrifted lockfile trusts the recorded pin without re-fetching its publishing evidence. Nub's re-validation pass, its on-disk cache and the `warm_trust_revalidate` embedder gate are removed (about 500 lines); the docs paragraph and the two nub-side install tests follow. Also adopted: side-effects marker v2 and the `/v1` global-virtual-store layout (warm state re-extracts once), the age-gate blocked-update warning in `outdated` and `update -i`, `update --ignore-scripts`, `store prune --dry-run`, and the built-in `disableGlobalVirtualStoreForPackages` list, which now keeps the global virtual store on for nuxt and parcel (upstream #1312) and turns it off for expo, react-native and metro (upstream #1297).

**`store prune` after the `/v1` layout.** Entries earlier versions wrote at the store root, and the project records beside them, are neither read nor pruned by upstream. Nub's prune now derives that root from the versioned path, keeps an entry a still-existing project links into, holds an unreferenced one for the same 30-day window as the extracted-tree cache (against the root's own `.gc-state`), and removes the old bookkeeping once every entry is gone. Existing projects keep linking into the old entries until their next install. Behavior change to note on the versioned store itself: with upstream's registry, an entry no registered project links to is removed on the first prune; the 30-day hold now applies to the extracted-tree cache and the previous layout only. The virtual-store docs page is rewritten to match, with its sample output re-captured.

**CLI.** Every clap surface in `nub-cli` is on `usage-rs` 6: the root tree, the shared flag groups, and the engine-verb bridge, which now stamps one static `usage_rs::Cli` root per verb. User-visible help changes are usage's vocabulary (`Flags:` for `Options:`, `[FLAGS]`, `…`, `<SUBCOMMAND>`), `-V` advertised beside `-v`, and clap's `tip:` line on an unknown-verb error is gone. Usage errors still exit 2. Engine doc hosts (`aube.sh`, `github.com/aubepkg/aube`) join the brand rewrite.

Gates: remote clippy + brand lints, `cargo test` with the `compile` leg, aube's own `cargo test --workspace`, the nub-adapted bats suites, the lockfile conformance matrix, and a differential sweep of the CLI surface and install/workspace sequences against the merge-base build. Ad-hoc on a builder: `@parcel/watcher` builds through the floor with the disclosure line; a fabricated pre-`/v1` layout keeps the linked entry, removes the expired orphan and leaves `v1` intact, then holds the rest for 30 days; an exact pin that lost publishing trust is still refused at resolve time (`ERR_NUB_TRUST_DOWNGRADE`), and a frozen install of an undrifted lockfile succeeds without it. Thirteen bats tests new in this range hit documented nub divergences (the `nub.lock` name, `.store`, rebranded codes, the branded-config boundary, the hidden GVS toggle flags, nub's install summary) and are on the skip lists with their reasons; four upstream engine tests are adapted to nub policy (a seeded index needs an integrity key; a per-project install still registers the project; a default hoist vetoes the shared store).

Open PRs touching `vendor/aube` will need a rebase: #758 (likely superseded by upstream #1326/#1369), #727, #726, #718, #583, #577.


colinhacks:
Fixed legacy store pruning in 125de28b60. Relative links, aliased store paths, and custom modules directories now retain live packages. Both defects also reproduce on the merge-base. Added regression coverage; remote workspace tests, Clippy, and end-to-end pruning fixtures pass.

colinhacks:
Fixed the remaining traversal case in 11d9b6f146. Scan boundaries now compare physical paths, including `..` after a symlink. The expanded regression test fails before the fix and passes after it; remote workspace tests, Clippy, and pruning fixtures pass.

colinhacks:

PR #880, opened 2026-09-04
compile: bracket the summary block with an intro and a success line

colinhacks:
A compile printed nothing that survived in the scrollback before its summary block: the only pre-block line was the live spinner, which erases itself and is off entirely when stderr is not a terminal. And a build that worked ended on its own last fact rather than on anything saying it succeeded.

Two lines now bracket the block, in the vocabulary the CLI already spends on these meanings:

```
nub 0.8.3 · compiling cli.ts → cli

    output  cli  29.4 MB  (node 28.2 MB · app 57 KB · launcher 1.2 MB)
   runtime  Node 26.7.0, embedded  (package.json#engines.node)
  platform  darwin-arm64
       app  extracted on first run  it extracts its embedded Node anyway

✓ compiled cli in 3.0s
```

The success line is the engine's install sign-off, and it carries the elapsed time, so the `elapsed` row is gone with it. The duration picks up the engine's three bands (`240ms` / `4.0s` / `3m12s`) in place of a flat `{:.1}s`, which rendered a target-Node download as `92.4s`. Both lines print with and without color; the spinner stays transient.

### DIFF
--- a/crates/nub-cli/src/compile/mod.rs
+++ b/crates/nub-cli/src/compile/mod.rs
@@ -110,8 +110,8 @@ type AppFiles = Vec<AppFile<Vec<u8>>>;
 
 pub fn run(mut opts: CompileOptions) -> Result<i32> {
     // Started before anything that can touch the network or the disk, so the
-    // `elapsed` row measures the wait the user actually sat through rather than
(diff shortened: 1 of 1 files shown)


colinhacks:

PR #881, opened 2026-09-04
release: build the canary nightly instead of on every push to main

colinhacks:
Every canary publish adds a permanent version entry to the packument of all 19 published packages, which npm serves on every install. The per-push channel added 82 versions in 42 days — 45% of `@nubjs/nub`'s packument, and 88% of `@nubjs/loader`'s.

Canary now builds on a 09:20 UTC schedule rather than on every push to main, plus a `canary` dispatch input for an on-demand run. The gate baselines on the commit the rolling `canary` tag points at, so a nightly is skipped when main has not moved as well as when the span is inert. Only 23 of those 42 days carried a code change, so together that cuts the publish rate about 3.6x.

Same `canary` dist-tag, same rolling release, same `nub upgrade --canary` — only the cadence changes.

Two things a schedule trigger breaks that are fixed here: the pre-publish smoke gate was gated on the push event, so it would have cascade-skipped the whole canary publish chain, and `linux_only` on a canary dispatch would have published a partial platform set.


colinhacks:

PR #882, opened 2026-09-04
compile: split the output size by measured components

colinhacks:
The output row computed `launcher` as `size − node − app`, so the remainder also carried the code signature, the payload region’s 64 KiB alignment, the container header and manifest, and the Node LICENSE. A 29.4 MB embed build reported an 851 KB launcher template as `launcher 1.2 MB`.

Every component is measured now; only `overhead` is a remainder.

```
29.4 MB  (node 28.2 MB · app 57 KB · launcher 851 KB · signature 228 KB · overhead 66 KB)
990 KB   (app 103 KB · launcher 851 KB · signature 8 KB · overhead 28 KB)
```

Both sum to the file size exactly, checked against `otool -l`. The signature holds one SHA-256 per 4 KiB page, so it scales with the artifact; ELF and PE omit it.

### DIFF
--- a/crates/nub-cli/src/compile/inject.rs
+++ b/crates/nub-cli/src/compile/inject.rs
@@ -15,7 +15,7 @@
 //! not a re-read of whatever the writer happened to emit.
 
 use std::fs;
-use std::io::Write;
+use std::io::{Read, Write};
 use std::path::Path;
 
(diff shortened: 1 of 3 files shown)


colinhacks:

PR #883, opened 2026-09-04
compile: drop `overhead` from the output size row

colinhacks:
The row reached 106 columns on a real embed build, so it wrapped at 80 — and it is the line every recorded transcript quotes.

The `overhead` part is gone. It named the container header, the manifest, the per-file framing and the 64 KiB alignment padding: tens of kilobytes nobody acts on. The other four parts are unchanged.

```
output  cli  29.4 MB  (node 28.2 MB · app 57 KB · launcher 851 KB · signature 228 KB)
```

Also drops the live line's redraw interval from clx's 200 ms default to 80 ms.

Refs #882


colinhacks:

PR #884, opened 2026-09-05
compile: cut the artifact's startup work before it reaches the app

colinhacks:
Five changes to what a compiled artifact does before it reaches application code. Each is priced separately on the runners against interleaved controls, because the dev Mac's noise floor is larger than most of the effects here.

## What each one is worth

Linux and macOS, three instruments — `hyperfine` minimums, child CPU time, and in-process `nodeTiming` marks. A blank means the effect did not resolve above that instrument's spread on that platform.

| change | Linux wall | Linux CPU | in-process | macOS |
| the bootstrap preload dropped when the payload needs no eager builtin | −0.3 to −0.5 | ~0 | −0.8 | ~0 |
| single-file emission | −0.6 to −2.0 | −0.8 to −1.8 | −0.6 to −2.1 | −1.5 CPU, −2.6 in-process |
| the chunk shaped for a complete compile cache | −0.7 to −1.2 | −0.75 to −1.2 | −1.05 to −1.5 | −0.6 CPU |
| the bundled polyfills installed as lazy globals, Node 22 target | **−25.1** | −25.2 | −24.4 | −17.8 to −20.1 |
| the launcher without Security and CoreFoundation (macOS only) | — | — | — | −1.75 CPU per artifact |

A hello-world artifact goes from +6.3–6.9 ms over plain `node` on Linux to **+4.4–5.5**. A Node 22 target goes from **+35 ms to +9.9**.

## Node's compile cache was half-empty by construction

`CompileCacheHandler::MaybeSave` serializes a module's code cache immediately after compilation, so it holds bytecode only for the functions V8 chose to compile at parse time. Every function compiled lazily — which is most of them — is recompiled from source on every warm start, cache or no cache. Marked in-process on the hello chunk, the preamble cost 1.35 ms, and **0.77 ms of that, 55%, was recompilation of code the cache already had the source for**.

The fix is a shape rather than a flag. V8 compiles a parenthesized function literal eagerly, and that eagerness reaches parenthesized literals nested inside it. So on a target with a compile cache (22.1 and later) the chain from the chunk's top level down to each runtime helper is emitted as parenthesized function expressions: runtime declarations are hoisted to `var f = (function f(){})`, each forwarder in a single-chunk bundle is bound the same way, and a wrapper's callback is parenthesized. The parentheses go on after Rolldown's minifier has run, and the map is recomposed with `collapse_sourcemaps`.

The cache entry for hello grows from 7.5 KB to 79.5 KB. With no cache at all — the inline `--smol` shape, or a first run — the two trees measure 3.18 against 3.14 ms, so nothing is gated on the cache existing.

**The shape switches on at Node 24, not at 22.1 where the cache first appears.** Those are different facts and only the second is the gate. Largest cache entry for the same hello chunk, shaped against unshaped:

| Node | unshaped | shaped |
| --- | --- | --- |
| 22.15.0 | 49,948 | **46,196** |
| 22.23.2 | 49,964 | **46,164** |
| 24.19.0 | 48,068 | 311,564 |
| 24.20.0 | 48,052 | 311,548 |
| 26.5.0 – 26.8.1 | 7,500 | 79,772 |

On 22.x the shaped tree caches *less*, so the transform works against its own premise there and is off for the whole 22 and 23 band. The small 26.x baseline is V8 being much lazier by default on that line, which is why the same change is worth 10.6x there and 6.5x on 24.

The test's own assertion had the same fault, requiring the unshaped control to stay under 20,000 bytes — true on 26, false on 24. It asserts the ratio now, which holds wherever the shape is on and which a broken transform destroys by leaving the two shapes equal. Swept across 20.19, 22.15, 22.23, 24.19, 24.20, 26.5, 26.7 and 26.8.1.

An application's own module bodies are eager too, so this scales with the application rather than being a fixed win. Inner helpers are deliberately left lazy, so a large program's cache does not fill with code that never runs.

## Verification

- 353 `compile::` unit tests and the compile integration tests pass. The compile feature is not on by default.
- `eager_startup_fills_the_compile_cache_on_the_first_run` is a positive control: it runs the emitted files under the PATH Node with the option on and off and compares cache sizes, so the assertion cannot pass against a build that emits nothing.
- `tests/compile-augmentation` parity is unchanged on 26.7.0 and 22.23.2, with the `a-global-parity` digests identical either side. The one failure on 26.x, `a-sourcemap`, predates this branch.
- The measurement workflow is green on both runners with every arm reported.

## Rejected, with the measurement

- `--no-lazy` on the command line: V8's flag hash makes Node reject its own embedded caches. +4 ms.
- Turning `--lazy` back on at runtime through `v8.setFlagsFromString`: a complete cache, but +7.0 ms CPU, because the flag is live during Node's own bootstrap and loading `node:v8` costs 0.9 ms by itself.
- V8's compile-hints magic comment: Node does not opt into it.
- Converting Rolldown's arrow callbacks to function expressions: no change to either cache size or time, so the code was removed again.

## Also here

`-Wl,-dead_strip_dylibs` on both Apple targets. `rustls-platform-verifier` emits `link-lib=framework=Security` and `framework=CoreFoundation` unconditionally, and fat LTO removes every function that used them — `nm -u` shows no `_Sec*` or `_CF*` imports — but a load command is a link-time fact rather than a symbol, so dyld loaded Security, CoreFoundation and their private-framework tail on every exec and ran the initializers. The flag keeps only dylibs an actual reference reaches. Initializer count goes from 404 to 2; 365 of the 404 were Network.framework's.

**The effect is release-only, and that is worth knowing before someone reports it as broken.** A `release` launcher links `libSystem` alone. A `debug` launcher still lists both frameworks, even though the flag reaches rustc and no undefined symbol refers to either — fat LTO has to remove the code before the linker will drop the load command. The release template is what `nub compile` embeds, so the shipped artifact gets the benefit; a debug build checked with `otool -L` will not show it.

The measurement scripts under `tests/preamble-eval/` and `tests/startup-attack/` are the instruments every number here came from, so they stay. Their branch-scoped workflows do not: they key on probe branches this merge deletes, and on main they would only answer `workflow_dispatch`. A future probe adds its own workflow on its own branch.

## Not addressed

Version-gating the eleven polyfill installers is now worth about 0.05 ms, because their compilation is cached. Two in-process segments remain unexplained, together about 0.36 ms. A Node 22 target still costs +9.9 ms over plain Node 22, roughly double the Node 26 figure.

Separately, and pre-existing: V8 serializes a script's `//# sourceMappingURL` into the code cache, so an artifact built with `--sourcemap inline` carries its whole map in every cache entry. The default artifact has no map.


colinhacks:

PR #887, opened 2026-09-05
compile: build the default embed shape as a Node single-executable application

colinhacks:
Rebuilds the default `nub compile` shape as a [Node single-executable application](https://nodejs.org/api/single-executable-applications.html). An embed artifact was a Rust launcher with a compressed payload appended, which extracted a Node and the bundled chunks to a cache directory on first run and then spawned that Node against them. It is now one file that *is* Node, with the app in the blob Node's own embedder loader reads: nothing is extracted, and no second process is created.

`--smol` is unchanged and still uses the launcher — a single-executable application is a Node, and the sub-1 MB shape embeds none. The payload format, the bundler output and the compile bootstrap are shared; only the container differs.

## Start time

Node 26.7.0, linux-x64, a dedicated 4-vCPU box, `hyperfine -N --warmup 60 --runs 500`. Every row appears twice in the same invocation and both replicates are shown, because a slow drift on the machine otherwise lands entirely on whichever rows ran first. Medians, relative to `node hello.mjs`.

| row | Δ median (ms) |
| --- | --- |
| `node hello.mjs` | 0.00 / 0.06 / 0.12 |
| a plain Node single-executable application of the same file | 0.07 / 0.25 |
| the bundled app under plain `node` — what the launcher spawns, minus the launcher | 3.81 / 3.68 |
| **the artifact, single-executable** | **3.65 / 3.72** |
| the artifact, launcher | 3.39 / 3.35 |

The container is free either way: the app's own cost is ~3.7 ms and both shapes land on it. The single-executable one sits 0.3 ms behind the launcher, which is inside the spread between the two shapes' own replicates.

Getting there needed a different loader from the inline (`no-extract`) shape. That shape has to make each chunk a `data:` URL, because it has nothing on disk and no hook API on its floor; inside a blob the same choice costs 8.3 ms on a 60 KB chunk, in a base64 encode, a URL parse and a base64 decode that no cache covers. The blob's chunks are served through `module.registerHooks` at `file:` URLs instead, with `sea.getRawAsset` as the source and Node's on-disk compile cache on. Serving real URLs also means the chunks travel verbatim: no cross-chunk specifier substitution and no eval-global cleanup, both of which exist only to work around a `data:` URL having no identity and no base.

There is ~0.8 ms of headroom left in the ELF writer, and it is measured rather than estimated: the same blob and the same Node injected by `postject` runs 0.6–0.9 ms faster, reproduced four times. It is not the main source, the assets, the strip, huge pages, `ld.so`, syscall counts or page faults — each was isolated and each is free. Naming it needs hardware performance counters, which the measurement box does not expose.

## Size

A hello-world artifact, all eight targets, Node 26.7.0:

| target | size | | target | size |
| win32-arm64 | 90.1 MB | | linux-arm64-musl | 112.1 MB |
| win32-x64 | 103.4 MB | | linux-arm64 | 113.3 MB |
| darwin-arm64 | 107.8 MB | | linux-x64 | 115.3 MB |
| darwin-x64 | 110.6 MB | | linux-x64-musl | 128.3 MB |

That is 3.7× the launcher artifact on disk for linux-x64 (31.4 MB), and the strip is still worth 32.6 MB against a plain `node --build-sea` output of the same app. Compressed for distribution, a darwin-arm64 artifact is 28.8 MB at zstd-19 and 39.2 MB at gzip -9.

`--icu` no longer changes the size on disk: the trim rewrites ICU's package in place without shortening it, and there is no compression pass left to turn that into a smaller file. It still pays on the wire, where the zero-filled remainder compresses away — 28.8 MB → 24.2 MB at zstd-19, 39.2 MB → 33.0 MB at gzip -9.

## What a user can notice

- **An artifact's Node flags are fixed at build time**, where the launcher recomputed them per run. `NODE_OPTIONS` still reaches the process, since the blob's `execArgvExtension` is Node's own `env` default, but a `--no-…` in it no longer subtracts a flag nub injects, because argv beats the environment.
- **A payload that names `child_process` keeps the launcher.** `fork()` in a single-executable artifact spawned the artifact, which ignores `argv[1]` and re-ran the application, which forked again; a two-line fixture printed its first line until it was killed. The cause is the bootstrap's fork identity fix-up, which points a fork at the Node the artifact runs on — and inside a single-executable application there is no such Node, because the artifact is Node. This joins the eligibility rules that already refuse `--external` packages, retained computed `import()`, `--include`d files, native addons, traced worker chunks, `node:cluster` and linked source maps.
- **A target Node below the `execArgv` backport bands keeps the launcher too**, since an artifact that cannot carry nub's flags would diverge from `nub <file>` on the first program that uses one. Those bands are not one floor: the field reached 22.20, 24.7 and 25.0 as three backports, so the whole 23.x line is without it.

## How the blob is written

nub serializes Node's blob format and writes the container itself rather than calling `node --experimental-sea-config`, which would need a host Node of the target's exact version for every cross-build. Two header fields are version-gated: the `mainFormat` byte that arrives in 25.7, and `execArgv`.

The vendored libsui gains `write_section_in_segment` and `append_note`, so the blob lands under the segment, section and note names Node's runtime has compiled into it, and `from_keeping_signature`, so a Mach-O template keeps the `LC_CODE_SIGNATURE` it arrived with — stripping it first makes the later `codesign` fail strict validation. `codesign_adhoc` passes `--force` for the same reason: without it `codesign` refuses an already-signed input, and the warn-and-continue path left a stale signature on a binary that runs but reports itself modified. Writing into the existing header slack is also what gives darwin-x64 an artifact at all, since LIEF-based injection shifts the `__thread_*` sections and the result dies at exec.

Every write is read back before publication: the fuse byte must be set, and a blob carrying Node's magic must sit where that platform's Node will look for it. That check has already caught a real failure — it is what refuses the stripped-signature Mach-O rather than shipping it.

## Verification

`tests/compile-sea-probe/` is a branch-scoped probe over five runners. It builds nub on each, compiles a fixture, and asserts the artifact is really a single-executable application, prints exactly what `nub app.mjs` prints on the same machine, and extracts nothing into a fresh cache root. Native darwin-x64, linux-x64 and linux-arm64 pass; the musl and Windows cells are in the run on this branch.

Locally on darwin-arm64: `--smol` unchanged at 990 KB and running; `--icon`, `--metadata` and `--hide-console` verified on a cross-built win32-x64 artifact; a below-floor target verified to fall back; a forking fixture verified to fall back and terminate. `wiki/commands/compile-architecture.md` is updated — six of its sections described a container the default no longer uses.


colinhacks:
Fixed in c0f6518756. Fixtures in `tests/compile-sea-probe/semantics.mjs`, confirmed red against the unfixed loader.

Entry completion: the import is now observed. Unsettled top-level await gives 13; a rejection is rethrown, since a failed ESM entry is an uncaught exception, not an unhandled rejection. The inline loader had the same defect and the same fix.

Web storage: reproduced on Node 22.20. The run reads the getter instead of the build predicting it.

Relative resolution is scoped to a parent under the virtual root. The route named is unreachable: a computed `require` counts as reaching `child_process`, which refuses this container.

colinhacks:
Confirmed and fixed in 9513212b43. `process.emitWarning` queues its write behind a tick, so one emitted from an `exit` listener is composed and dropped. The warning now goes on `beforeExit`, the exit code stays on `exit`. Both no-extract loaders had it.

Kept as a warning rather than a raw stderr write, so `--no-warnings` and `--trace-warnings` still govern it — Node's own copy is gated the same way.

`semantics.mjs` now asserts stderr by marker on both sides, plus a suppressed case: under `--no-warnings` the artifact exits 13 and says nothing, as `nub` does. Red-confirmed against the unfixed loader.

colinhacks:
Right — an enumeration is what failed twice, so the filter is inverted in 8af740f2f5. It now lists only what cannot reach a compiled artifact (`wiki/`, `site/`, `docs/`, `**/*.md`); every other push runs the probe, `compile-preamble.mjs`, `polyfills.cjs` and `flags.rs` included.

60fb41cbbd adds the concurrency group that makes that affordable: a run is half an hour across five runners, two of them scarce, so superseded runs now cancel instead of stacking.

colinhacks:
Reproduced and fixed in c08f4d9915. The loader registers its `beforeExit` listener before importing the entry, so it ran ahead of the application's.

The first `beforeExit` now only schedules, buying the loop another turn for the remaining listeners and their promise jobs. Still pending when the loop empties again means diagnosed. Both no-extract loaders had it.

`semantics.mjs` gains that case, red-confirmed, green on both shapes.

colinhacks:
Right — a bounded guard is wrong at the next cycle. Fixed properly in 3dbe7aecd1: both the exit code and the diagnostic now come from `exit`, where Node reaches its own verdict.

That means writing to fd 2 rather than emitting, since `emitWarning` never gets a turn from an `exit` listener. Node writes this line synchronously too and gates it on `--warnings`; the gate is readable without parsing a flag, because Node attaches its own `warning` listener during pre-execution exactly when that option is on and `NODE_NO_WARNINGS` is unset.

`semantics.mjs` gains the cycle-two case and is 8 cases, red against the previous loader.

colinhacks:
Fixed in the push above. `process.noProcessWarnings` is a read-only alias of the parsed `--no-warnings` and is what Node gates this diagnostic on, so both loaders read it directly. Present since Node 10, below either shape's floor.

Confirmed the divergence first: with both spellings set the control exits 13 silently and the proxy printed. `semantics.mjs` gains that case, red against the proxy, and `--smol` agrees — silent with both set, printing under `NODE_NO_WARNINGS=1` alone.

colinhacks:

PR #888, opened 2026-09-05
nub-phantom: fetch metadata from the abbreviated packument, not /latest

colinhacks:
A wide scan stalls on `registry.npmjs.org/<pkg>/latest`: it returns 429 and no backoff clears it, so `scan --top N` never finishes.

Five consecutive probes per endpoint, same host: `/express/latest` 429 × 5, `/express` with `Accept: application/vnd.npm.install-v1+json` 200 × 5. Same eight packages — old path incomplete after 5 min, new path 1.8 s with zero failures. A top-10000 scan now runs end to end.

The abbreviated document carries both fields the fetcher needs (`dist-tags.latest`, that version's `dist.tarball`). The manifest is read from the tarball, so the fields it drops cost nothing.

Fmt and clippy green on the crate.

### DIFF
--- a/crates/nub-phantom/src/fetch.rs
+++ b/crates/nub-phantom/src/fetch.rs
@@ -11,6 +11,8 @@ use flate2::read::GzDecoder;
 use reqwest::blocking::Client;
 
 const REGISTRY: &str = "https://registry.npmjs.org";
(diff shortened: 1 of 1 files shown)


colinhacks:

PR #889, opened 2026-09-05
compile: keep an unconditional module.exports assignment CommonJS

colinhacks:
A compiled artifact that carries a native addon, in a project whose `package.json` declares `"type": "module"`, fails at startup:

```
ReferenceError: Cannot determine intended module format because both require() and top-level await are present.
  code: 'ERR_AMBIGUOUS_MODULE_SYNTAX'
```

Nub generates the shim that loads an embedded addon (`native::addon_module`), and that shim assigns `module.exports`. It is served under the `.node` id, whose owning manifest is the *application's* — so a project declaring `"type": "module"` made Nub's own CommonJS module look like an ESM dependency to `preserve_dependency_esm_classification`, which appended `export {};` to keep the ESM reading. The marker turned off Rolldown's CommonJS wrapper, `module` was left unbound, and Node reports that `ReferenceError` under the ambiguous-syntax code.

The marker now skips a module that assigns `module.exports` or `exports.<name>` as an unconditional top-level statement. Such a module is CommonJS by construction: run as ESM the statement throws the first time the module executes, so no working package ships one. The shape the marker exists for is the guarded probe — `try { module.exports = … } catch {}` — which still gets it. `compile::closure` emits the same unconditional shape for a CommonJS chunk entry, so it is covered by the same rule.

Caught by the nightly release run, which failed the pre-publish smoke gate on all eight platforms.

## Verification

Three builds against the smoke gate's own fixture, on darwin-arm64 with a Node 22.23.2 target:

| build | result |
| --- | --- |
| 0.8.3 | runs, prints the expected line |
| `main` at `c005160854` | `ERR_AMBIGUOUS_MODULE_SYNTAX` |
| `main`, same fixture, manifest without `"type": "module"` | runs |

The third row is the control: one field differs, and it is the guard clause the marker turns on.

The new unit test calls the real shim generator rather than a copy of its output, so the two cannot drift apart. No pull-request gate covered this shape — `compile-native.yml`'s fixture has no `"type": "module"`, and its addon belongs to `sharp` rather than to the application, so the package-defaults-to-CommonJS guard short-circuits there.

Refs #884

### DIFF
--- a/crates/nub-cli/src/compile/bundle.rs
+++ b/crates/nub-cli/src/compile/bundle.rs
@@ -3114,32 +3114,50 @@ fn preserve_dependency_esm_classification(path: &str, source: &str) -> Option<St
     let allocator = Allocator::default();
     let source_type = SourceType::from_path(path).unwrap_or_else(|_| SourceType::mjs());
     let parsed = Parser::new(&allocator, source, source_type).parse();
-    if parsed.panicked
-        || has_esm_syntax(&parsed.program)
-        || assigns_commonjs_exports_at_top_level(&parsed.program)
-    {
+    if parsed.panicked || has_esm_syntax(&parsed.program) {
         return None;
     }
     let semantic = oxc_semantic::SemanticBuilder::new()
         .build(&parsed.program)
         .semantic;
-    let misread = semantic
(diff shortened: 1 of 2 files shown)


colinhacks:

PR #890, opened 2026-09-05
compile: run a single-executable artifact before publishing it, and narrow the fork decline

colinhacks:
Two correctness gates for the single-executable shape.

The build now runs the artifact on a host target. Every other check reads nub's own bytes back, so a drift in Node's blob layout would be invisible — the producer and the checker would share the mistake. The artifact reports its entry name and byte length, so the reply covers the asset table and not only the blob's presence.

The fork decline now reads what a chunk resolves rather than what it spells. It was a substring search for `child_process` or `cluster`, so a payload importing `clusterApiUrl` was refused the container. It now uses the AST scan the cluster decline already uses.


colinhacks:
Both real, both fixed in 216e947976.

The probe now removes `NODE_OPTIONS` and `NODE_REPL_EXTERNAL_MODULE`, matching `node_runs`. Verified: a build under `NODE_OPTIONS=--totally-bogus-flag` now succeeds.

The alias fix alone did not hold, and a compiled fixture is what said so. The emitted chunk opens `const require = __nubCjsRequire`, which ends in a capital R, so the suffix test was false for the compiler's own helper — and minification had already renamed the binding to one letter. Every CommonJS payload's require calls were invisible here, including to the cluster decline. A CommonJS `node:cluster` payload took the no-extract container.

The name test is now a case-insensitive substring. Both fixtures decline correctly; 368 tests pass.

colinhacks:
The remaining required outcome cannot be met by this scan, and here is the counterexample:

```js
globalThis.__holder = { r: require };
const cp = globalThis.__holder.r("node:child_process");
```

That compiles to a single-executable artifact today. The callee is a member expression; the alias collector records identifier and member initializers, and an object literal is neither. Following it takes interprocedural dataflow over the bundle.

So 1c2e7be0b1 fixes what happens when the scan is wrong instead. The loader replaces `child_process.fork` with one that throws, installed before the entry so cluster's destructured binding gets it too. A miss is now an error at the call, not an unbounded re-entry.

`tests/compile-sea-probe/fork-guard.mjs` compiles that exact fixture and asserts the refusal and that no second generation starts. Red-confirmed: with the guard removed it prints FORKED three times.

colinhacks:
Both addressed in 0a572ca83f.

The skip is now a failure. A target Node below the container's version bands still returns early — that is the host, not the code — but a payload declining fails and says to find a shape the scan still misses. Confirmed against both failure modes.

On preloads: the route a preload takes without meaning to is closed. `internal/cluster/primary` keeps its own `fork`, but every use goes through `cluster.fork`, which the loader now replaces when that module is already loaded. Conditional, because reading it otherwise loads a builtin the artifact does not use.

A preload that deliberately keeps `child_process.fork` and hands it back is not closable, and no runtime patch in any JavaScript program is. The alternative required outcome — keep every payload that might fork on the launcher — is the undecidable question the scan already answers as well as a syntactic pass can. The guard changes the default outcome from unbounded re-entry to an error at the call. It is not a boundary and the comment says so.

colinhacks:
Both fixed in caa96bc.

The fixture now takes `node:cluster` through the same holder shape and calls `cluster.fork()`, and the artifact runs twice: plain, and under `NODE_OPTIONS=--require` with a preload that loads cluster before main. The halves refuse by different routes and say so — plain prints `child_process.fork("./fg")`, because cluster's primary destructured the replacement; preloaded prints `cluster.fork()`, because only the conditional patch can reach it. Disabling that patch fails the preloaded half.

The preloaded half also asserts its preload printed, so a `NODE_OPTIONS` that did not apply fails rather than leaving the branch untested.

`ran.status.success()` restored, with stderr in the message.

`cluster.fork` carries its own text now. It takes an environment, not a module path, so the shared message printed `[object Object]` and closed on "instead of that module".

colinhacks:

PR #891, opened 2026-09-05
nub-phantom: walk legacy deep-path entry points for packages without an exports map

colinhacks:
A package with no `exports` map has no encapsulation boundary: Node resolves `<pkg>/<any/published/file>`. The walk seeded only `main`/`bin`/`exports`, so `redux-persist@6.0.0`'s `lib/integration/react.js` and its undeclared `require("react")` were never parsed.

Published JS files now seed the walk as speculative roots — only without an `exports` map, and tagged `from_deep_path`. Seeding is opt-in and the shipped disk-eject scan does not enable it.

A/B over the top 2000 packages: +4 hard edges, 0 removed. Two are documented deep imports (`memfs/lib/fsa-to-node`, `recast/parsers/acorn`) naming an undeclared package. An earlier build added 14; 11 were browserify bundles, whose inlined factories bind their own `require` parameter.

A parameter-bound `require` is no longer recorded as an edge. That guard is in the shared extractor, so it applies everywhere including the shipped scan, and it removes a pre-existing false positive on packages whose `main` is a bundle.

### DIFF
--- a/crates/nub-phantom-core/src/extract.rs
+++ b/crates/nub-phantom-core/src/extract.rs
@@ -21,9 +21,10 @@
 //! hardness), never toward a false phantom — safe for the never-false-flag bar.
 
 use oxc_allocator::Allocator;
+use oxc_ast::AstKind;
 use oxc_ast::ast::{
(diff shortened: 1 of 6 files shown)


colinhacks:

PR #892, opened 2026-09-05
fix: preserve CommonJS cache under standalone loader imports

colinhacks:
Exclude the standalone loader’s own import from foreign-loader detection, preserving CommonJS cache APIs on Node 22.15.

Adds packed-loader regression coverage for argv and NODE_OPTIONS delivery, plus the existing fixture matrix in CI.

Verified remotely on Node 22.15 and 24.11.1 using matched JavaScript packages and the released 0.8.3 addon. The merge-base control fails the new import cases; the patch passes. CI rebuilds the addon.


colinhacks:
Fixed both startup detection cases in 2bb659222b. Added argv and NODE_OPTIONS regressions for the underscore alias and earlier sync/async preloads. The packed-loader matrix passes on Node 22.15 and 24.11.1; the preceding head fails the new guard cases.

colinhacks:

PR #893, opened 2026-09-05
compile: flush the extracted app cache concurrently on first run

colinhacks:
A compiled artifact unpacks its payload on first run, then seals the staged tree by flushing every file. That seal was the bulk of cold start.

`File::sync_all` is `fcntl(F_FULLFSYNC)` on macOS — a full device cache flush, not the `fsync(2)` it is on Linux — and it runs immediately after extraction wrote the payload, so every barrier waits on real dirty data, one file at a time.

Measured over a 2683-file payload, staging then syncing immediately so the data is genuinely dirty:

| | |
| --- | --- |
| F_FULLFSYNC serial (before) | 2661 ms |
| F_FULLFSYNC, 8 threads | 161 ms |
| plain `fsync(2)` serial | 46 ms |

The files now go out across a bounded thread pool. Plain `fsync(2)` is faster still, but it would make macOS durability weaker than this code asks for, and a cache a power loss can tear is not a trade to make silently. Concurrency weakens nothing — read-only handles on distinct paths, every barrier complete before the function returns — and directories are still synced afterwards on one thread, depth-first.

Two smaller costs in the same loop, both from it running per file over thousands of files: `create_staging_subdirs` re-walked every ancestor of every file (~19k redundant `create_dir` calls on a payload of ~2700 files across ~500 directories), and `zstd::decode_all` built a fresh streaming context per file where the build stores each file as its own frame. Directories are now remembered, and one `zstd::bulk::Decompressor` is reused and sized from the manifest's recorded plain length.

End to end on that payload, three cold runs on a quiet host: extraction loop ~840 → ~560 ms, seal ~1800 → ~330 ms, **total ~2800 → ~980 ms**. On a loaded host, seven alternating pairs against the same payload built with the previous launcher: **3.34x, faster in 7 of 7**.

The launcher grows about 20 KB.

## Verification

- `cargo clippy --all-targets -- -D warnings`, `cargo build`, `cargo test` in `crates/nub-launcher` (its own workspace).
- `cargo fmt --check` in both the root and launcher workspaces.
- A new test drives `collect_cache_tree` over a nested tree with more files than workers and asserts every file is reachable, the directory order lets a child sync before its parent, and an empty tree still succeeds. Coverage is what the thread pool can now get wrong: a file the walk misses publishes unflushed and nothing else notices. Falsified by deliberately making the walk skip files and confirming it goes red.
- Rebuilt a real 2683-file artifact against the new launcher and re-measured cold start against a control built from the identical tree with only the launcher differing.

`__NUB_LAUNCHER_TIMING` gains a split of the extraction loop into mkdir, decode and write, plus two phases around the seal. It is accumulated only when the variable is set, since two `Instant::now()` calls per file would tax the path it exists to measure.

### DIFF
--- a/crates/nub-launcher/src/main.rs
+++ b/crates/nub-launcher/src/main.rs
@@ -931,7 +931,9 @@ fn app_cache_is_ready(view: &PayloadView<'_>, cache_dir: &Path) -> bool {
         }
         let extracted = match file.plain_size {
             Some(size) => Extracted::Size(size),
-            None => match app_bytes(&view.manifest, file) {
+            // Only a payload with no recorded size gets here, which is exactly
+            // the case a reused context could not size its output buffer for.
+            None => match app_bytes(&view.manifest, file, None) {
                 Ok(bytes) => Extracted::Bytes(bytes),
                 Err(_) => return false,
             },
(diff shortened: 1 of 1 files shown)


colinhacks:

PR #894, opened 2026-09-05
compile: stop declining every source map from the containers that can carry one

colinhacks:
`Decline::SourceMap` fired on any source map, which is broader than the measurement behind it. Its comment records `--enable-source-maps` failing to apply an inline map to a `data:` URL module — and a `data:` URL is the no-extract launcher. The single-executable container serves each module through a `registerHooks` load hook, where V8 does honour the `sourceMappingURL` it finds.

Measured on Node 26.8, one throwing `app.ts`, frames as reported:

| `--sourcemap=` | extracted (control) | single-executable | no-extract launcher |
| --- | --- | --- | --- |
| `inline` | `app.ts:2:13` | `app.ts:2:13` | `app.mjs:3:10743` |
| `linked` | `app.ts:2:13` | needs a file | needs a file |
| `external` | `app.mjs:2:10743` | `app.mjs:2:10743` | `app.mjs:3:10743` |

The single-executable container gains two modes, `inline` and `external`. The no-extract launcher gains none: it declines every map, because `rewrite_chunk` edits its chunks after the map is made — one prepended line, plus in-place specifier rewrites — so no bundler-made map still describes what it runs. The line-3 cells above are that shift. An unresolvable map is a missing map, which a reader notices; a map that resolves to the wrong line is worse, because nothing reports it.

`Inputs::sourcemap` carries the mode rather than a bool. `linked` is also caught by `NonChunkFile`; it stays here because this reason reads better. Composing `rewrite_chunk`'s edit into each map would admit both modes for the no-extract launcher, and is recorded on `Decline::SourceMap` as what that would take.

Also carries an unrelated Windows fix (b37a8dde) that this branch's CI surfaced. The build's self-probe runs the staged artifact immediately before `MoveFileExW`, and Windows releases an executable's image section asynchronously, so the publish could lose the race with `ERROR_SHARING_VIOLATION`. It now retries on that code alone. `a-sourcemap` passed in the same run and win32-x64 passed on identical code, so the failure was not the source-map change.

### DIFF
--- a/crates/nub-cli/src/compile/inline.rs
+++ b/crates/nub-cli/src/compile/inline.rs
@@ -155,24 +155,33 @@ pub enum Decline {
     /// those are is per container and per mode, and the answer is measured rather
     /// than argued — the same throwing `app.ts`, compiled six ways on Node 26.8:
     ///
-    /// | `--sourcemap=` | `Mode::Sea` | `Mode::Inline` |
-    /// | --- | --- | --- |
-    /// | `inline` | `app.ts:2:13` | `app.mjs:3:10743` |
-    /// | `linked` | needs a file | needs a file |
-    /// | `external` | `app.mjs:2:10743` | `app.mjs:2:10743` |
+    /// | `--sourcemap=` | extracted (the control) | `Mode::Sea` | `Mode::Inline` |
+    /// | --- | --- | --- | --- |
+    /// | `inline` | `app.ts:2:13` | `app.ts:2:13` | `app.mjs:3:10743` |
+    /// | `linked` | `app.ts:2:13` | needs a file | needs a file |
+    /// | `external` | `app.mjs:2:10743` | `app.mjs:2:10743` | `app.mjs:3:10743` |
     ///
     /// `linked` ships a `.map` beside the chunks and names it in a
     /// `sourceMappingURL`, which neither container can resolve; it is caught by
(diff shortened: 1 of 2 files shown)


colinhacks:
Confirmed. Fixed in 21e5bfc.

`rewrite_chunk` has one caller, inside the `Mode::Inline` path, so the no-extract launcher now declines every map mode rather than just `inline`. `Mode::Sea` rewrites nothing and still declines only `linked`.

Measured: `external --smol` reported `app.mjs:3:10743` against a line-2 map. It is back to `2:10743` and matches. The doc table had that cell as line 2; corrected.

colinhacks:
win32-arm64 was red on `a-env-strict` with os error 32 on the publish rename. Not this branch: `a-sourcemap` (`--sourcemap=inline`) passed in the same run, `a-env-strict` has no flags file, and win32-x64 passed on identical code.

The self-probe runs the staged artifact immediately before `MoveFileExW`, and Windows releases an image section asynchronously, so a reaped child is not a free file. b37a8dde retries on ERROR_SHARING_VIOLATION only, 8 attempts, 1.27s total backoff.

Unrelated to the source-map change, but it blocks this PR.

colinhacks:
Fixed in 5b2581a. The note beside `Inputs::sourcemap` described the first version, where a detached map was admissible for both containers. PR body table and paragraph updated to the corrected matrix.

colinhacks:

PR #895, opened 2026-09-05
nub-phantom: do not report a package's own baseUrl imports as phantoms

colinhacks:
The detector reported six packages for `pusher-js` and five for `react-zoom-pan-pinch` that no install can satisfy. Both publish `baseUrl`-compiled TypeScript, so their own sources import each other by bare-looking specifiers — `core/utils/url_store`, `isomorphic/runtime`, `components`, `utils/ref.utils`. Those resolve against the source root at compile time, never against `node_modules`.

Deep-path seeding (#891) is what exposed them: a package with no `exports` map now seeds every published file as a walk root, so source that only a bundler resolves gets parsed.

Before recording a bare specifier the walk now probes the package's own published file set for it. Four rules keep the probe from erasing real findings. Each was added after an A/B over the 5000 most-downloaded packages showed it losing edges, and each has a test that fails when the rule is reverted.

| Rule | Regression it prevents |
| --- | --- |
| Probe TypeScript sources only | `@azure/core-rest-pipeline` and `@typespec/ts-http-runtime` import the real `react-native` from `dist/react-native/util/*.mjs`, beside their own `dist/react-native/`. A `baseUrl` import is compile-time only, so emitted JavaScript naming a bare specifier is a real module request. |
| A module root is an ancestor of the importing file, or a root the package's own `tsconfig.json` declares — never an unrelated directory | `next` vendors its compiled dependencies under `dist/compiled/<pkg>/`, which a wider search matched against its undeclared `react-server-dom-webpack/client`. |
| Under an ancestor root, a single-segment specifier counts only as a directory's `index` | One segment is exactly the shape of a package name. `@nx/js` re-exports `dist/typescript.js` and imports the real `typescript`; recast's `parsers/babel.js` requires the real `babylon` beside its own `parsers/babylon.js`. |
| A shipped `node_modules/` tree is not a module root | A bundled copy would answer for every undeclared import around it. |

A tsconfig-declared root is exempt from the single-segment rule, since there the package itself has said which directory bare specifiers resolve against. That is what makes pusher-js's declaration-only `runtime` internal.

## Measurements

A/B against a binary built at 248608c5ad, over the same fixed package lists.

| Slice | Hard edges | Removed | Added | Soft removed |
| Top 2000 | 118 → 118 | 0 | 0 | 0 |
| 2001–5000 | 446 → 445 | 1 | 0 | 0 |
| Union of all 396 offender packages, plus the cases above | 576 → 566 | 10 | 0 | 0 |

The ten removals are the nine target-class edges (`pusher-js` × 4, `react-zoom-pan-pinch` × 5) and `node-readfiles -> src`, whose `lib/readfiles.d.ts` declares `module "src/consts"`. Nothing was added.

Still reported, and correctly: `swagger-ui-dist` keeps `lodash`, `react-syntax-highlighter`, `remarkable` and `swagger-client`; `redux-persist` keeps `react`; `object.assign` stays clean.

The probe runs for the shipped extract-time scan as well as the eval tool, since a self-reference is not a dependency on either path. Indexing costs a directory walk, so a package that publishes no TypeScript never pays for it.

Refs #891

### DIFF
--- a/crates/nub-cli/src/dynamic_phantom.rs
+++ b/crates/nub-cli/src/dynamic_phantom.rs
@@ -343,7 +343,7 @@ pub(crate) fn phantom_cache_dir() -> Option<PathBuf> {
 /// after upgrade; harmless and expected (the whole point is to pick up the better
 /// verdict). Just bump the number when the scanner logic changes — the coupling
 /// is structural, nothing else to remember.
-pub(crate) const PHANTOM_SCANNER_VERSION: u32 = 4;
+pub(crate) const PHANTOM_SCANNER_VERSION: u32 = 5;
 
 /// Version of what the linker's GVS-populate pass WRITES TO DISK for a given eject
 /// set — bumped when the same plan produces a different on-disk shape.
--- a/crates/nub-phantom-core/src/extract.rs
+++ b/crates/nub-phantom-core/src/extract.rs
@@ -24,7 +24,7 @@ use oxc_allocator::Allocator;
 use oxc_ast::AstKind;
 use oxc_ast::ast::{
     Argument, BindingPattern, ConditionalExpression, Expression, FormalParameters, IfStatement,
-    ImportDeclarationSpecifier, LogicalExpression, Statement, TryStatement,
+    ImportDeclarationSpecifier, LogicalExpression, Statement, TSImportType, TryStatement,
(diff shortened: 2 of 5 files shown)


colinhacks:

PR #896, opened 2026-09-05
nub-phantom: two type-surface classification fixes

colinhacks:
Two defects on the `.d.ts` type surface. They are one PR because the second manufactures instances of
the first, so it is not safe to land alone.

## 1. A declared `@types/<pkg>` never satisfied a type-only reference

`@turf/destination` declares `@types/geojson` and writes `import('geojson').Position` in its
declarations. TypeScript resolves that through the `@types` fallback, so nothing is undeclared — but
the classifier only ever checked the bare name, so it reported a phantom and asked consumers to
install a package the author correctly did not depend on.

The scan holds 971 type-surface findings. Sampling 40 across a deterministic spread of the ranking,
**15 are satisfied by a declared `@types/` counterpart** — 37.5%, so on the order of 360 of them. The
targets are exactly the DefinitelyTyped-only names, where no runtime package is even intended:

| target | findings | | target | findings |
| `geojson` | 112 | | `estree` | 37 |
| `react` | 46 | | `mdast` | 19 |
| `hast` | 39 | | `json-schema` | 19 |

Verified individually: `@rollup/pluginutils` → `estree` declares `@types/estree`; `@lit/react` →
`react` declares `@types/react`; `@stoplight/spectral-formats` → `json-schema` declares
`@types/json-schema`.

Scoped names use DefinitelyTyped's flattening — `@babel/core` → `@types/babel__core` — which is easy
to get wrong and would silently misclassify every scoped case.

**Scope of the rule, and what it deliberately excludes.** It applies only when *every* occurrence is
on the type surface (`types_only`, the same shape as the existing `deep_path_only`): a runtime
`require('geojson')` is not satisfied by `@types/geojson`, and that is a test. It accepts the types
package from `dependencies`, `bundled`, or either peer set — a peer is supplied by the consumer — but
**not** `devDependencies`, which is not installed downstream, so a `.d.ts` leaning on a dev-only types
package really does break. Measured over the sampled false positives: 13 `dependencies`, 2 peer, 2
dev-only, the last correctly still phantoms.

## 2. A type-position `import()` was invisible

`typeof import("typescript")` and `import("pkg").Type` parse to `TSImportType`, a TypeScript **type**
node. The visitor handled `Expression::ImportExpression` — the value-position dynamic import, a
different node — so these were never read. `.d.ts` files already extract with `capture_types: true`,
so the surface was reachable; the gap was one unvisited node type.

`volar-service-typescript-twoslash-queries@0.0.71` shows it in two lines:

```ts
import type { Plugin } from '@volar/language-service';                    // found
export declare function create(ts: typeof import('typescript')): Plugin;  // missed
```

It declares `typescript` nowhere, and Yarn hand-wrote a rule for the whole Volar family
(yarnpkg/berry#7232) precisely because no scan finds these. This is compiler output rather than a
hand-written idiom: tsc emits the form whenever a generated declaration references a type from a
module with no top-level import.

Gated on `capture_types` like `import type … from …` beside it, so a plain `.ts` still drops it and a
devDependency named only in a type annotation is never promoted to a runtime phantom.
`TSImportType.source` is a `StringLiteral`, so unlike a dynamic import there is no computed form to
skip.

## Blast radius: this changes shipped nub, not only the scan

`nub-phantom-core` and `nub-phantom-scan` ship inside `nub-cli`, and `has_unguarded_phantom` is the
global-virtual-store eject decision (`pm_engine/phantom_closure.rs:518`). Net effect is **fewer**
ejections: fix 1 removes hard phantoms wholesale, fix 2 adds a small number back. Both directions are
correct — an undeclared type import genuinely cannot resolve from a store realpath (the nub#450
reasoning), and a `@types`-satisfied one genuinely can.

Downstream, the published extensions dataset loses roughly 360 entries. That is a correctness gain;
every removed entry was advice to install a package that was never needed.

## Verification

- `cargo test -p nub-phantom-core -p nub-phantom-scan` green.
- The type-position test **fails for the right reason** with its recording disabled: it sees only
  `["@volar/language-service"]`, missing `typescript` and `webpack`.
- Eight acceptance cases against a built binary, and the **same script against a pre-fix control
  binary**, which fails exactly the cases the change is meant to add. Running that control is what
  caught a false third finding I had claimed — `@react-native-community/cli-platform-apple` was
  already detected via an ordinary import.
- A 120-package binary-vs-binary differential, run twice. The first run is what surfaced defect 1:
  the only two edges fix 2 added were both false positives of that class. The second, with both fixes
  in, removes **8 edges across 7 packages and adds none** — `rollup` -> `estree`, `remark-rehype` ->
  `hast`/`mdast`, `@turf/bbox` -> `geojson`, `vega-expression` -> `estree`, `@types/d3-scale` ->
  `d3-time`, `@types/d3-transition` -> `d3-selection`, `@turf/line-chunk` -> `geojson`. Every one is a
  package that declares the `@types/` counterpart.
- A doc-comment `import('foo')` must not become an edge; it is a comment, so it produces no node.
  Asserted as a control, and the case a regex-based survey gets wrong.

## Checked and deliberately not included

`import x = require("pkg")` (`TSImportEqualsDeclaration`) is also unvisited, and unlike a type query
it compiles to a real `require`. It appears in **0 of 59** sampled packages, so it is not worth the
node. That matcher was validated against five known-answer cases before the zero was trusted.

### DIFF
--- a/crates/nub-phantom-core/src/extract.rs
+++ b/crates/nub-phantom-core/src/extract.rs
@@ -24,7 +24,7 @@ use oxc_allocator::Allocator;
 use oxc_ast::AstKind;
 use oxc_ast::ast::{
     Argument, BindingPattern, ConditionalExpression, Expression, FormalParameters, IfStatement,
-    ImportDeclarationSpecifier, LogicalExpression, Statement, TryStatement,
+    ImportDeclarationSpecifier, LogicalExpression, Statement, TSImportType, TryStatement,
 };
 use oxc_ast_visit::{Visit, walk};
 use oxc_parser::Parser;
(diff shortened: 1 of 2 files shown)


colinhacks:
Folded into #895 — one PR for all three detector fixes.

PR #897, opened 2026-09-05
runtime: reuse inspected source for JavaScript lowering

colinhacks:
Reuse source already read by the JavaScript syntax gate when lowering is required. Loader execution order and cache keys remain unchanged.

Adds supplied-source, transformed module identity, and early-hook ordering coverage. Verified on Node 18.19, 22.15, 24.20, and 26.8; remote workspace tests, compile tests, and Clippy pass.


colinhacks:

PR #898, opened 2026-09-05
compile: retry the publish rename past the self-probe's Windows lock

colinhacks:
`nub compile` fails intermittently on Windows at its very last step, with the artifact already written, signed and verified:

```
error  atomically replacing ./bin-a-resolution.exe with staged output
       ...\.nub-compile-artifact-6948-0\bin-a-resolution.exe.tmp
       The process cannot access the file because it is being used by another process. (os error 32)
```

`verify_artifact` spawns the staged executable as a self-probe, and the publish rename runs as soon as it exits. Windows can still hold that image's section object after the process is gone, so `MoveFileExW` reports `ERROR_SHARING_VIOLATION` for a file nothing is really using. The lock is self-inflicted and clears on its own.

Seen twice on 2026-09-05 in `Compile native package islands`, on both Windows architectures and on different fixtures — [win32-arm64 on `a-decorators`](https://github.com/nubjs/nub/actions/runs/33992824346) and [win32-x64 on `a-resolution`](https://github.com/nubjs/nub/actions/runs/33993294485). Naming each build's output uniquely does not prevent it: the handle is on the staged file that was just probed, not on a name reused from an earlier build, and the second failure already had unique names.

The publish rename now retries `ERROR_SHARING_VIOLATION` and `ERROR_ACCESS_DENIED` for up to two seconds, backing off from 5 ms to a 100 ms cap. Those are the two codes `runtime_cache::classify_target_rename_failure` already reads as an in-use target. Every other failure is reported on the first attempt, exactly as before, and a destination something really holds still fails with the message it always had.

How long the lock lasts is not measured, and the comment says so: the failures are CI races that cannot be reproduced on demand. The budget is set by what it costs to be wrong in each direction rather than by a reading.

So the tests cover the loop, not the race. `ci.yml` runs the compile unit tests on `ubuntu-latest` alone, so a `#[cfg(windows)]` test would compile on every platform and execute on none. The retryable-or-not question is a parameter instead, which puts the loop on the leg that actually runs it:

- a lock that clears is retried until it does, and each attempt is a fresh rename
- a lock that never clears gives up and reports, which is what proves the loop terminates
- only the two Windows codes count as transient; a vanished staging file and a non-OS error do not

Locally: the compile unit suite at `900 passed; 0 failed; 0 filtered out`, `cargo fmt --check` clean, and CI's own `clippy --all-targets --all-features -- -D warnings` at zero warnings. A red control with the budget zeroed fails exactly the two retry tests, on different lines, and leaves the code-mapping test green.

### DIFF
--- a/crates/nub-cli/src/compile/mod.rs
+++ b/crates/nub-cli/src/compile/mod.rs
@@ -18,6 +18,7 @@
 use std::fs;
 use std::io::Read;
 use std::path::{Component, Path, PathBuf};
+use std::time::Duration;
 
 use anyhow::{Context, Result, anyhow, bail};
 use nub_core::compile::{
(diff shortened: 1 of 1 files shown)


colinhacks:

PR #900, opened 2026-09-06
compile: retry the artifact rename through a transient Windows hold

colinhacks:
The compile harness on the win32 CI legs intermittently fails the publish step with `atomically replacing ./bin-<name>.exe with staged output … The process cannot access the file because it is being used by another process. (os error 32)`. It hit #718's win32-x64 run and #583's win32-arm64 run within a few hours of each other today. An antivirus scan or the search indexer holds the freshly written executable for a moment after its last handle closes; nothing of ours still has it open.

The publish step now retries a rename that reports `ERROR_SHARING_VIOLATION` or `ERROR_LOCK_VIOLATION` with a short backoff (25 ms doubling to 400 ms, about 1.6 s in total) and adds the attempt count to the error when the hold outlasts it. A non-transient failure is still reported at once. graceful-fs retries a win32 rename on the same class of error for the same reason.

Tests: the retry path (two violations, then the rename lands), the exhausted path (the destination is untouched and the error names the attempt count), and the existing non-transient test now asserts a single attempt.

Gates: CI's clippy invocation and `cargo test -p nub-cli --features compile --bin nub atomic_publish` on a builder.

### DIFF
--- a/crates/nub-cli/src/compile/mod.rs
+++ b/crates/nub-cli/src/compile/mod.rs
@@ -3639,17 +3639,52 @@ fn same_filesystem(_a: &Path, _b: &Path) -> bool {
     true
 }
 
-fn publish_staged_with<F>(staged: &Path, destination: &Path, rename: F) -> Result<()>
+/// Backoff between publish attempts while the rename reports a transient
+/// Windows hold; about 1.6s in total before the failure is reported.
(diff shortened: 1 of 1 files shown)


colinhacks:
Superseded by #898, which retries the same publish rename with the self-probe root cause and a Windows-gated predicate.

PR #901, opened 2026-09-06
compile: run static workers inside the single-executable container

colinhacks:
A payload with a statically traced `new Worker(...)` kept the launcher. `Module.registerHooks` is per thread, so a worker started on a payload URL cannot load it — the hooks that serve that URL never ran in the new realm.

The loader now replaces `Worker`. A target naming a worker payload file starts on a generated bootstrap that calls the same function the main thread does, so a worker's own worker works at any depth. Any other target passes through, so a worker whose entry really is a file on disk still reaches that file. Both spellings arrive: `node:worker_threads` hands over the URL the chunk built; the WHATWG global converts it to a path first.

The blob carries the neutralized bootstrap as an asset as well as its main, because the wrapper a worker runs installs the record in a realm that inherits nothing. The replacement is built only when the payload has a worker — reading `node:worker_threads` loads it, at 1.4 ms per start.

Verified: 7-case adversarial sweep incl. nested and disk-path workers, with a red control; `compile-augmentation` 16/16 in both shapes; Node 24.19 / 25.7 / 26.7.


colinhacks:

PR #902, opened 2026-09-06
compile: fix the module-cycle deadlock and add a CommonJS output shape

colinhacks:
Rolldown lowers every import edge into its own `await`, which deadlocks a cycle: `a -> b -> c -> a` becomes three initializers that each await the next, and the third finds the first's in-flight promise in the memo. Exit 13, "Detected unsettled top-level await", where plain Node runs the same source.

Real ESM never has a module wait on another in its own SCC. Async initializers now carry a re-entrancy guard restoring that rule; sync ones are untouched, since their callers do not await them.

Also adds a CommonJS output shape behind `__NUB_COMPILE_CJS`. Default path unchanged.

Red to green on a cycle fixture and on opencode v2. Remote gates green.


colinhacks:

PR #903, opened 2026-09-06
launcher: probe the payload map instead of scanning it per directory

colinhacks:
The warm-start check that decides whether an extracted app still matches the payload asked "is anything expected under this directory?" with a full `keys()` scan, once per directory — on a real tree that is directories x files, 261 x 2431 for a mid-size app.

The map is already sorted, and `Path`'s component-wise ordering keeps every key under a prefix contiguous and immediately after it, so the first key at or after the directory settles it.

Compiled opencode v2, warm `--version`, 15 alternating rounds: the validation walk goes 26.26 ms to 17.85 ms, faster 15/15.

Guarantee unchanged. Verified against the binary that a one-byte truncation is still caught and re-extracted, and an unexpected directory still rejected whether it sorts before or after every payload name.

### DIFF
--- a/crates/nub-launcher/src/main.rs
+++ b/crates/nub-launcher/src/main.rs
@@ -961,6 +961,30 @@ fn app_cache_is_ready(view: &PayloadView<'_>, cache_dir: &Path) -> bool {
     matched && expected.is_empty()
 }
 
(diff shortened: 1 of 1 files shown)


colinhacks:

PR #904, opened 2026-09-06
compile: point an unresolved native addon at --unbundled

colinhacks:
The hint for an unresolved `.node` require ended on "A self-contained binary carrying its own copy of a native package is not supported yet." Not true — a package the unbundlable rules recognize is copied into the binary in its own installed layout, addon and all. Measured: `sharp` ships a 7.2 MB island and runs with `node_modules` deleted.

What the message reports is narrower: the rules did not reach this site. `--unbundled <package>` is the flag for that, and the hint named only `--external`, which needs the package already installed on the target machine.

Both flags are named now, plus a test — there was none, which is how the claim outlived the capability.


colinhacks:

PR #905, opened 2026-09-06
pm: apply the bundled compatibility database under nub

colinhacks:
The vendored Yarn + pnpm `packageExtensions` catalogs (161 entries) were gated on the embedder being standalone aube, so nub applied none. pnpm merges the same data into every install unless `ignoreCompatibilityDb` is set.

`reactcss@1.2.3` requires `react` and declares it nowhere:

| | `require("reactcss")` |
| --- | --- |
| `main` | `Cannot find module 'react'` |
| this branch | passes |
| + `ignore-compatibility-db=true` | fails |

`auto_install_peers` already defaults on, so the catalog is the whole fix.

`ignoreCompatibilityDb` leaves the inert-settings list, which had made `config set` refuse it. Without that, the repairs have no opt-out; row 3 checks it.


colinhacks:

PR #906, opened 2026-09-06
pm: withhold lockfile reuse when packageExtensions drift

colinhacks:
A `packageExtensions` edit discarded the lockfile as drift, but the same graph still went to the resolver as a reuse hint. Reuse resolves a locked package without fetching its packument, the only place an extension is applied, so the re-resolve loaded the new extensions and returned the locked deps unchanged: checksum rewritten, nothing installed.

Withholds the hint on drift, matching the release-age path above it. Steady-state installs keep full reuse.

`top_level_package_extensions_shapes_resolution_and_invalidates_freshness` covered this and was failing. It is `#[ignore]`d for network and no CI job passes `--ignored`, so the decision is now unit-tested offline, where aube-parity runs it.


colinhacks:

PR #907, opened 2026-09-06
pm: bundle the phantom-derived packageExtensions database

colinhacks:
Adds the database published as `@nubjs/extensions` — derived by running nub's phantom detector over the top 10,000 npm packages — on top of the vendored Yarn and pnpm catalogs. It is a verified superset of Yarn's.

Of the 654 entries Yarn lacks, 25 add a hard `dependencies` edge; the rest are optional peers that install nothing and only repair strict layouts. So it diverges from pnpm for 25 packages, deliberately.

Supplied via `EngineContext::bundled_package_extensions` rather than by editing the vendored catalogs: lowest precedence, and never read by the lockfile checksum, so a refresh cannot drift a lockfile. `ignoreCompatibilityDb` still declines everything.

### DIFF
--- a/crates/nub-cli/assets/bundled-package-extensions.json
+++ b/crates/nub-cli/assets/bundled-package-extensions.json
@@ -1,0 +1,9775 @@
+{
+ "_source": "@nubjs/extensions",
+ "_version": "1.0.0",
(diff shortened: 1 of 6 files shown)


colinhacks:

PR #910, opened 2026-09-07
nub-phantom-scan: record the file each phantom reference was found in

colinhacks:
Every phantom finding now records the files it was found in, plus the pre-cap count.

The provenance bits say which entry surface *reaches* a reference; they cannot say where it lives. That gap made the deep-path class unauditable: `redux-persist/lib/integration/react` (a real legacy entry point) and `pusher-js/integration_tests_server/index.js` (a test server shipped because the package declares no `files`) reach the same verdict.

Auditing it without the path got it wrong. `swagger-ui-dist`'s 33 edges were written off as a bundled `dist` with every target inlined — true of `swagger-ui-bundle.js`, false of the two files flagged. `swagger-ui.js` externalizes them, the package declares only `@scarf/scarf`, and a clean install throws `MODULE_NOT_FOUND` on `require("swagger-ui-dist/swagger-ui.js")`.

This also supplies what `crates/nub-phantom`'s probe harness asks for in its own header: the file, so it can require an offending path directly rather than settling for whether a bare import breaks.

Verified end to end against three real packages; the new test goes red alone under a negative control.

### DIFF
--- a/crates/nub-phantom-scan/src/classify.rs
+++ b/crates/nub-phantom-scan/src/classify.rs
@@ -144,7 +144,14 @@ pub fn classify(manifest: &Manifest, references: &[Reference]) -> Vec<Finding> {
         if !e.specs.contains(&r.raw) {
             e.specs.push(r.raw.clone());
         }
-        e.files.insert(r.file.clone());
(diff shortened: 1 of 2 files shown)


colinhacks:

PR #912, opened 2026-09-07
runtime: enable AsyncContextFrame for AsyncLocalStorage on Node 22.9–23.x

colinhacks:
Node 24 rebuilt `AsyncLocalStorage` on V8 context frames and made it the default (nodejs/node#55552). Node 22 LTS and 23 carry the same implementation behind `--experimental-async-context-frame`. This adds a matrix row that injects the flag on [22.9.0, 24.0.0).

Floor is 22.9.0 rather than the 22.7.0 landing: 22.7 and 22.8 throw `TypeError: Method Map.prototype.set called on incompatible receiver` under the flag (nodejs/node#54503, fixed in 22.9.0). From 22.16 the implementation file is byte-identical to Node 24's default apart from the option name.

Measured on an idle c3-standard-8 (Linux x64, Node 22.23.2), three rounds each:

| workload | `node` | `nub` | `nub --node` |
| --- | --- | --- | --- |
| store/await loop, 40k requests × 8 awaits | 182 / 193 / 190 ms | 96 / 96 / 97 ms | 191 / 207 / 197 ms |
| Fastify 5 + OpenTelemetry SDK, autocannon -c 50 -d 10 | 16119 / 16359 / 16170 req/s | 17008 / 17236 / 17858 req/s | 16148 / 16055 / 16709 req/s |

Plain `node --experimental-async-context-frame` lands where `nub` does (94–96 ms; 17622–17708 req/s), so the gain is the flag and nothing else Nub injects.

Verified with the built binary: `process.execArgv` carries the flag on 22.23.2 and 23.11.0, and not on 24.19.0 or under `--node`. A user `--no-experimental-async-context-frame` in argv or `NODE_OPTIONS` opts out through the existing negation subtraction. Like every gated flag it rides argv, so an embedded Electron below the tree never sees it.

Memory: nodejs/node#61882 (an idle HTTP parser keeps the last request's store until reuse) does not change how much Node 22 retains. Counted by store liveness (a WeakRef per store, the issue's request shape, after GC) on 22.23.2: with the flag, 50 stores alive at concurrency 50, 1 at concurrency 1, 5 with `http.setMaxIdleHTTPParsers(5)`; without the flag, 52 and 3 on the same runs, held by a different pooled resource that the parser cap does not move. Both are flat from 500 to 5000 requests and bounded by peak concurrency. The upstream fix (#61995) reached 24.15 and 25.8.1 only, where the same probe reports 0. Injecting the flag moves the retainer, not the amount retained.

Gates: `cargo fmt`, scoped matrix and flag tests, `lat check` locally; `clippy --all-targets --all-features` on a remote builder.

Docs: new `AsyncLocalStorage` entry under Node.js APIs on the Modern APIs page; architecture feature counts updated.


colinhacks:

PR #913, opened 2026-09-07
runtime: set the `nub` export condition on augmented runs

colinhacks:
Nub sets a `nub` export condition on the runs its CLI augments, so a package can carry a `nub` branch alongside `bun` and `deno` ones. The name follows the WinterTC runtime-keys convention; the registry entry is proposed separately in https://github.com/WinterTC55/runtime-keys/pull/39

Conditions are a set, so a package with no `nub` key resolves as it does on plain Node. Compat mode is unaffected — all five call sites were already gated. The standalone `@nubjs/loader` deliberately leaves Node's conditions alone, so a file resolves the same under it as under tsx.

The compile bundler and the bare-preload resolver take the condition too, since both resolve ahead of the child Node.

Removing the condition turns the run-path test red on `{"nub":"default"}`; swapping the key out of the resolver turns the bundler test red.


colinhacks:

PR #915, opened 2026-09-07
config: a prefix field that wraps file runs, scripts, and watch

colinhacks:
Adds `prefix` to `nub.jsonc`: a command Nub puts in front of `nub <file>`, `nub run <script>`, and `nub watch`.

```jsonc
{ "prefix": "dotenvx run --" }   // or ["dotenvx", "run", "--"]
```

- A file run and a watch wrap Node; a script wraps the shell, so a script that starts no Node process still runs behind it.
- The program resolves from the project's `node_modules/.bin` chain, then `PATH`; a path form anchors to the file that set it. A miss is `ERR_NUB_PREFIX_NOT_FOUND`.
- A marker keeps a nested `nub run`, or a wrapper whose own bin is a Node script, from wrapping the same project again.
- `nubx`, `nub dlx`, and the `node` PATH shim do not take it. Env files keep loading.
- Docs, schema (`latest` + `v0.8`), `nub config init` templates, and `nub config set prefix …`.

Verified against a real `@dotenvx/dotenvx` install on all three surfaces, plus a six-test integration file with a stub wrapper.


colinhacks:

PR #917, opened 2026-09-07
Tune the main nursery for memory-constrained Node launches

colinhacks:
## Change

Nub increases the main JavaScript heap's young-generation capacity on measured memory-constrained launches. It starts the installed, unmodified Node binary with `--max-semi-space-size=16`, then clears that process-global override before application code runs so later Workers retain their own heap limits.

A small young generation fills quickly on allocation-heavy workloads, causing frequent garbage collection. **16 MiB is the semi-space setting, not the total heap size or a process memory cap.** The old-generation setting is unchanged.

## Systems and scenarios

Automatic tuning applies only on **Linux x64**, with readable cgroup v1 or v2 limits, within these inclusive ranges:

| Exact Node release | Leaf cgroup memory budget | Default semi-space | Nub |
|---|---|---|---|
| 22.23.2 | **512 MiB–1 GiB** | 1–4 MiB | 16 MiB |
| 24.20.0 | **512 MiB** | 1 MiB | 16 MiB |

The Node 22 range is continuous: 640 MiB, 768 MiB, and other intermediate budgets qualify. **Node 24 has a genuine sizing cliff:** its default jumps from 1 MiB at 512 MiB to 16 MiB immediately above that budget, so it needs no larger-budget override.

The detector also accounts for tighter ancestors and physical memory. The effective limit must remain at least 512 MiB. A 1 GiB leaf under a 256 MiB parent is excluded for insufficient memory; an unlimited or oversized leaf under a 512 MiB parent is excluded because the parent restriction does not establish that Node selected a small nursery.

The launch must be a direct file run or a Node-backed `nub exec` / `nubx` binary using Nub's normal CJS preload. Tuning is disabled by:

- Node flags before the entry path, nonempty inherited `NODE_OPTIONS`, or configured Node options, V8 flags, preloads, or additional export conditions. Nub's built-in `--conditions=nub` is allowed.
- Compatibility mode (`--node`, `NODE_COMPAT`, or configured `nodeCompat`).
- Yarn PnP, a command prefix, or an external environment-loader wrapper.

Application arguments after the entry path do not disable it. macOS, Windows, other CPU architectures, unconstrained Linux processes, Node 26 and other unlisted releases, watch mode, compiled executables, ordinary package-script subprocesses, and the persistent `node` shim keep their existing behavior. Unsupported versions skip the cgroup scan entirely.

## Worker and child-process behavior

Leaving the global semi-space flag active would override explicit Worker young-generation limits, even with `execArgv: []`. The startup helper instead:

1. Lets V8 initialize the main isolate with the larger semi-space.
2. Resets the global flag before application preloads or the entry point execute. Main's stored limit is unchanged; subsequent Workers use their own defaults and `resourceLimits`.
3. Removes Nub's injected arguments from `process.execArgv`, including in Workers with `env: {}`, so forks and explicit argument forwarding do not propagate the override.

This is a one-shot startup operation, **not adaptive tuning during execution**. Extending the release allowlist requires auditing V8's flag readers and Node's startup ordering, then repeating compatibility and performance tests. No new user-facing setting is introduced.

## Built-Nub measurements

Dedicated Linux x64 c3-standard-8 VMs, no swap, interleaved before/after repetitions, matching workload checksums. The clean environment leaves `NODE_ENV` unset: **React SSR in this table uses the development build**, not production mode.

| Node / budget | Workload | Before median | After median | Runtime reduction |
|---|---|---:|---:|---:|
| 24.20.0 / 512 MiB | React SSR, development | 11.260 s | 8.626 s | 23.4% |
| 24.20.0 / 512 MiB | Retained JSON | 2.995 s | 1.989 s | 33.6% |
| 24.20.0 / 512 MiB | Terser | 3.721 s | 3.106 s | 16.5% |
| 22.23.2 / 768 MiB | React SSR, development | 9.411 s | 8.798 s | 6.5% |
| 22.23.2 / 768 MiB | Retained JSON | 3.365 s | 1.971 s | 41.4% |
| 22.23.2 / 768 MiB | Terser | 3.489 s | 3.245 s | 7.0% |

The Node 24 comparison used five repetitions against base `4f49597a62`. The Node 22 comparison used three repetitions against `9a7d1754f9` with the previous 512-only policy; both arms otherwise used identical sources and the same addon. Low-allocation medians remained approximately 0.508 s. These SSR/Terser cases used 19–34 MiB more median peak RSS; retained-object RSS varies with GC timing.

## Production-mode controls and rejected ranges

A separate confirmation ran stock Node with `NODE_ENV=production`, comparing its defaults with the same semi-space flag. Ten interleaved repetitions per arm and budget ran on a fresh, quiet c3-standard-8 VM without swap. These isolate the GC setting; they are **not built-Nub production benchmarks**.

| Node / budget | Production SSR before | After | Runtime change | Policy |
|---|---:|---:|---:|---|
| 22.23.2 / 512 MiB | 1.539 s | 1.280 s | -16.8% | Enabled |
| 22.23.2 / 768 MiB | 1.290 s | 1.265 s | -1.9% | Enabled |
| 22.23.2 / 1 GiB | 1.281 s | 1.262 s | -1.5% | Enabled |
| 22.23.2 / 2 GiB | 1.223 s | 1.276 s | **+4.3%** | Excluded |
| 24.20.0 / 512 MiB | 1.980 s | 1.174 s | -40.7% | Enabled |
| 26.8.1 / 512 MiB | 1.081 s | 1.107 s | **+2.4%** | Excluded |
| 26.8.1 / 768 MiB | 1.029 s | 1.101 s | **+7.0%** | Excluded |
| 26.8.1 / 1 GiB | 1.028 s | 1.095 s | **+6.6%** | Excluded |

The rejected cases were slower in all ten pairs. This changed the policy: retained-JSON gains alone did not justify enabling those SSR regressions. All sampled runs completed with matching checksums and zero OOM kills, but these remain workload-specific measurements, not a universal speed or memory guarantee.

**Budgets below 512 MiB are excluded.** Earlier built-binary pressure tests at 256 MiB found extra OOM kills with tuning: Node 24 retained-JSON failures were 1/3 versus 0/3 at baseline. Workers, native allocations, and other processes still consume the same container budget.

## Verification

- Remote full tests, native/launcher gates, and 376 compile tests passed. After integrating current main, a fresh-addon build passed **143 acceptance cases**, all **500 active core tests**, and all-target/all-feature Clippy. Formatting, wiki checks, and the site production build also passed.
- The acceptance suite covers all three tested Node releases, including unchanged Node 26 behavior, across 14 budgets per release. It checks stock-Node sizing, explicit overrides, Worker limits, untuned fork heaps, startup exclusions, and paired allocation pressure.
- Rust tests cover byte-exact boundaries, lazy detection, and distinct leaf/ancestor limits. An additional 18 real hierarchical-cgroup Node/Nub launches passed with Workers and forks.
- Exact-head CI passed: **111 green checks**, with four intentionally skipped checks. The Linux CI run independently passed all **143 acceptance cases**; Linux/macOS/Windows embedded-runtime checks also passed. The startup helper is included in embedded-runtime integrity verification.

Refs nodejs/node#42511


colinhacks:

PR #918, opened 2026-09-07
pm: accept the npm lockfiles and projects that npm ci accepts

colinhacks:
Eight refusals of npm-written lockfiles and projects, found by frozen-installing the most popular npm-using repos on GitHub. All are green under `npm ci` in the repos' own CI.

- A workspace member's required peer was not recorded on its importer, so the freshness check read every member with a peer as `manifest adds <peer>@<range>` (apollo-server, socket.io). The reader now records required peers the way the resolver seeds them under `auto_install_peers`.
- A member nested inside another member (`test/installation`) never looked in the parent member's own `node_modules`, jumping straight to the root (puppeteer). `resolve_nested` walks directory ancestors for a member target.
- A link target with no `version` failed to parse; npm writes none for a name-only local package (mocha, lerna, lit). The reader takes the resolver's own `0.0.0` default.
- A package listed under two manifest sections produced two importer rows and tripped the section check (promptfoo, theia). One row per name, first section wins.
- A space-separated multi-hash SRI value (`sha1-… sha512-…`) was read as one token and failed integrity verification (a git dependency's nested `prepare` install in insomnia). `parse_sri` picks the strongest supported algorithm and accepts any of its digests, with `?option` suffixes ignored, as ssri does.
- A workspace member's `install`/`postinstall`/`prepare` ran with only the member's own `node_modules/.bin` on PATH, so a bin that only the root declares was a 127 (puppeteer's `tools/eslint` runs `wireit` from its `prepare`), reported as `root prepare script failed`. `run_member_hook` walks every ancestor `.bin` out to the filesystem root, as npm's run-script does, and the failure names the member.
- A root `preinstall` that runs `node-gyp install` (theia) found no node-gyp on PATH, which npm provides to every lifecycle script from its bundled copy. Importer hooks now get the same lazy node-gyp shim dependency builds already got; it stays out of the way when a project or system node-gyp resolves.
- `nub install` refused to start when the project's tsconfig extends a package the install is about to provide (ava). The PM verbs build their lifecycle options with a best-effort tsconfig gate; runs that execute user code keep the fatal gate from #731.

Each fix has a test beside the code (`aube-lockfile` npm reader, `aube-store` integrity, `aube-scripts` bin chain, `nub-cli` `install_engine`). Verified on a builder VM: the reader tests go red on the merge-base reader, and `nub install --frozen-lockfile` now succeeds on the real lockfiles of mocha, socket.io, apollo-server, promptfoo, lerna, lit, theia, puppeteer and ava.

### DIFF
--- a/crates/nub-cli/src/cli.rs
+++ b/crates/nub-cli/src/cli.rs
@@ -4185,7 +4185,32 @@ pub(crate) fn runtime_node_options(
     runtime: &mut crate::project_config::RuntimeConfig,
     node: &nub_core::node::discovery::ResolvedNode,
 ) -> Result<Vec<String>> {
-    runtime_node_options_with(runtime, node, FoldInherited::Yes)
+    runtime_node_options_with(runtime, node, FoldInherited::Yes, TsconfigGate::Required)
+}
+
+/// The options a PM verb hands its lifecycle scripts. Same set as a run, except a
+/// tsconfig that will not read is tolerated: the install is what makes the
+/// `extends` target exist (ava's `"extends": "@sindresorhus/tsconfig"` is a
+/// devDependency), so refusing here left every fresh clone unable to install.
+pub(crate) fn lifecycle_node_options(
+    runtime: &mut crate::project_config::RuntimeConfig,
+    node: &nub_core::node::discovery::ResolvedNode,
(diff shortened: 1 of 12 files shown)


colinhacks:

PR #919, opened 2026-09-07
runtime: size libuv's threadpool to the cores on augmented runs

colinhacks:
Node leaves libuv's threadpool at its default of 4 threads whatever the core count, and every `fs` call, `dns.lookup` (so every `fetch` to a new host), async `zlib`, and async `crypto` (`pbkdf2`, `scrypt`, `randomBytes`, `bcrypt`, `sharp`) queues on those four. Node reads `UV_THREADPOOL_SIZE` at the first pool use, so the spawn is the only place it can be set. This sets it to `max(4, available cores)` on augmented runs when the user has not set it. Cores come from `std::thread::available_parallelism`, so a container gets its cgroup quota. The value is clamped to the thread headroom the existing `resource_limits` detector reports (cgroup `pids.max`, `RLIMIT_NPROC`), never below 4, because libuv creates the whole pool on first use and aborts when a thread cannot be created; the detector moves from `nub-cli` into `nub-core` for that.

## Why Node itself does not do this

nodejs/node#61533 proposed the same default for Node, was approved eight times, and was closed by its author on 2026-08-10: a host with many CPUs does not have them free, and a pool that saturates the box is worse than queued I/O. The same thread measured fs reads at −15% with a 16-thread pool on 16 vCPU (nodejs/node#57911 adds that Windows commits ~8 MB of stack per pool thread). Three things here answer those objections; the size itself is unchanged.

1. **The threads beyond Node's four run at nice 10 on Linux** (Chromium's best-effort tier). The preload builds the pool with one `fs.access` (libuv creates every worker synchronously inside the first submit, and `access` never takes the io_uring path that lets `stat`, `read` and `open` skip the pool on Node 20.3–20.11.0), takes the threads that appeared across that call as the pool, and calls `os.setPriority(tid, 10)` on workers 5..n. On the compat tier (Node 18.19–22.14) the ESM loader reads the `--import` preload through the pool, so the pool would already exist and pre-1.50 libuv names no worker; the launcher `--require`s `runtime/threadpool-snapshot.cjs` ahead of it on Linux, which builds the pool the same way and records its threads. That sidecar is its own `NODE_OPTIONS` token ahead of a PnP `--require` and of the preload, outside the re-entrancy key, and hashed as an entrypoint like `watch-env-guard.cjs`.
2. **Children keep Node's default.** The preload deletes nub's own value from `process.env` after the pool exists, so a `child_process` spawn or a `cluster` worker starts with Node's four, while a child that runs through `nub` is sized again. Ownership is the launcher's marker plus the compat presence bit, so a shell or env-file value inherits exactly as under Node.
3. **Windows stops at 8.**

## Measured

All on Node 26.8.1, GCE spot VMs, Fastify, autocannon `-c 64 -d 15`, 3 rounds; `node` is the same binary with the pool left at 4.

c3d-standard-16 (16 vCPU, 8 physical cores), pool 4 → 16:

| route | node req/s | nub req/s | p50 |
| --- | --- | --- | --- |
| bcrypt cost 12 | 17.9 | 66.2 (3.7×) | 3538 → 952 ms |
| scrypt N=2^15 | 45.0 | 115.4 (2.6×) | |
| pbkdf2 600k | 47.2 | 130.7 (2.8×) | |
| sharp 12 MP → 320 px | 56.8 | 181.2 (3.2×) | |
| gzip 1 MB, streaming | 697 | 1032 (1.5×) | |
| gzip 1 MB, one-shot | 701 | 1075 (1.5×) | |

c3d-standard-8 (8 vCPU, 4 physical cores), pool 4 → 8: bcrypt 17.9 → 33.1, scrypt 43.8 → 58.5, pbkdf2 47.1 → 65.4, sharp 54.8 → 91.4, gzip 606 → 714 / 627 → 759. The gain tracks physical cores.

Plain Node with the pool at 16 on the same 16 vCPU (`tests/bench/runtime/threadpool.sh`, 5 rounds): pbkdf2 +100%, stat +4%, a 270 KB file read +9%, a 55 KB gzip −7%, a 400-lookup DNS burst 84 → 59 ms. The fs regression upstream measured does not show on these routes.

Busy host, same box, bcrypt beside 12 nice-0 busy-loop processes (`tests/bench/runtime/pool-priority-nub.sh`, 2 rounds), the shipped binary against `node`:

| | idle req/s | busy req/s | neighbours' throughput while busy |
| --- | --- | --- | --- |
| node, pool 4 | 17.8 | 17.0 | 1781 Mloops/s |
| nub, pool 16, 12 threads at nice 10 | 66.2 | 20.0 | 1763 Mloops/s (−1.0%) |

Without the demotion (`pool-priority.sh`, plain node + a renice preload) a 16-thread pool at nice 0 took the neighbours to 92.5% and served 37 req/s; nice 10 gives them 98.6% and the idle number is unchanged.

## Launchers, precedence, nesting

Every augmented launcher installs it: the direct file run, `nub run`, `nubx` / `nub exec`, `nub watch`, and lifecycle scripts, through the shared `AugmentationEnv`. A value from `.env` or `--env-file` is the user's as much as a shell value: the ownership marker tells nub's installed default apart from a shell value, and every env-file precedence check routes through one predicate that lets a file set a key that is absent or that holds nub's marked default. That covers the two boundaries that otherwise shadowed the file, a `nub run` script re-entering through the `node` shim and `nub watch` handing Node its `--env-file`, where watch now stands down and removes an inherited default; the `nub run` and non-Node `exec`/`nubx` launchers stand down the same way when the explicit `--env-file` layer sets the key, and the key comparison folds case on Windows. `--node` and `NODE_COMPAT=1` set nothing, and a `--node` child of an augmented parent sees no variable, because the variable joins the restorable set (`__NUB_COMPAT_UV_THREADPOOL_SIZE` and its ownership markers, denylisted from `.env` sources like the rest).

Verified with the built binary on a Linux 8-vCPU VM per Node 18.19, 20.11.0, 22.13, 22.15 and 26.8: pool 8 with workers 5..8 at nice 10, `process.env` without the variable, a `node` child and a cluster worker at Node's default, a `nub` child at 8 again, `UV_THREADPOOL_SIZE=6` left at 6 with nothing demoted, `UV_THREADPOOL_SIZE=3` inherited by every child, `--node` at 4.

## Tests, docs

`crates/nub-cli/tests/threadpool_size.rs` with `tests/fixtures/threadpool/{size,watch-size,children}.js`: sizing, `nub run`, shell value, `.env` on a direct run and under `nub run` and `nub watch`, shell over file, `--env-file` on a non-Node `exec` bin, a differently cased key on Windows, `--node`, `NODE_COMPAT`, children and a cluster worker at Node's default with a `nub` child sized again, a user value inheriting, and on Linux the demotion through both preload entries under plain `node` (CI's runners have four cores, so the launcher itself sizes nothing to demote there) and a user pool left undemoted. Unit tests cover the sizing helper (floor, cores, headroom clamp, Windows cap), the compat-tier sidecar token and its classification as nub's own. Gates: `cargo fmt`, `clippy --all-targets --all-features` and the full `cargo test` on a remote builder.

Docs: a Threadpool page under Runtime (`site/content/docs/runtime/threadpool.mdx`), linked from the Augmentations list on the Runtime page.


colinhacks:

PR #920, opened 2026-09-07
runtime: let a user hook layered above nub's keep ownership of its files

colinhacks:
tsx layered over nub — `tsx script.ts` under `nub run`, bruno's `bruno-sqlite` `prepare` under `nub install` — failed with `require is not defined in ES module scope`. Three composition gaps, each pinned by the new `loader-outer-hook` fixture on both tiers:

- nub's load results omitted `responseURL`, which Node's default load always sets; tsx's CommonJS branch keys on it. Every file-URL result nub short-circuits now carries it. Node's own sync-hook chain drops it between hooks through 22.21, 24.11.0 and 25.0, so the fixture probes for pass-through before checking.
- tsx's resolve hook labels a `.ts` file with a bare `commonjs`/`module` format and expects the raw source back to transform itself. Node never labels TypeScript that way, so both tiers now step aside on that signal (the old rule only knew the `typescript` spelling). On the fast tier the registration must also carry a load hook; a resolve-only hook that labels a file keeps nub as its transformer.
- On the compat tier nub's `--import` preload overwrote the `require.extensions['.ts']` handler tsx's `--require`d preflight had installed first, so a mixed `import` + `require` file died in nub's handler. A pre-existing handler now stays in charge of its extension, and the dependency-resolution retry strips only a key that still holds nub's own handler, so `require("dep/sub")` keeps finding a dependency's `sub.ts` through a user handler installed before or after nub's preload, and a user's `.cjs` handler that nub only wraps stays theirs too.

Verified against real tsx 4.23 on Node 26, 22.15 and 20.15, and bruno's `prepare` on its pinned Node 22.12 (same output as `npm run prepare`).

### DIFF
--- a/crates/nub-cli/tests/integration.rs
+++ b/crates/nub-cli/tests/integration.rs
@@ -1956,6 +1956,19 @@ fn esm_source_reuse_preserves_loaded_bytes_and_module_identity() {
 }
 
 #[test]
+fn user_hooks_layered_above_nubs_keep_ownership_of_their_files() {
+    // tsx under `nub run` is the case in the wild. The fixture's hooks model it
+    // on whichever tier this Node runs: an outer load hook keys on the
+    // `responseURL` nub's transpiled result carries; an outer resolve hook that
+    // assigns a `.ts` file a bare `commonjs` format gets the raw source back to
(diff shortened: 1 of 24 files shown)


colinhacks:

PR #921, opened 2026-09-07
build: rename the docs bake key to __NUB_DOCS_KEY

colinhacks:
The build wrapper exports the docs bake key, cargo passes its environment to every test binary, and from there it reached the scripts nub spawns: `brand_boundary_no_globals_no_env` failed under `scripts/rust-build.sh test` and passed under plain cargo. Renamed to the `__NUB_` prefix nub's internal plumbing already uses, which the test does not count. Verified: the wrapper's environment now carries `__NUB_DOCS_KEY` only, and `cargo check -p nub-cli` runs the build script clean.


colinhacks:

PR #922, opened 2026-09-07
tests: an npm-incumbent corpus, frozen-installed under nub and checked against its lockfile

colinhacks:
Adds `tests/npm-corpus/`: 28 real npm-incumbent projects pinned by commit, frozen-installed under nub on a cold store and checked against their own `package-lock.json` (every importer's declared dependencies present at the pinned version, by Node's `node_modules` walk, so npm's hoisted layout and nub's isolated one meet one contract). `npm ci` runs on the same clone to classify a red; `expected-failures.txt` carries the one known red (rollup's patch-package postinstall).

`npm-corpus.yml` builds nub once and fans out one job per project on its own Node version: push to main on the PM surface, opt-in PR runs, a nightly that also times `npm ci`, and on demand.

PR #923, opened 2026-09-08
pm: record the peer each npm lockfile placement resolves to as a graph edge

colinhacks:
npm records the copy of a peer each dependent sees (`peer: true`, hoisted or nested) and ties it to no dependency edge. The reader left peers to the peer pass, which (1) let a peer-only package reached only through a transitive package's peer edge be pruned as unreachable — `react-dom` under react-element-to-jsx-string in ava and dotenv, `search-insights` in promptfoo, `@types/node` in axios — and (2) with no satisfying ancestor, picked the newest version in the graph or an out-of-range ancestor copy (eslint-plugin-react got eslint 10 with npm's nested eslint 9 in the lockfile).

The reader now resolves each peer by placement and records it as a dependency edge; the writer keeps such an edge out of `dependencies`. Found by the corpus in #922.

A third shape, also from the corpus: a `link` entry inside `node_modules` whose target is a registry package above it (puppeteer's nested browserslist) was read as a local package with no integrity, so the package was never fetched. Such a link now resolves to its target.

The root's and a workspace member's own declared peers, which the reader enters on their importers as direct deps, keep npm's `peer: true` on write; a member-only dep the existing lockfile hoisted stays at the root on a rewrite, and a member entry names itself only when the name differs from its directory. A second native npm fixture covers both importers.

### DIFF
--- a/tests/npm-corpus/expected-failures.txt
+++ b/tests/npm-corpus/expected-failures.txt
@@ -8,13 +8,3 @@
 # isolated layout keeps only direct dependencies there, so the patch targets are
 # absent. Installs with `node-linker=hoisted` in the project's .npmrc.
 rollup/rollup postinstall patch-package expects transitive packages hoisted to the root
-
-# A peer npm auto-installs and marks `peer: true`, reached only through a transitive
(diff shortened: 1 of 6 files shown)


colinhacks:

PR #925, opened 2026-09-08
pm: run npm ci and npm install on nub's engine behind `nub pm shim --route-installs`

colinhacks:
The shims run the manager the project pins, so a bare `npm ci` under them is still npm's install. With `nub pm shim --route-installs` the `npm` shim runs `npm ci` as `nub ci` and a bare `npm install` as `nub install --no-frozen-lockfile`, from the project's own `package-lock.json`, and says so on stderr. Only those verbs, only the flags the engine honors verbatim, only at top level, only with an npm lockfile; any other argv reaches the real npm untouched. The default is unchanged; `--no-route-installs` turns it off.

npm's script semantics carry over: `ignore-scripts` from the environment and `.npmrc` (the command line outranks both), explicit boolean values, `--include` over `--omit` in either order, and `NODE_ENV=production` for lifecycle scripts when dev is omitted.

Engine: npm's link build pass. A `file:` directory dependency's `preinstall`/`prepare`/`install`/`postinstall` now run in the link target on an npm-lockfile install, as npm runs them: the implicit `node-gyp rebuild` for a `binding.gyp` target, bins relinked after the pass, and a link reachable only through `optionalDependencies` whose scripts fail leaves the tree with its bins (npm trashes the node) and is retried on the next install. The dependency build loop no longer reaches a link.

Docs: the shim page.

### DIFF
--- a/crates/nub-cli/src/cli.rs
+++ b/crates/nub-cli/src/cli.rs
@@ -3399,6 +3399,7 @@ fn dispatch_subcommand(rest: Vec<String>) -> Result<i32> {
                 node_linker,
                 registry,
                 dir,
+                allow_all_builds: false,
                 filter: crate::pm_engine::WorkspaceFilterFlags {
                     filter,
                     filter_prod,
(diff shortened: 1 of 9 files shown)


colinhacks:

PR #928, opened 2026-09-08
pm: accept a dependency declared in more than one manifest section

colinhacks:
`--frozen-lockfile` failed `ERR_NUB_OUTDATED_LOCKFILE` on a lockfile matching its manifest, whenever a dep is declared in two sections. opencode declares `@cloudflare/workers-types` in both `dependencies` and `devDependencies`; `bun.lock` records both rows. npm, pnpm and bun accept this.

The drift check kept one `DepType` per name (`or_insert`, `dependencies` winning) and compared it against every lockfile row, so the second never matched. It now collects every declared section.

A control test keeps the reproduction honest: a dep in `dependencies` only, with an `optionalDependencies` row, is still `Stale`. 595 tests pass; clippy and fmt clean.

### DIFF
--- a/vendor/aube/crates/aube-lockfile/src/drift.rs
+++ b/vendor/aube/crates/aube-lockfile/src/drift.rs
@@ -575,6 +575,21 @@ impl LockfileGraph {
             .iter()
             .filter_map(|d| d.specifier.as_deref().map(|s| (d.name.as_str(), s)))
             .collect();
+        // Keyed by section as well, because a name declared in two sections has
+        // two rows and `lockfile_specs` keeps only whichever collected last. A
+        // faithful lockfile for `dependencies.foo = "^1"` plus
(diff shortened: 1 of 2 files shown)


colinhacks:
Specifier matching is now keyed by name and section (`5810e7df90`), so a dual-section dep with different specifiers reads `Fresh`. A dev spec that genuinely disagrees with the dev row is still `Stale`.

Not doing the second half. A declared section with no lockfile row stays `Fresh` deliberately: the npm reader emits one row per name on purpose — see the comment at `npm/read.rs:460` about promptfoo declaring `@anthropic-ai/claude-agent-sdk` as both dev and optional — so requiring a row per declared section would refuse every frozen install on an npm project with a dual-declared dependency.

colinhacks:

PR #929, opened 2026-09-08
pm: accept a bun-style source identity as a patchedDependencies key

colinhacks:
`ghostty-web@github:anomalyco/ghostty-web#83c0a07` failed with `ERR_NUB_PATCH_NON_SEMVER_RANGE`: `classify_patch_key` mirrors pnpm, which reads any non-`*` selector as a semver range.

Bun never parses these keys — it hashes the key and looks a package up under `<name>@<resolution>`. That resolution is already what `classify_bun_ident` stores as `LockedPackage.version`, so the key classifies as `Exact` and the existing lookup matches.

The branch is reached only after the range parse fails, and the gate is a protocol token, so `foo@^^1:2` still errors.

Verified on opencode: unfixed nub refuses, fixed nub links a patched `ghostty-web`.

### DIFF
--- a/vendor/aube/crates/aube-lockfile/src/bun/read.rs
+++ b/vendor/aube/crates/aube-lockfile/src/bun/read.rs
@@ -3,9 +3,10 @@ use super::raw::{BunEntry, RawBunLockfile};
 use super::source::{
     bin_value_to_map, bun_key_to_alias_name, classify_bun_ident,
     rebase_workspace_scoped_local_source, resolve_nested_bun, resolve_workspace_dep, split_ident,
-    unrecognized_protocol,
 };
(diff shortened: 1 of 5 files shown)


colinhacks:
Scoped. `classify_pnpm_patch_key` is pnpm's grammar exactly; `classify_patch_key` is that plus bun's source identity, reached only on pnpm's own rejection. Keys from `pnpm.patchedDependencies`, this tool's namespace and `pnpm-workspace.yaml` validate through the strict one in `load_declared_patch_paths`, where the origin is still known — after the merge it is gone.

`pnpm/write.rs` keeps the permissive classifier: it builds groups from keys a lockfile already recorded, so refusing one there would reject a lockfile rather than a declaration, and its loop skips `local_source.is_some()` packages anyway.

`beff4c1c5d`.

colinhacks:

PR #931, opened 2026-09-08
tests: leave the tests the reference node skips out of the node-relative denominator

colinhacks:
A test the reference Node skips (`common.skip()`, exit 0) counted as a Node pass, so every runtime was credited for the guard and a runtime that runs the test was charged for it. The official Node build skips all 245 `quic/` tests; Bun ships `node:quic` and runs them.

`judge` records `skipped: true`; `nodeRelative` drops node-skipped files from numerator and denominator (raw scoring unchanged). Node re-measured in full: 364 skips. README tables, homepage, blog and wiki figures refreshed; `readme-table.mjs --check` and `lat check` green.


colinhacks:

PR #935, opened 2026-09-11
pm: fill node-gyp's header cache from the provisioned Node

colinhacks:
node-gyp downloads `node-v<ver>-headers.tar.gz` from nodejs.org the first time it builds for a version, though every Node in nub's store already carries those headers. This writes the cache entry node-gyp would have downloaded (`<devdir>/<version>/include` plus `installVersion`) just before node-gyp runs, so a from-source build works offline.

No `npm_config_*` is exported. node-gyp folds `npm_config_*` over its own argv, and the variable is inherited, so an exported `nodedir` beats a target that a child process picked — which is how `@electron/rebuild` and electron-builder pick one. A cache entry is read only after node-gyp has resolved that version itself.


colinhacks:
Fixed in f0c968c. The env-over-argv precedence is real — `lib/node-gyp.js` assigns `npm_config_*` into `opts` after the nopt parse, unconditionally — so nodedir now stands down when the caller already picked headers or a target runtime.

colinhacks:
Reworked in 4cffbd17. The `npm_config_nodedir` export is gone; nub now seeds node-gyp's own `<devdir>/<version>` cache entry, which node-gyp reads only after resolving that version from its own argv and env.

That closes the override channels behind all three reviews, including the ones nub cannot see from the outside: in a fixture where a child process sets `npm_config_target` (the @electron/rebuild and electron-builder shape), the export made node-gyp compile against the running Node and the build succeeded where it had to fail. Without it, node-gyp honors the target.

colinhacks:

PR #939, opened 2026-09-12
phantom: treat host-provided modules as declared

colinhacks:
`electron` and `vscode` are injected by their host process; no install can supply them. A reference is now `Verdict::HostProvided`, not a phantom, so `electron-log` stays in the shared store. `PHANTOM_SCANNER_VERSION` 5 -> 6 so cached verdicts re-scan.

An exports-condition skip was tried in earlier commits and dropped: condition-selected TypeScript from `node_modules` is a supported surface (`@zod/source` under tsx and under Nub), and the skip changed no eject in either measured tree.


colinhacks:

PR #940, opened 2026-09-12
pm: run Windows dependency lifecycle scripts through the bundled busybox sh

colinhacks:
On Windows, dependency lifecycle scripts (`preinstall`, `install`, `postinstall`) ran through `cmd.exe`, while `nub run` has used the bundled busybox-w32 `sh` since 0.6.0. This moves lifecycle scripts onto the same shell.

This is #630 ported to `main`. That PR was merged only into the build-jail integration branch, and nothing in it depends on the jail.

## Why

- **One shell.** A POSIX script body behaved differently depending on which surface ran it.
- **No compatibility cost measured.** Across the 344 widely used packages that run install scripts, none of the 363 script bodies uses cmd-only syntax (`%VAR%`, `if exist`, `copy`/`del`, `>nul`, `set VAR=`, `.cmd` calls).
- **Fixes packages that break today.** `detox-recorder` and `svf-lib` run a shipped `./*.sh` from `postinstall`, which `cmd.exe` cannot execute.

## Changes

- `EngineContext::default_script_shell` replaces the platform default when the user sets no `script-shell`. Nub sets it on Windows to `busybox.exe sh -c`. A missing sidecar warns and leaves the engine on `cmd.exe`.
- Shell selection and argument quoting now come from one resolver, so a POSIX default on Windows no longer gets cmd-style quoting.
- busybox-w32 up-cases every environment name, so `$npm_package_name` in a lifecycle body expanded to nothing. `nub run` already re-binds the lowercase names (#830); that prologue moves into `aube-scripts` and the lifecycle spawn uses it too. The body is appended after the last `env` call, because `run_script` sets `npm_lifecycle_event` and the manifest variables after it builds the shell. Shell-mode `dlx` gets the same ordering.
- A change of lifecycle shell rebuilds dependency output. A tree built under `cmd.exe` can hold wrong bytes, so it must not survive the switch. The shell goes into three keys: the install freshness hash (so the warm path does not return early), the build-policy hash (so the lifecycle phase rescans built dependencies), and the side-effects cache key (so a cached build from the old shell is not restored). The platform default adds nothing to any of them, so installs and caches on macOS and Linux are unchanged.
- Nub's own PM-layer warnings print without `RUST_LOG`. Before this, a missing busybox sidecar left lifecycle scripts on `cmd.exe` with no visible notice.
- Docs: the script-shell section of the run page says install scripts use the same shell.

## Verification

`tests/busybox-lifecycle-probe/` is a Windows differential: one binary, each case installed twice, and the only variable is whether `busybox.exe` sits beside `nub.exe`. From [run 34719883850](https://github.com/nubjs/nub/actions/runs/34719883850) on `windows-latest`, every case split in both directions:

| case | control (`cmd.exe`) | with the sidecar (busybox) |
| --- | --- | --- |
| `${LIFECYCLE_PROBE:-posix}` and `test -d .` | exit 0, wrote the literal `${…}` | exit 0, `MARK=posix` |
| `./scripts/build.sh` | exit 1, `'.' is not recognized …` | exit 0, `SH_RAN` |
| `for f in src/*.txt; …` | exit 1, `f was unexpected at this time.` | exit 0, `A\nB` |
| `$npm_package_name:$npm_lifecycle_event` | exit 0, wrote both names literally | exit 0, `probedep-lowercase-npm-env-names:postinstall` |

The first and last rows are the ones worth noting: `cmd.exe` exits 0 while writing the wrong bytes.

Committed tests: `install_runs_lifecycle_scripts_under_a_posix_shell` runs a POSIX-only root `postinstall` that also reads `$npm_package_name`, so on the Windows CI leg it fails without busybox or without the prologue. `a_lifecycle_shell_change_rebuilds_a_warm_install` installs a `file:` dependency whose build script appends to a marker at the project root, repeats the install as the warm control, then changes `script-shell` and requires one more build. Measured on a builder VM: 1 build, 1 after the repeat (`Already up to date`), 2 after the shell change, 2 after a further repeat. With only the freshness hash changed, the install left the warm path but did not rerun the build, which is why the build-policy hash carries the shell too.

Unit tests pin the busybox applet form, the re-binding of a name set after the shell is built (with a plain shell as the control), both hash contributions, and the cache key.

Gates, on builder VMs: `clippy --all-targets --all-features` clean, whole-workspace `cargo test` green, the vendored aube suite green (`aube-scripts` plus the side-effects cache tests), `cargo fmt --check` clean locally.

Not verified: the win32-arm64 sidecar, which only the release matrix runs.

Refs #630


colinhacks:
Shipped in v0.9.2: https://github.com/nubjs/nub/releases/tag/v0.9.2

PR #941, opened 2026-09-14
config: refuse a backslash in the `prefix` string form

colinhacks:
The string form of `prefix` is split like a POSIX shell, so a backslash escapes the next character and is dropped. A Windows path did not survive that: `"C:\\tools\\wrap.exe --"` launched `C:toolswrap.exe` — a confusing not-found error as a program, silent corruption as an argument.

The lexer cannot switch on the host, since `nub.jsonc` is read on every platform. The string form now refuses the character, and the error names the array form.

Checked against a built binary: the string form errors, the array form carries `C:\dir\.env` intact, and plain and shell-quoted string prefixes are unchanged.


colinhacks:

PR #943, opened 2026-09-14
nubr: a third name on the nub binary, the unified runner on every channel

colinhacks:
The unified runner — a file, a `package.json` script, or an installed bin under one name — was reachable only through `@nubjs/runner` on npm. The nub binary already embeds `runtime/nubr.mjs`, so it now answers to `nubr` too.

```sh
curl -fsSL https://nubjs.com/install.sh | bash
nubr app.ts        # a file
nubr build         # a package.json script
nubr vitest run    # an installed node_modules/.bin entry
```

## What changed

- A third argv[0] personality, `Argv0::Nubr`, runs the embedded `runtime/nubr.mjs` on the resolved Node. It is the same file the npm package ships, so the two installs behave identically by construction; nothing about resolution or the spawn is reimplemented. The binary answers `--version` itself, because the extracted runtime carries no `package.json`.
- Every channel that manufactures the `nubx` alias manufactures `nubr` the same way: `install.sh`, `install.ps1`, `nub upgrade` on both platforms, the tap formula generator (with a `brew test` line that proves the alias dispatched without touching Node), `flake.nix`, `make install-dev`, and the `@nubjs/nub` launcher (`bin/nubr` + the `bin` map).
- Docs: a "From the nub installer" section on the runner page, a README section, and the docs index install line.

## Verification

- The built binary symlinked as `nubr` and `nubr-dev`: `--version` prints `0.9.2`, `--help` prints the runner's usage, a TypeScript file with an `enum` runs, a `package.json` script runs with `npm_lifecycle_event` set, a `node_modules/.bin` entry runs, exit code 3 propagates, and a Node flag before the file passes through. The same fixture under `node runtime/nubr.mjs` prints the same output.
- `tests/verb-dispatch/probe.mjs` now covers the third verb: 16 ok on macOS, including the healed trampoline carrying `__NUB_ARGV0='nubr'`.
- `tests/installer/run.sh`: 0 failed. `cargo fmt --check` clean. Remote clippy job rc=0.

## Outside this repo

- The Homebrew core formula and the mise registry `bins` list each need a one-line change once a release carries this; the tap bumps itself.
- winget is unchanged: `Nubjs.Nub` was never submitted, and winget needs a distinct file per alias.


colinhacks:

PR #945, opened 2026-09-15
build: verify Cargo.locks correctly; docs: say what actually bounds the threadpool

colinhacks:
Two independent changes, no behaviour change to the runtime. This PR started as a threadpool stand-down; measuring the premise disproved it, so that commit is gone and what remains is the tooling fix plus the corrections the measurements do support. The open item is filed as #947.

## 1. Lockfile verification

The recipe this repo documented ran `cargo metadata --locked --offline` per workspace and treated any non-zero exit as a stale lock. Two unrelated failures exit 101 there, and the misleading one is the common one:

```
stale lock   error: cannot update the lock file ... because --locked was passed
cold cache   error: attempting to make an HTTP request, but --offline was specified
```

A fresh checkout, container or CI runner has no populated crate cache, so `--offline` dies on the first uncached crate. It reported all five locks stale here on a tree where every one was clean, which is how it surfaced. That fails in the expensive direction, since the obvious response is to regenerate good locks and commit the churn.

`scripts/check-lockfiles.sh` keeps three verdicts instead of two. It reports staleness only on cargo's own wording for it, and `unverified` at exit 0 when neither cache nor network can answer. The workspace list comes from `git ls-files`, so it cannot drift the way `crates/nub-phantom` did while ungated. Wired into `make verify` and `.githooks/pre-push`, range-scoped to Cargo.toml/Cargo.lock, bypass `NUB_SKIP_LOCKFILE_CHECK=1`.

Checked against known answers:

| Case | Old recipe | New checker |
| --- | --- | --- |
| Clean tree, warm cache | ok | 5/5 ok, exit 0 |
| Dep in manifest, absent from lock | STALE | STALE, that lock alone, exit 1 |
| Clean tree, empty CARGO_HOME, no network | STALE (false) | unverified, exit 0 |
| `--fix` on a stale lock | n/a | re-resolved, entry present |

The pre-push gate was confirmed to block by committing a deliberately stale manifest and watching it refuse the push, then resetting.

## 2. The Threadpool docs promised the wrong mechanism

The shipped page said the demoted workers "yield most of the CPU to the other work on the machine" under contention. That holds only within one cgroup. A nice value ranks a thread against tasks in its own group, and the kernel never reads it when dividing CPU between groups ([`sched(7)`](https://man7.org/linux/man-pages/man7/sched.7.html)), so the sentence promised a neighbouring container a protection the demotion does not provide.

What does provide it is the group's CPU weight, which bounds a cgroup's share however many threads it runs. One cgroup fixed at four threads while the other varies, on 4 CPUs:

| Other cgroup's threads | Fixed cgroup's throughput |
| --- | --- |
| 4 | 195 Mloop/s |
| 16 | 192 Mloop/s |
| 64 | 191 Mloop/s |
| 16, both in one cgroup (control) | 58 Mloop/s |

Sixteen times the threads costs a neighbouring cgroup about 2% and a same-cgroup neighbour 70%. The two mechanisms cover different halves, so the page now names both. Both priority benchmarks also note which topology they measure: they launch co-tenants and server from one shell, so they share a cgroup, and nothing in their output said so — which is how a same-cgroup result reached the docs as a general claim about busy hosts.

`wiki/research/libuv-threadpool.md` records the scheduling and memory measurements, including the reversal, so the next person does not re-derive the wrong conclusion.

## What is deliberately not here

No clamp on the pool. The real cost is memory: in a cgroup bounded by weight, 24 `sharp` resizes of a 12 MP PNG run at 2.02, 1.87 and 1.90 ops/s for pools of 4, 16 and 64 while peak RSS goes 157 → 314 → 455 MB. Flat throughput, 2.9x the memory, because the extra threads cannot run but each in-flight task still holds its working set. Fixing that needs a defensible megabytes-per-thread budget or an entitlement estimate, and the per-task working set spans 4x between two inputs to one library, so nothing ships here. Tracked in #947.

The earlier stand-down commit is dropped rather than reverted, since it was never merged and its reasoning is preserved in #947 and in this PR's comments.

## Gates

`cargo fmt --check`, `clippy --all-targets --all-features`, `nub-core` 507 passed, `nub-launcher` clippy + build + 84 tests, `threadpool_size` 14 passed, all five lockfiles verified, `lat check`, and the pre-push site build. The Rust gates were run against the earlier head that carried the code change; this head touches no Rust.

Refs #919, #947

https://claude.ai/code/session_014jkAYPaPaAeURxPh8nYM1A


colinhacks:
Moved to draft: the premise above is wrong and this should not be merged as written.

The claim was that a nice value does not cross a cgroup boundary, so a cores-sized pool is unprotected on a dense host. The first half is true. The second does not follow, because cgroup weight already bounds a group's CPU share independently of how many threads it runs. Measured on 4 CPUs, cgroup A against cgroup B at equal weight, B fixed at four threads:

| A's threads | B throughput | B share |
| --- | --- | --- |
| 4 | 195 Mloop/s | 51% |
| 16 | 192 Mloop/s | 49% |
| 64 | 191 Mloop/s | 47% |
| 16, both in one cgroup (control) | 58 Mloop/s | 19% |

Across a boundary the neighbour loses ~2% as the pool grows 16x; inside one cgroup it loses 70%. The control rules out a failed cgroup assignment, which would have read 19% throughout.

So the two mechanisms are complementary and #919 already covered both cases: thread count steals only within a group, which is where nice 10 applies (`pool-priority-nub.sh` measured ~1% neighbour loss), and across groups the weight does the work. This PR stands down precisely where protection already existed, and it would disable the sizing for Docker without `--cpus`, Kubernetes without `limits.cpu`, systemd services with the cpu controller enabled, and CI.

What survives is the memory axis, which is independent: pool size sets concurrency, and 32 queued `sharp` resizes of a 12 MP PNG peak at 137 MB over baseline with four threads against 585 MB with sixty-four. A pod with a memory limit and no CPU quota still gets a host-core-sized pool, and is also the case where the extra threads buy least, since it can only use its weighted slice.

Next: withdraw the CPU-based stand-down, keep `wiki/research/libuv-threadpool.md`, and treat memory as a separate change with its own evidence for any constant it needs. Gates on the current head were green (fmt, clippy all-targets/all-features, nub-core 507 passed, nub-launcher 84 passed, threadpool_size 14 passed, all five lockfiles verified), which is not the issue here.

---
_Generated by [Claude Code](https://claude.ai/code)_

colinhacks:
Added a second, unrelated commit to this branch: `b7cc7f3`, the lockfile-verification fix. It is independent of the threadpool change above and stands on its own whatever happens to that one.

The recipe this repo documented ran `cargo metadata --locked --offline` per workspace and read any non-zero exit as a stale lock. Two different failures exit 101 and the misleading one is the common one:

```
stale lock   error: cannot update the lock file ... because --locked was passed
cold cache   error: attempting to make an HTTP request, but --offline was specified
```

A fresh checkout, container or CI runner has no populated crate cache, so `--offline` fails on the first uncached crate. It reported all five locks stale here on a tree where every one was clean, which is how it surfaced. That is the expensive direction, since the reflex fix regenerates good locks and commits the churn.

`scripts/check-lockfiles.sh` keeps three verdicts instead of two, reporting staleness only on cargo's own words for it and `unverified` at exit 0 when neither cache nor network can answer. The workspace list comes from `git ls-files`, so it cannot drift the way `crates/nub-phantom` did while ungated. Wired into `make verify` and `.githooks/pre-push` (range-scoped to Cargo.toml/Cargo.lock, bypass `NUB_SKIP_LOCKFILE_CHECK=1`), since a stale lock otherwise costs a ~20-minute CI round trip.

Checked against known answers, not just a green run:

| Case | Old recipe | New checker |
| --- | --- | --- |
| Clean tree, warm cache | ok | 5/5 ok, exit 0 |
| Dep in manifest, absent from lock | STALE | STALE, that lock alone, exit 1 |
| Clean tree, empty CARGO_HOME, no network | STALE (false) | unverified, exit 0 |
| `--fix` on a stale lock | n/a | re-resolved, entry present |

The pre-push gate was confirmed to block by committing a deliberately stale manifest and watching it refuse the push, then resetting.

---
_Generated by [Claude Code](https://claude.ai/code)_

colinhacks:
Rewritten and opened for review. The threadpool stand-down commit is dropped — force-pushed rather than reverted, since it was never merged and the reasoning is preserved in the comments above and in #947.

What is left touches no Rust or runtime code:

- `scripts/check-lockfiles.sh` and its wiring, unchanged from the earlier head.
- The Threadpool docs page, which promised that the nice-10 demotion protects other work on the machine. It protects same-cgroup work; a neighbouring container is bounded by cgroup weight instead.
- A note on both priority benchmarks naming the topology they measure, which is the same-cgroup one.
- `wiki/research/libuv-threadpool.md`, carrying the measurements and the reversal.

The memory cost is real and unfixed here, because no constant for a clamp is defensible yet. #947 has the numbers and the candidate rules.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #948, opened 2026-09-15
blog: replace the extensions table's labels with a hover explanation

colinhacks:
The undeclared-imports column of the extensions table showed up to two grey words after every import (`unscanned dependency`, `guarded`, `optional`) that nothing on the page explained. The column now lists the import names alone. Hovering, focusing or tapping a name opens a popover that says how that package leaves the import undeclared and what the rule adds.

Verified in `next dev` on desktop and phone widths, light and dark, with the site typecheck and the pre-push `next build`.


colinhacks:

PR #949, opened 2026-09-16
blog: keep each extension rule's version range in the table

colinhacks:
The extensions table dropped the version range from every rule selector, so the 128 carried Yarn rules read as if they applied to every version of their package: `redux-thunk` was listed for `redux`, which 2.4.0 declares, because Yarn's rule is `redux-thunk@<=2.3.0`.

Each edge now keeps its range. The package column shows it when every rule on the row shares one, and the popover names the package with its range. An unknown finding class fails the generator instead of being coded as a carried rule.


colinhacks:

PR #950, opened 2026-09-16
blog: hide extension rules the package's current release has outgrown

colinhacks:
A ranged rule rendered identically to one covering every version, so the table read as an accusation against current releases. `debug` ranks first in the figure on downloads and its rule is scoped to `<4.2.0` against a current 4.4.3.

Measured across the database: 186 of 203 ranged rules no longer match their package's latest, accounting for 95 of the 791 rows.

## Changes

- The generator resolves each ranged package's `dist-tags.latest` and checks it against the rule's range with semver, emitting an `outgrown` flag per edge and the latest version per row.
- The table filters those rules out by default and drops any row left empty, showing 696 of 791. A checkbox brings them back, dimmed and dashed, with a popover that names the version the phantom was fixed in.
- The rules are unchanged and still ship — an install resolving a version inside the range needs them.
- Class sentences say "references" rather than "imports": the scan counts `require.resolve('x')` as an edge, correctly, since it throws `MODULE_NOT_FOUND` on a miss, but it loads nothing.
- Popover pinned to the opaque `fd-popover` token. fumadocs ships `bg-fd-popover/60`, which composited to within ~3 of 255 of the page background here.
- Short code tokens held on one line: a hyphen is a CSS break opportunity, so `redux-thunk@<=2.3.0` split after `redux-` and read as two packages.

Prose counts (791 / 649 / 142) are unchanged — they are database counts and stay true.

## Notes for review

- semver is unpacked into a temp dir rather than added as a dependency, since the root is nub-identity and installs under a frozen lockfile. 103 distinct ranges including compound, alternation and prerelease forms, so hand-rolling was not viable.
- `--keep-downloads` reuses the recorded counts without revalidating the week, for a reshape on a host that cannot reach `api.npmjs.org`. It is the one path that can emit a week-old count; the data here carries the 2026-09-05..09-11 window verbatim.

## Verification

- Data diff: 791 rows unchanged, zero unintended changes to `d`, `y`, or `edge[0..2]`.
- Generator output reproduces an independently written analysis: 186/203, 95 hidden, 696 shown.
- `tsc --noEmit` and `next build` clean.
- Headless Chromium, dark + light + 390px: checkbox coverage 1.0, `clipped: false`, no page horizontal overflow. The 49-character worst-case token wraps inside the popover with nothing painting outside it.

https://claude.ai/code/session_01FxkN49P7MyFZvah6GWBpfG

---
_Generated by [Claude Code](https://claude.ai/code/session_01FxkN49P7MyFZvah6GWBpfG)_


colinhacks:
`pullfrog` is red for a provider quota error, not this PR: `action failed: provider error: The usage limit has been reached`, after six `AI_APICallError` retries on `openai/gpt-sol`. The run is `workflow_dispatch` on `refs/heads/main` at `2a4573e`, not this branch's head.

No fix exists on this side — nothing in a blog/table diff reaches an LLM provider usage limit. Not re-running, since the log names the cause outright. The rest of CI is unaffected and still in progress.

---
_Generated by [Claude Code](https://claude.ai/code)_

colinhacks:

PR #951, opened 2026-09-17
microbe: an incubated minimal npm installer

colinhacks:
An incubation, linked by nothing in the `nub` binary. One crate that installs a package and its dependency tree into a directory the caller names, verifies integrity, and reports the bin entries. No lockfile, no store, no lifecycle scripts.

Stripped, `opt-level = "z"` + fat LTO, aarch64-apple-darwin: 493 KB default, 805 KB with `--features tls`. TLS is the whole budget, so the default links none and borrows an HTTPS client the host has — Node's `fetch`, then curl, then wget. An embedder implements `Transport` instead.

Six tests cover the install path against an in-memory registry. Against the real registry, express, typescript, esbuild, biome and prettier install and run.


colinhacks:
Moved to nubjs/microbe as its own repository; the crate no longer lives in this tree.

PR #960, opened 2026-09-19
plugin-yaml: a YAML plugin as loader hook, content mapper, and Vite plugin

colinhacks:
Prototype of the plugin shape from the content-mapper discussion: one package with a `./register` loader hook (plus its ambient types), a TypeScript 7.1 content mapper, and a `./vite` transform. No Rust changes; `nub check` and the `plugins` config key are not in this PR.

Verified locally on typescript 7.1.0-dev.20260919.1 (real per-file types, parse errors mapped into the YAML file), typescript 7.0.2 (wildcard fallback), `node --import`, Nub with a preload, and `vite build`. `node --test npm/plugin-yaml/test.mjs` covers emit, the hook, and the transform.

Refs microsoft/typescript-go#4712


colinhacks:
241511c: initialize returns protocolVersion 1; the hook and the Vite transform emit through the mapper's serializer, so .nan/.inf/-0.0 round-trip and an own __proto__ key is a computed property; the suite adds a fidelity test on all three paths and a framed handshake against the mapper process, and the docs-links CI job installs the package and runs it. The package is an unpublished prototype; the README says so.

colinhacks:
279128b: a collection mapping key now renders as the same flow YAML text parse() produces (a flow clone through stringify; a node's bare toString() is JSON), and a null key is the empty string. The fidelity test compares against parse() on nested collection keys. The earlier 5ac9a10 went out with that test red; 279128b is the green one.

